PR #63, opened 2015-01-04
test harness improvements

askmyc:


jorgebay:
Looks good.

Could you address the following issues:
- The KillAllCcmProcess() is killing java process, this cause java process (not related to the tests) from other apps to be killed. It can be replaced by a "TryRemove()" method (tries to remove and ignore remove exceptions).
- Use .NET Trace.Information(), Trace.Warning() and Trace.Error() methods. Do not use the driver's logger.

askmyc:
Updated as per your feedback, re-ran all tests to make sure all is well

jorgebay:
Looks good, I will merge it on 2.1 and then merge 2.1 into master.

PR #65, opened 2015-01-06
tests for CqlOperator, csharp_81

askmyc:
There are a few straggling questions, 
Notes as seen in code: 
NUnit.Framework.Ignore("Issue with converting back to array, pending question")
NUnit.Framework.Ignore("Pending question -- this might not be possible for C* Maps, but the error message doesn't look good either")


jorgebay:
I've merged your tests b357044418f7f11167d77a4d2dbdab2076384808
As you pointed out, the conversion back from array was not working correctly for array types in the mapper (and Linq). I fixed it in e16c148fbeacb7c5eb9142f402502cf63d226640.

PR #68, opened 2015-01-14
test additions and general clean up

askmyc:


jorgebay:
There are some problems with the PR:
- Exceptions tests class has category long and most methods category short. This causes those tests to belong to both categories. This also happens in other test classes (PoolTests, MetadataTests).
- There are 19 file mode changes. Avoid this by disabling file mode in git: $ git config core.fileMode false
- PoolTests still share the test cluster. This is not desirable and results in weird behaviour. No long test should share the test cluster. This should be applied to 2.1 branch.

Also, we should try to avoid massive pull requests as they are difficult to review and difficult to rebase. I think there are 3 different subjects here (possible prs):
- Fixes to be made to tests that exists on 2.1 and master: PoolTests, Metadata, ExceptionTests.
- Add the final tests for 2.5 release: (MapperSingle, MapperUpdate, etc).
- Fixed to be made to the tests from master (Linq/Mapper tests).

askmyc:
this was restructured in PR 71

PR #73, opened 2015-01-25
removing test ignores, adding nested collection mapping test

askmyc:


jorgebay:
Looks good, should I merge it with the ignored tests on mapping + nested?

askmyc:
Yes, I figure we can keep that test and enable it when/if that capability is added

PR #75, opened 2015-02-16
RPTokenFactory: Fixing BigInteger parsing hashes as negative

mintsoft:
In situations where the last byte has it's most significant bit set to 1, the BigInteger will parse it as signed and therefore make it negative. This is not the desired behavior for the RandomPartitioner. Padding with an empty byte forces the parser to treat the number as unsigned.

Fixes https://datastax-oss.atlassian.net/browse/CSHARP-253

### DIFF
--- a/src/Cassandra/RPToken.cs
+++ b/src/Cassandra/RPToken.cs
@@ -63,8 +63,14 @@ namespace Cassandra
 
             public override IToken Hash(byte[] partitionKey)
             {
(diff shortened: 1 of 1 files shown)


jorgebay:
I cherry picked 1984ebcc984169b98a2ea8c26e046d6a56dad843 into 2.1, as this affects both 2.1 and master branches.
I will create some tests for it asap.

Thanks,
Jorge

PR #76, opened 2015-02-17
RPToken: BigInteger needs little endian; MD5 outputs big endian

mintsoft:
This is the only implementation that I can get to to work. It feels like I'm handling the bytearray incorrectly, however I can't see it. In situations when the most significant byte is 0xFF; it flips the value
negative, however if I add 0x00 byte padding at the end of the array thenthe number is totally wrong
```
Expected: 188090901659892217213821499076270282
But was:  340094276019278571246160785932691941174
```
Really not sure what's going on there, hopefully a fresh pair of eyes can figure it out. Either way, this is (at least) a working implementation

### DIFF
--- a/src/Cassandra/RPToken.cs
+++ b/src/Cassandra/RPToken.cs
@@ -63,8 +63,20 @@ namespace Cassandra
 
             public override IToken Hash(byte[] partitionKey)
             {
-                if (_md5 == null) _md5 = MD5.Create();
-                return new RPToken(new BigInteger(_md5.ComputeHash(partitionKey)));
+                if (_md5 == null) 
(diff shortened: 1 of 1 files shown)


mintsoft:
Related to: https://datastax-oss.atlassian.net/browse/CSHARP-253

jorgebay:
I've also checked yesterday on the rest of the DataStax drivers and they are using `abs()` for negative values with the RandomPartitioner.

Do you mind making the pull request based on 2.1 and in a single commit?
If you don't have the time, let me know, and I will implement it myself.

mintsoft:
@jorgebay I'll do that in a minute, no problem

PR #77, opened 2015-02-18
RPToken: Hash calculation handles negatives incorrectly

mintsoft:
MD5 hash returns bytes in big endian whilst BigInteger requires little endian
Cassandra & other drivers use ABS for negative values rather than the padding that was implemented in 1984ebcc984169b98a2ea8c26e046d6a56dad843

As requested; change to 2.1 and PR against it!

Related to: https://datastax-oss.atlassian.net/browse/CSHARP-253

### DIFF
--- a/src/Cassandra/RPToken.cs
+++ b/src/Cassandra/RPToken.cs
@@ -65,12 +65,17 @@ namespace Cassandra
             {
                 if (_md5 == null) 
                     _md5 = MD5.Create();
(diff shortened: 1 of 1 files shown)


mintsoft:
@jorgebay adding tests for this is a little more problematic than I'd like, unfortunately the visibility makes it a little awkward, however the test cases attached to CSHARP-253 will all pass.

jorgebay:
No problem, we will add them.

PR #78, opened 2015-02-18
TokenMap: A host should only be counted as a replica if it's not been seen for this token before

mintsoft:
References to: https://datastax-oss.atlassian.net/browse/CSHARP-254

This was developed against 4 nodes, RF=3 on the RandomPartitioner with the https://github.com/datastax/csharp-driver/pull/77 merged in


jorgebay:
I rebased it and merged to 2.1 d1c96edf20fc202d65bf8f6ef56d5038999618cb

Thanks!

PR #79, opened 2015-02-18
add setup helper scripts, reformat some comments

askmyc:


jorgebay:
The integration tests just have 1 environment requirement: ccm. If that requirement is not met, I think its correct to fail.

I think this scripts don't belong in the repository for 2 reasons:
- All scripts should be platform independent. If we want to run the tests on linux, we state as a prerequisite to install ccm (how to install ccm? its platform dependent, so we don't care).
- I don't want us to deal with the maintenance and support of build scripts, other than: 
  - For windows: `msbuild build/build.proj /v:m /t:test`
  - For POSIX: [`xbuild src/Cassandra.sln`](https://github.com/datastax/csharp-driver/blob/master/.travis.yml#L10)

PR #80, opened 2015-02-20
fix for CqlInsert.SetTimestamp

vytautassurvila:
CqlInsert.SetTimestamp inserts timestamp in milliseconds. According to specs it should be in microseconds. The fix makes timestamp calculation for inserts same as update and delete statements already have


jorgebay:
Thanks for the fix!

PR #82, opened 2015-03-05
Fix memory leak: unsubscribe from SocketEventArgs event handlers, dispos...

dolphy17:
Hi,

This is another memory leak we found when using the following code:

Looping many-many-many times the following:
using (Cluster cassandraCluster = Cluster.Builder().AddContactPoint(address).Build())
{
   using (cassandraCluster.Connect("system"))
   {
      ...doing Cassandra stuff...
   }
}

SocketEventArgs should be disposed in the Dispose() method too. 
In addition, event handlers should be properly unregistered.

Please review.

### DIFF
--- a/src/Cassandra/TcpSocket.cs
+++ b/src/Cassandra/TcpSocket.cs
@@ -316,7 +316,9 @@ namespace Cassandra
             }
             if (_receiveSocketEvent != null)
             {
+                _sendSocketEvent.Completed -= OnSendCompleted;
                 _sendSocketEvent.Dispose();
+                _receiveSocketEvent.Completed -= OnReceiveCompleted;
                 _receiveSocketEvent.Dispose();
             }
             else if (_socketStream != null)
@@ -377,6 +379,16 @@ namespace Cassandra
                     return;
                 }
                 _isClosing = true;
+                if (_sendSocketEvent != null)
+                {
+                    _sendSocketEvent.Completed -= OnSendCompleted;
+                    _sendSocketEvent.Dispose();
+                }
+                if (_receiveSocketEvent != null)
+                {
+                    _receiveSocketEvent.Completed -= OnReceiveCompleted;
+                    _receiveSocketEvent.Dispose();
+                }
                 //Try to close it.
                 //Some operations could make the socket to dispose itself
                 _socket.Shutdown(SocketShutdown.Both);


jorgebay:
About the usage you describe:
You should use and reuse the [same session instance](http://www.datastax.com/documentation/developer/csharp-driver/2.5/csharp-driver/csdFaq.html), within your application lifetime.

About the call to SocketAsyncEventArgs `Dispose()` method:
We are calling it in the `OnClosing()` method. The `OnClosing()` method is invoked asynchronously by an IO thread, that normally happens after Socket.Close() (called on TcpSocket.Dispose() method). The freeing of the resources does not happen synchronously by design, so I'm happy with the current implementation.
Calling `Dispose()` before Socket.Close(), according to the [.NET implementation detail](http://referencesource.microsoft.com/#System/net/System/Net/Sockets/Socket.cs,8883), should have the same effect, it will be invoked again by an io thread after operations complete, (never trust implementation details) but it won't happen synchronously by design.

About the unregistering of the event handlers, they are useful only if the even handler (TcpSocket method) has a shorter life that the event emitter (SocketAsyncEventArgs), which is not the case.

That said, calling `Dispose()` on a resource does not mean that resources are synchronously freed and GC'ed.

dolphy17:
Hi Jorge,

I agree that under normal circumstances we should only use one instance of
Session object, but that doesn't it's allowed to leak memory. In the above
code we test multiple Cassandra instance nodes one-by-one and we have to
create-destroy the whole Cluster instance in order to properly isolate the
tests. I second the unsubcribing event handlers is not necessary, however
it doesn't make any harm.

We (LogMeIn) run the test in an IIS process, maybe that's causing something
different from a desktop application. The OnClosing() method doesn't get
called every time a request is completed and resources are freed up for
sure. I tested it. The Dispose() method itself should care about destroying
all resources no matter how the actual logic is expected to work.

Please double check and test the logic, because it's clearly fragile and
I'll raise a ticket at Datastax side to get this fixed if we can't deal
with it here.

Thanks,
Andor

jorgebay:
You mean different Cassandra cluster or nodes?
If you mean clusters (different independent rings), you are using it correctly, if not, you should use and reuse the same `ICluster` and `ISession` instances, as the driver will connect to all the cluster nodes (not just the contact point, but all of them) according to the load balancing policy, thanks to the autodiscovery feature.

If you want to test node by node inside the same cluster, you will need a "white list" load balancing policy (which is not really of much use).

If you are experiencing memory leaks, I will do whatever I can to fix it, but this is probably not the root cause.

Using your example:
```csharp
public void UseAndDispose()
{
  using (var cluster = Cluster.Builder().AddContactPoint(address).Build())
  {
    using (var session  = cluster.Connect("system"))
    {
       //do whatever here
    }
  }
  //If I measure the memory consumption here
  //I will still get the allocations made by Cluster instance as the Garbage Collector didn't collect it yet.
  //Additionally, as the freeing of the resources on the Socket are not synchronous, it will still be in memory
}
```

dolphy17:
I'd like to test the individual nodes of the same cluster.
Would you please provide some pseudo-code example, because I don't really
get your point on "white listing" policy?

class SingleNodeLoadBalancingPolicy : ILoadBalancingPolicy
{
     public SingleNodeLoadBalancingPolicy(string configuredNode)
     {
     }
...should return the configured node all the time...
}

But node balancing policy is part of cluster configuration, not the
session. So the cluster must be re-created every time I'd like to switch
over another node.
If I change the configuration of my custom load balancer, will it be
reflected in the driver immediately or there's some caching in place?

Andor

jorgebay:
It is possible to reuse the same load balancing policy and change the target node, as you suggest.

It could be something like this:
```csharp
public class OneNodeAtATimePolicy : ILoadBalancingPolicy
{
    private ICluster _cluster;
    private volatile Host _currentHost;

    public void Initialize(ICluster cluster)
    {
        this._cluster = cluster;
    }

    public HostDistance Distance(Host host)
    {
        return HostDistance.Local;
    }
    
    public void SetNode(IPEndPoint address)
    {
        _currentHost = _cluster.AllHosts().First(h => h.Address == address);
    }
    
    public void GetCurrentHost()
    {
        return _currentHost ?? _cluster.AllHosts().First();
    }

    public IEnumerable<Host> NewQueryPlan(string keyspace, IStatement query)
    {
        return new [] { _currentHost };
    }
}
```

And by calling to SetNode(), you change the active host.

The functionality of this load balancing policy is limited and most common use cases are covered in the standard policies, but if you need it for a reason like testing purposes, go ahead.

Hope it helps!

dolphy17:
Thanks Jorge. Unfortunately, it doesn't seem to solve my original problem.
(although requests have become a bit faster)

Using like this:

_loadBalancingPolicy.SetNode(IPAddress.Parse(address));
using (_cassandraCluster.Connect("system"))
{
  return
     _cassandraCluster.Metadata.AllHosts()
          .Select(host => new KeyValuePair<string,
IPAddress>(host.Datacenter, host.Address))
          .ToList();
}

And after a few hundreds of requests it's clear that GC is unable to catch
up and free up resources.
So, again. We have a memory leak here. :(

Regard,
Andor


[image: Inline image 1]

[image: Inline image 2]

jorgebay:
I think you tried to attach images but it was not included (filtered by github).

Please reply in the pull request to include images.

Could you provide a full code sample and a detail of what is the issue you are experiencing in order to reproduce it?

You can open a [JIRA ticket](https://datastax-oss.atlassian.net/projects/CSHARP/issues), if you want.

dolphy17:
The pull request has already been deleted by you, I cannot find a suitable
spot to include images. :)
Could you please provide some regular email address that I can post my
sample app?
Anyway I'm trying to attach here as well.

Regards,
Andor

jorgebay:
The pull request is closed, which means that it is not in process to be merged, but you can still access online at https://github.com/datastax/csharp-driver/pull/82  I'm actually replying there.

You can join the [C# driver mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!forum/csharp-driver-user) and post your images there via email or you can create a [Jira ticket](https://datastax-oss.atlassian.net/projects/CSHARP/issues).

If you need to post sensible data, you should use the DataStax support channels.

dolphy17:
cassandra_memory_leak1
cassandra_memory_leak2
cassandra_memory_leak3

Sample code: (Driver version 2.1.5)
    class Program
    {
        static void Main(string[] args)
        {
            Cluster cluster = Cluster.Builder().AddContactPoint("192.168.11.110").Build();
            for (int i = 0; i < 10000; i++)
            {
                using (ISession session = cluster.Connect("system"))
                {
                    using (session.Execute("select keyspace_name from schema_keyspaces limit 1;"))
                    {
                    }
                }
                Console.WriteLine("Loop: {0}", i);
            }
            Console.ReadLine();
        }
    }

Thanks,
Andor

jorgebay:
Thanks for providing more info. I still can not reproduce it, but it could be possible that the OnClosing() method is not being invoked, depending on your platform socket implementation.

I've checked the source code for .NET 4.5 and mono 3.10 and it is safe to call SocketAsyncEventArgs.Dispose() multiple times, also by [IDisposable.Dispose() specification](https://msdn.microsoft.com/en-us/library/system.idisposable.dispose.aspx): _If an object's Dispose method is called more than once, the object must ignore all calls after the first one. The object must not throw an exception if its Dispose method is called multiple times. Instance methods other than Dispose can throw an ObjectDisposedException when resources are already disposed_.

I've created a temporary repo branch with the additional calls to SocketAsyncEventArgs.Dispose(), if you could check it out and run your test again to see if its fixed, that would be great:
```bash
git clone https://github.com/datastax/csharp-driver.git
git checkout socket-multiple-dispose
```

Compile using Visual Studio or using MSBuild:
```bash
msbuild.exe build/build.proj /v:m /t:compile
```
 I could send you the binaries if needed.

dolphy17:
Confirmed that issue is fixed on the new branch.

ANdor

jorgebay:
Thanks for giving it a try and your feedback.
I will include it for the upcoming 2.5.1 release.

dolphy17:
Thanks Jorge. Do you have an ETA for 2.5.1?

One more thing that just crossed my mind. In the OnClosing() method, I
think you should also set _sendSocketEvent and _receiveSocketEvent to null,
like this:

if (_receiveSocketEvent != null)
{
    //It is safe to call SocketAsyncEventArgs.Dispose() more than once
    _sendSocketEvent.Dispose();
    _sendSocketEvent = null;
    _receiveSocketEvent.Dispose();
    _receiveSocketEvent = null;
}

...to make them GC'able until next Connect() happens.
What would be even more better, I believe, is to use a pool of
SocketEventArgs for sending and receiving which could help making the
memory less fragmented. However, I'm not entirely sure about that bit.

Ref:
http://www.codeproject.com/Articles/83102/C-SocketAsyncEventArgs-High-Performance-Socket-Cod

Anyway, thanks for the quick help.
I'm looking forward to integrate with the new driver.

Cheers,
Andor

dolphy17:
Hi Jorge,

Would you please provide an ETA on 2.5.1 release? 

Thanks,
Andor

jorgebay:
I can't provide any specific target dates but you can expect it the "next few weeks".
You can follow the progress of the tickets included in the release on [Jira](https://datastax-oss.atlassian.net/browse/CSHARP-211?filter=-4&jql=project%20%3D%20CSHARP%20AND%20fixVersion%20in%20(2.5.1)%20ORDER%20BY%20status%20ASC%2C%20updated%20).

dolphy17:
Ok, thanks.
 2015.03.10. 11:43 ezt írta ("Jorge Bay Gondra" <notifications@github.com>):

jorgebay:
It is now merged into 2.1: da230910379b553da878cd89fc048a8d6168680f.
I've also created a stress test that creates hundreds of session instances, disposes the cluster and dereferences and checks the memory consumption: 353294edfd9105a18ab3a315a3b69b42b95ee81b.

It will merged into master and released in 2.5.1 soon.

dolphy17:
Hi Jorge,

Thanks for fixing the memory leak issues. We've just deployed the new
driver in production and it works fine.

Quick question regarding Performance Counters. I see there's a static class
CassandraCounters which should populate the counters, but it doesn't work
even if I enable it with the Diagnostics switch. I see there's nothing
calling that class. Was it intentional removing it?

Thanks,
Andor



On Thu, Mar 12, 2015 at 10:54 AM, Jorge Bay Gondra <notifications@github.com

PR #83, opened 2015-03-05
Builder.WithPort  not working

aNutForAJarOfTuna:
"withPort" in Builder is not setting the port, but using default port 9042.

### DIFF
--- a/src/Cassandra.Tests/BuilderTests.cs
+++ b/src/Cassandra.Tests/BuilderTests.cs
@@ -96,5 +96,21 @@ namespace Cassandra.Tests
                                                   .Build());
             Assert.AreEqual(ex.SocketErrorCode, SocketError.HostNotFound);
         }
+        
+        [Test]
+        public void AddContactPointsWithPortShouldHaveCorrectPort()
+        {
+        	const string host1 = "127.0.0.1";
+        	const string host2 = "127.0.0.2";
+        	
+        	int port = new Random().Next(9000, 9999);
+        	var builder = Cluster.Builder().AddContactPoint(host1).WithPort(port);
+        	var cluster = builder.Build();
+        	Assert.AreEqual( cluster.AllHosts().Last().Address.Port, port);
+        	
+        	builder = Cluster.Builder().AddContactPoints(host1, host2).WithPort(port);
+        	cluster = builder.Build();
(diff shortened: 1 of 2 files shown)


jorgebay:
Nice catch, I think this was bug was caused by the introduction of the address translator #42, as the contact points are now ip endpoints.

I will review it shortly.

aNutForAJarOfTuna:
Thanks jorgebay, may I know normally how long would the changes in master reflect on nuget packages ?

jorgebay:
It depends...

Could you also fix the use of `_ports` in the `AddContactPoints()` method and git squash your commits?

It should be something like:
```csharp
public Builder AddContactPoint(IPAddress address)
{
  AddContactPoint(new IPEndPoint(address, _port));
  //Use _port                              ˆˆˆˆ
  return this;
}
```

I'll try to include it in the upcoming version 2.5.1.

aNutForAJarOfTuna:
Ok. So do we want the interface to set different port for different contact point in the same line like this: 

builder
.AddContactPoint("addr1").WithPort(9140)
.AddContactPoint("addr2").WithPort(9240)
...

or the last "WithPort" would override the previous ports in the contact points?

jorgebay:
The port setting is for all nodes of the cluster, the driver will also use it when discovering the rest of the nodes in the cluster.
In the following example:
```csharp
builder
  .WithPort(9011)
  .AddContactPoints("host1", "host2");
```
If there is a host3 discovered by the driver, it will use the port setting as well.
For different port configurations and stuff like that, the user must create an AddressTranslator.

aNutForAJarOfTuna:
Sounds good, just made a small change as you state and should be good to merge now.

jorgebay:
I've cherry picked it to master: 3a750876e185a81f0b759fcf7102d6edbb2c539e.
Also, I've found another bug in the method `AddContactPoints()` method, fixed in 0faf3cb582eb11de15abafe27bf5baaa76c96e1f, with additional tests.

Thanks!

jorgebay:
I've also changed `List<T>.ForEach()` call to a language `foreach` iteration as `ForEach()` method is [not supported in all platforms](http://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,583).

PR #84, opened 2015-03-11
Linq and Mapper support for Lightweight Transactions CSHARP-251

jorgebay:


LukeTillman:
I left a comment in the JIRA ticket RE: batching conditional updates.  Have a look there, otherwise LGTM!

https://datastax-oss.atlassian.net/browse/CSHARP-251

jorgebay:
I've added support for batches containing conditional batches, using the `bills` schema with static columns, it would look like this:
```csharp
var batch = mapper.CreateBatch();
batch.Update<Billing>("UPDATE bills SET balance = ? WHERE user=? IF balance = ?", 501, "user1", 30);
batch.Insert(billing);
var info = mapper.ExecuteConditional<Billing>(batch);
if (!info.Applied)
{
  var existingBalance = info.Existing.Balance;
}
```

It will only insert the next "bill" if the balance haven't changed. If it changed, the user gets the existing balance.

LukeTillman:
Looks good, :shipit:

PR #86, opened 2015-03-23
Linq Select for Update supports variables (not only constants)

medvekoma:
This pr fixes the bug that appears when trying to specify a variable in the Select part of an Update or UpdateIf.

    var dateTimeValue = DateTime.Now;
    table
        .Where(t => t.IntValue == 100)
        .Select(t => new AllTypesDecorated { DateTimeValue = dateTimeValue })
        .Update()
        .Execute();

Unit tests were added that demonstrate the problem.

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUpdateUnitTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUpdateUnitTests.cs
@@ -113,6 +113,7 @@ namespace Cassandra.Tests.Mapping.Linq
                 parameters = v;
             });
             var table = session.GetTable<AllTypesDecorated>();
+
             table
                 .Where(t => t.BooleanValue == true && t.DoubleValue > 1d)
                 .Select(t => new AllTypesDecorated { StringValue = "updated value" })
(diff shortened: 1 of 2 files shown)


jorgebay:
Nice catch @medvekoma

jorgebay:
There is a workaround for this: create the new entity with the values you want to update and then use this values in the projection.

It is kind of wasteful but at least is a workaround until this bug fix is released.

```csharp
var user = new User { Name = variableName};
userTable
    .Where(t => t.Id == userId)
    .Select(t => new User { Name = user.Name })
    .Update()
    .Execute();
```

medvekoma:
Thanks Jorge, this works indeed.

Just an idea, wouldn't be more logical to define Update as

    Update<T>(Expression<Action<T>> updateAction)

so that we can write something like

    userTable
        .Where(t => t.Id == userId)
        .Update(t => t.Name = userName)
        .Execute();

jorgebay:
I agree about the `Update(expression)` method, but it was defined long ago, so we would have to still support `Select()` method.

medvekoma:
Yes, we need to stay compatible, but it could be an overload. Anyway, I am fine with the current approach too.

jorgebay:
Rebased and merged daca387626d5c354a06c3207ca106124752de8fe.
I've added additional support for new instance creations in the projections 6418242864d310d875882c7ee3fd9665c8dc1337.

Thanks @medvekoma

ablanchet:
I've just encountered this issue too. The workaround is fine but I was wondering when you are planning to release a new version (2.5.2 ?) of the driver ? I would definitely prefer to use the fixed version rather than the workaround.

jorgebay:
Thanks for locating the issue you are experiencing.
I can not give release target dates, but it is coming really soon. If you are developing an application using the driver and you want to use the fix, I would recommend you to use the master branch (clone and compile) and then replace it with the Nuget version when 2.5.2 is released.

ablanchet:
Ok, I'll have to wait until next week to work on this project now, wait and see :). Thanks

ablanchet:
Thanks @jorgebay for the release :)

PR #88, opened 2015-03-25
Determine nearest datacenter

medvekoma:
Currently, local datacenter is determined by simply selecting the first endpoint from all hosts. This is not only suboptimal, but could even lead to selecting a data center that is not accessible by the client. 

In my network I can reach DC1, but cannot reach DC2. This logic selected DC2 as local datacenter even if I only passed DC1 nodes as connection points. As result, all connections to the cluster have failed.

This fix takes the first node from each datacenter, and tries to connect to each of them in parallel. Once the fastest node replies, it selects that datacenter and doesn't wait for other connections.

As the connections are executed in parallel, this solution has a minimal performance impact (according to my tests 56 ms between a client in Hungary and cluster in France) and is executed only once per session.

### DIFF
--- a/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
+++ b/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
@@ -138,7 +138,7 @@ namespace Cassandra
 
             // try to connect to one node in every datacenter
             var tasks = firstHostByDatacenter
-                .Select(item => new TaskFactory().StartNew(state => new TcpClient().Connect(item.EndPoint), item.EndPoint))
+                .Select(item => new TaskFactory().StartNew(() => new TcpClient().Connect(item.EndPoint)))
                 .ToArray();
             // the index of first task that returns
             var index = Task.WaitAny(tasks);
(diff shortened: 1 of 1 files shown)


jorgebay:
I think that it is a good idea and implementation for your needs but I don't see it as part of the built-in policies.

There is a constructor parameter in the policy to manually set the datacenter and this should be used for that. For different deployments, the local datacenter parameter can be set via application configuration.

I think it is a good idea to maintain the policies as is, with the same behaviour as the other DataStax driver.

medvekoma:
OK, I was not aware of the fact that I can control the local datacenter via configuration, I will check that option. I agree that the same behavior should be kept among all drivers. I am going to close this pr. 

Is there a possibility to fall back to another datacenter, if the current one is down?

jorgebay:
Ok, cool.
To use failover to another datacenter, you can use the [constructor overload specifying hostsPerRemoteDc parameter](https://github.com/datastax/csharp-driver/blob/master/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs#L91).

It won't connect to the hosts located on the remote datacenters until it has to (all other nodes failed), not before, so it would be safe to use a any number there.
//dc hungary is local but if it fails, connect to 5 remote hosts as a max
var policy = new DCAwareRoundRobinPolicy("hungary", 5);

medvekoma:
Thanks Jorge, this is very useful!

PR #89, opened 2015-03-25
Because mapper always sets pagingstate we should see if we are setting i...

ilushka85:
...t to null before disabling autopaging

### DIFF
--- a/src/Cassandra/Statement.cs
+++ b/src/Cassandra/Statement.cs
@@ -104,12 +104,12 @@ namespace Cassandra
         {
             ProtocolVersion = 1;
         }
-        
+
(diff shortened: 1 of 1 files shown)


jorgebay:
Looks good, I will create some covering tests, fix the current one and merge it.

ilushka85:
@jorgebay do you have an eta for nuget for this fix as it basically means all existing working code breaks as soon as you move to 2.5.1

jorgebay:
It is now merged into master, I created additional tests here 93a1304b28c742dc627b1b2c0e8a53bc464f72bf

I'm not able to speak about release dates, I hope it will be soon.

PR #90, opened 2015-03-25
Add ttl and timestamp support to insert/update operations

nicodeslandes:
This PR updates the `Mapper` and `CqlBatch` classes to provide the ability to specify a **TTL** or a **TIMESTAMP** option when performing an insert operation.

This is done using the new `CqlQueryOptions` class when calling `IMapper.Insert`, like this:
```csharp
mapper.Insert(newUser, new CqlInsertOptions
                       {
                           Ttl = TimeSpan.FromSeconds(42),
                           Timestamp = DateTimeOffset.Parse("2014-2-1")
                       });
```
The more "Java-like" syntax offered by `CqlQueryOptions` is also available:
```csharp
mapper.Insert(newUser,
    CqlInsertOptions.New().
        SetTtl(TimeSpan.FromSeconds(42)).
        SetTimestamp(DateTimeOffset.Parse("2014-2-1")));
```

The `InsertIfExists` variant can also take `CqlQueryOptions`.

The same can be implemented for Update operations, but I thought 'd get some feedback on the initial implementation for Inserts first before moving on to that.

### Note
Another way to add TTL and TIMESTAMP options would have been to either change or subclass `CqlQueryOptions`. But I thought it might be better to add a new class altogether for 2 reasons:
* The `Insert`/`InsertIfExists` methods already take a `CqlQueryOptions` and I didn't want to change their signature, to avoid breaking backward compatibility issues. I could have subclassed `CqlQueryOptions` without changing the signature of `Insert`, but then the new options would not have been easily discoverable
* The current `CqlQueryOptions` define options to apply at the `IStatement` level, and don't control how CQL query strings are generated, so it has a slightly different scope from `CqlInsertOptions`.

### DIFF
--- a/src/Cassandra.Tests/Mapping/CqlGeneratorTests.cs
+++ b/src/Cassandra.Tests/Mapping/CqlGeneratorTests.cs
@@ -145,5 +145,19 @@ namespace Cassandra.Tests.Mapping
             var cql = cqlGenerator.GenerateInsert<ExplicitColumnsUser>();
             Assert.AreEqual(@"INSERT INTO ""USERS"" (""UserId"", ""Name"", ""UserAge"") VALUES (?, ?, ?)", cql);
         }
+
+        [Test]
+        public void GenerateInsertWithTtl_CaseSensitive_Test()
+        {
+            var types = new Cassandra.Mapping.Utils.LookupKeyedCollection<Type, ITypeDefinition>(td => td.PocoType);
+            types.Add(new Map<ExplicitColumnsUser>()
+                .TableName("USERS")
+                .PartitionKey(u => u.UserId)
+                .CaseSensitive());
+            var pocoFactory = new PocoDataFactory(types);
+            var cqlGenerator = new CqlGenerator(pocoFactory);
+            var cql = cqlGenerator.GenerateInsert<ExplicitColumnsUser>();
+            Assert.AreEqual(@"INSERT INTO ""USERS"" (""UserId"", ""Name"", ""UserAge"") VALUES (?, ?, ?)", cql);
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


jorgebay:
Looks good, I'll make a full review shortly.

jorgebay:
I don't think its a good idea to add more cql generation to the Mapper, I think the purpose of the Mapper is to not get in the way of the queries. It generates and completes the default queries, but if the user wants cql control (like in this case field TTL), the user should write the query using the other method overloads.

Initially I thought it was related to protocol-level timestamp, but not query based timestamp and TTL, sorry for that.
    
@LukeTillman: What do you think?

LukeTillman:
Yeah, I agree about keeping the CQL generation to a minimum.  Although I do think we should support protocol level timestamp in v3:

https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v3.spec#L322

Since TTLs are in seconds, those are pretty easy to just provide as query parameters using the BCL (i.e. `TimeSpan.FromXXX().TotalSeconds`).  Query-based timestamps are a little trickier since they're microseconds since epoch and I have a utility class I've been using for to generate values for those:

https://github.com/LukeTillman/killrvideo-csharp/blob/master/src/KillrVideo.Utils/MicrosecondsSinceEpoch.cs

LukeTillman:
I wouldn't be opposed to including a query builder that builds `Cql` instances though (that could potentially include TTL and Timestamp methods).  Doesn't the Java driver have a builder API like that?

jorgebay:
The java driver has a query builder API, but we also have Linq that generates the cql queries from a typed expression. I think, from a product standpoint, it makes sense to recommend Linq when the user wants cql generation (and mapping) and recommend the Mapper when the user wants to write their own queries and get the results mapped.

collinsauve:
@jorgebay I believe this can be closed?  Ttl is now supported as an overload to insert (but not update), and timestamp is supported on both insert and update via CqlQueryOptions.

jorgebay:
Thanks for the reminder!

PR #92, opened 2015-04-13
Explicitly table name for Mapper class

sp9usb:
Hello,

In my company we using cassandra enginee with datastax csharp-driver. We have the problem with mapping the one model to more tables. This change solves our problem. 
Please, consider adding it to the main branch.

Regards,
Kamil Myśliwiec

### DIFF
--- a/src/Cassandra/Mapping/Mapper.cs
+++ b/src/Cassandra/Mapping/Mapper.cs
@@ -89,7 +89,7 @@ namespace Cassandra.Mapping
         public Task<IEnumerable<T>> FetchAsync<T>(Cql cql)
         {
             //Use ExecuteAsyncAndAdapt with a delegate to handle the adaptation from RowSet to IEnumerable<T>
-            _cqlGenerator.AddSelect<T>(cql);
+            _cqlGenerator.AddSelect<T>(cql, _table);
             return ExecuteAsyncAndAdapt(cql, (s, rs) =>
             {
                 var mapper = _mapperFactory.GetMapper<T>(cql.Statement, rs);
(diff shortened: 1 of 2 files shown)


jorgebay:
I agree that we should have support in the `Mapper` for mapping 1 entity to `n` tables, but I'm not sure if this is the best way to provide it.
Mapper methods use the poco type as a generics parameter (not the Mapper instance itself), so it will be best to add it by method.

@LukeTillman What do you think?

LukeTillman:
I agree that doing this at the method level is definitely more appropriate than at the `Mapper` level (for the reasons Jorge pointed out). I have some doubts though. I feel like the most common use case for this is to update some base table and the denormalized copies of that data living in other tables.  Those denormalized tables are likely to have a different `PRIMARY KEY` definition than the "base" table (i.e. the table that the POCO is mapped to by default).

For an `INSERT`, just specifying the table name would be enough (since we don't need to do any sort of `WHERE` clause there). But for `UPDATE`, `DELETE`, and many of the `SELECT` cases, we do generate a `WHERE` clause based on the `PRIMARY KEY` columns. So for those cases, just specifying a different table name probably isn't enough information, we'd probably need to know the key columns as well to generate the appropriate `WHERE` clause. And that probably means allowing users to specify the key columns somehow at the method level along with the table name.

I wonder if rather than adding these at the driver level, the [Global Indexes in CASSANDRA-6477](https://issues.apache.org/jira/browse/CASSANDRA-6477) that are coming in C* 3.0 take care of the need for this instead? Does that make any sense?

francischie:
I also use switching table source for the same entity like map to a materialized view with same identical columns and leveraging same fluent mapping settings. You can just implement this on your repository class. The table class already have overload or optional parameters for table name. 

public CqlQuery<T> CreateQuery(Expression<Func<T, bool>> expression = null, string tableName = null)
        {
            var query = string.IsNullOrEmpty(tableName)
                ? Table
                : new Table<T>(Session, MappingConfiguration.Global, tableName);

            if (expression != null)
                return query.Where(expression);

            return query;
        }

jorgebay:
Closing this pull request for no activity since the last comments.

Feel free to create a new ticket or submitting a new pull request if needed.

PR #93, opened 2015-04-24
Improve documentation for BatchType enum

godefroi:


jorgebay:
Merged, thanks for the fix!

PR #94, opened 2015-06-08
Replace 'select * ..' with explicit column names

ImDarrenG:
When building a cql string in CqlExpressionVisitor.GetSelect(), use explicit column names rather than 'SELECT *', to reduce risk of schema mismatch during schema migrations.

This way, the driver asks for the columns it knows about, rather than everything - which leads to an exception if the poco is out of step with the cassandra schema.

Most of the changes are unit test changes, which had to be changed from SELECT * to SELECT [all columns]

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqEntryPointsTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqEntryPointsTests.cs
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using Cassandra.Data.Linq;
+﻿using Cassandra.Data.Linq;
 using Cassandra.Mapping;
 using Cassandra.Tests.Mapping.Pocos;
 using NUnit.Framework;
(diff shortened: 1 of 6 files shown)


jorgebay:
The commit is clean and the pull request looks good, but I just wanted to know in which situation this could be a problem. In the mapping definitions, you can mark them as explicit or set to ignore a certain class property.

Additionally, from the user's perspective there is the possibility to use `Select()` to specify the fields in a select query: `table.Where(...).Select(...).Execute()`.

ImDarrenG:
No worries.

In our setup we have multiple frontend servers.  When deploying changes that include schema changes we do a rolling update of the frontend servers.  So what we get is a window where some of the servers contain code that represent old c* schema, and some servers on the new C* schema, and the cassandra schema has been updated.  This can lead to exceptions for the servers with the old code as in the case that a column has been added for example, the query returns columns that the poco doesn't know about yet.  By building the query to retrieve only the columns that the poco knows about, can mitigate this risk.

jorgebay:
Sorry for taking so long to get back to you, we are sprinting to provide [Cassandra 2.2 support in the driver](https://datastax-oss.atlassian.net/browse/CSHARP-304?filter=-4&jql=project%20%3D%20CSHARP%20AND%20fixVersion%20in%20(2.6.0-rc1)%20ORDER%20BY%20status%20ASC%2C%20updated%20).

I think your the approach you propose is the safest choice (for new users) and we should include it in the driver, but we should check if that could be a behavioural breaking change for users: existing users with no mapping definitions could use SELECT * type of queries via Linq and only the columns returned will be mapped (ignoring the rest of the POCO properties that are not really part of the table).

I created a JIRA ticket for it: https://datastax-oss.atlassian.net/browse/CSHARP-308

ImDarrenG:
Thanks for the update

jorgebay:
Landed in a54f011b38eec1ddb13f97704f54d6639ebca72c

Thanks @ImDarrenG

PR #99, opened 2015-06-17
Support Cassandra 2.2 Unset

jorgebay:
`Unset` is a class with a singleton, similar to .NET `DbNull`.


jorgebay:
Removed implicit `Unset` for `PreparedStatement.Bind()` based https://github.com/datastax/csharp-driver/commit/442725f4073b494818b3e426081246bbcaf51daf#commitcomment-11861820

Thanks @LukeTillman

PR #103, opened 2015-06-23
fix streamId not found in pendingOperation in Connection object

aNutForAJarOfTuna:
I'm trying to fix crash when streamId not found in _pendingOperation in Connection object; this is an issue when cassandra under high load and not easy to reproduce.

### DIFF
--- a/src/Cassandra/Connection.cs
+++ b/src/Cassandra/Connection.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 //      Copyright (C) 2012-2014 DataStax Inc.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
@@ -493,7 +493,8 @@ namespace Cassandra
                 if (header.Opcode != EventResponse.OpCode)
                 {
                     //Its a response to a previous request
-                    state = _pendingOperations[header.StreamId];
+                    if(! _pendingOperations.TryGetValue(header.StreamId, out state))
+                    	return false;
                 }
                 else
                 {


jorgebay:
I think we need to investigate the underlying cause and why there is just 1 report of this occurring and we could not reproduce the issue in any way, and you are the only one experiencing this issue.

Could you create a JIRA ticket containing all the info possible
- Schema
- Queries
- How many Cluster and Session instances you use
- Driver settings
- ...
 
About the pull request, it looks clean but throwing the exception (and crashing the app domain) is OK there, hiding it is not a good idea, we could add more logging into it just to give more information.

aNutForAJarOfTuna:
@jorgebay  Thanks, I'll create a JIRA ticket for this, but still not be able to reproduce it very easily since it might related to the work load. I'll try a stand alone project with a testing cluster to separate out the problem.

synaxin:
I'm part of same organization as @aNutForAJarOfTuna 
We have a 5 node cluster, dual quad core, 64Gb ram, SSD raid array of about 1Tb space.
Schema wise, we have 10 tables, 2 are high write, 2 are high read.

Is it possible that there is a race condition - 
Connection::SendQueueProcessItem throws an exception (due to a remote cassandra node that is under heavy load flapping), thereby causing the exception caught to remove the pending operation at
Connection.cs#663

Meanwhile a frame is being read in, which the line at Connection.cs#496 would throw the exception?

on that same notion, it could be a race condition between multiple calls to ReadParse, using the same thought.  The first removes the pending operation, the second throws the exception.

I don't know how we could reproduce this in the existing testing suite, other than operations being performed while simultaneously stopping the server.   We will try to reproduce, but any suggestions would speed that up for sure.  Currently we are enabling the trace switch to capture log output on error.

jorgebay:
Thanks for giving some thought into it and trying to find the root cause.

About the case you are describing, it will not cause any issues. The streamid is removed if it was not possible to successfully write it to the wire, or to say it differently: `pendingOperations` contains stream ids of requests that were successfully written but not received the response for it.

There are several integration tests in the repository that test the scenario that you describe: see 
- https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/Core/StressTests.cs
- https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/Core/PoolTests.cs
- https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/Core/SessionTests.cs

We should continue looking for it :smiley:

jorgebay:
Are you using v2.5.2 from Nuget.org or tag 2.5.2 from repository, right?

synaxin:
We are using v2.5 from tag, but the only difference between that and 2.5.2
tag is documentation.

 I'm trying to collect Trace information now - we've since resolved our
load issues though, so it may be hard for even us to reproduce the
environment now.  We will have to try on a staging cluster with our old
config.

jorgebay:
I'm closing this as we can not reproduce the underlying issue you describe in our integration and duration tests.
If you have any new information, please let us know.

PR #105, opened 2015-07-01
TypeConverter substitution in MappingConfiguration

Virus-X:
```MappingConfiguration``` has already the method ```ConvertTypesUsing``` that can be used to inject custom ```TypeConvertor```. But it turns out, that this method implemented incorrectly.
As ```MapperFactory``` contains cache, I think the most correct way is to recreate it.

### DIFF
--- a/src/Cassandra/Mapping/MappingConfiguration.cs
+++ b/src/Cassandra/Mapping/MappingConfiguration.cs
@@ -71,6 +71,7 @@ namespace Cassandra.Mapping
         {
             if (typeConverter == null) throw new ArgumentNullException("typeConverter");
             _typeConverter = typeConverter;
+            MapperFactory = new MapperFactory(_typeConverter, new PocoDataFactory(_typeDefinitions));
             return this;
         }
 


jorgebay:
Nice catch!
Instead of recreating the `MapperFactory`, we could expose the property and set it. What do you think?

Virus-X:
Thanks for response!
I think that just exposing property is not enough. Here's why:
```MapperFactory``` has cache of mapper functions (```_valueCollectorFuncCache```).
After first call to ```GetValueCollector()``` with some arguments, collector function would be cached and chages to ```TypeConvertor``` won't take effect for that particular query.
So, after setting the new TypeConvertor, we must clear the cache.

If recreating of ```MapperFactory``` would introduce some side effect, you can add special method that would set new type convertor and clear ```_valueCollectorFuncCache``` and ```_mapperFuncCache```.
Since they are not accessible outside, I think it should work without problems.

jorgebay:
The mapping configuration should be set before trying to use the Linq or the Mapper. 
As an API, maybe that is not clear (because a global configuration is used in case you don't specify one and it starts using its internal cache), but that is the way we found to share the configuration between 2 different components Linq and the Mapper. Maybe that should be stated more clearly in the [components' docs](http://docs.datastax.com/en/developer/csharp-driver/2.5/csharp-driver/reference/linqComponent.html?scroll=linq-component__linq-comp-configuration).

That is the reason why I don't think it's a problem to allow the user to set the `TypeConverter` without clearing the internal cache.

Virus-X:
Agreed, this would work correctly if ```TypeConverter``` is set before using the Linq or the Mapper.

jorgebay:
Thanks @Virus-X for your contribution.
I've added a [unit test](https://github.com/datastax/csharp-driver/commit/e0c9f4fa07c7f5a44ff00e8a2e7ec62f96cd5b86) to cover it.

PR #106, opened 2015-07-08
Timer-based read timeout

jorgebay:
Uses Hashed Wheel Timer (CSHARP-311) and uses a copy-on-write collection for connections (CSHARP-314) to allow removal of connections from the pool (defunctioning).


jorgebay:
Closed in favor #114

PR #109, opened 2015-07-14
Speculative executions

jorgebay:
- ISpeculativeExecutionPolicy
- ConstantSpeculativeExecutionPolicy

Chained with CSHARP-295 as most of the timer-related changes were needed.


jorgebay:
Rebased against master and squashed all speculative executions support in 1 commit

PR #111, opened 2015-07-21
Add support for StartsWith() in linq expressions

alprema:
I usually emulate `string.StartsWith(...)` by slicing from the **start string** to **start string + Char.MaxValue** so I wanted the driver to support it since it could be useful for people that don't know this trick.

I unit tested my change and tried it against some test data on my cluster, using both Latin and non-Latin characters but I'm new to expression visitors so I welcome any comment.

### DIFF
--- a/src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
+++ b/src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
@@ -29,6 +29,8 @@ namespace Cassandra.Data.Linq
 {
     internal class CqlExpressionVisitor : ExpressionVisitor
     {
+        private readonly string _utf8MaxValue = Encoding.UTF8.GetString(new byte[] { 0xF4, 0x8F, 0xBF, 0xBF });
+
         private static readonly Dictionary<ExpressionType, string> CqlTags = new Dictionary<ExpressionType, string>
         {
             {ExpressionType.Not, "NOT"},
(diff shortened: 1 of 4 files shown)


alprema:
Hi, any news on this ? Anything I should change ?

LukeTillman:
I think we should have a chance to take a look at this next week when Jorge is back. Sorry about the delay.

alprema:
Ok, I didn't know he was out, no problem.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -608,6 +608,30 @@ private bool EvaluateMethod(string name, MethodCallExpression node)
                         .Append(")");
                     return true;
                 }
+                case "StartsWith":
+                    Visit(node.Object);
+                    var startsWithArgument = node.Arguments[0];
+                    var startString = (string)Expression.Lambda(startsWithArgument).Compile().DynamicInvoke();
+                    var endString = startString + char.MaxValue;
```
jorgebay: I think `endString`must be
```csharp
                                      //hex representation of U+10FFFF
startString + Encoding.UTF8.GetString(new byte[] {0xF4, 0x8F, 0xBF, 0xBF})
```
As the [largest UTF8 codepoint according to the spec](https://tools.ietf.org/html/rfc3629#section-3) is [10FFFF](http://www.fileformat.info/info/unicode/char/10ffff/index.htm) and not [FFFF] (http://www.fileformat.info/info/unicode/char/ffff/index.htm) (Char.MaxValue).

alprema:
You are correct, I replaced Char.MaxValue with UTF8's max value + ran a few tests with dummy data.

jorgebay:
Looks good! Could you rebase against master and squash the commits so I can merge it? 

I created a [JIRA ticket to track this feature](https://datastax-oss.atlassian.net/browse/CSHARP-328).

alprema:
Done

PR #113, opened 2015-08-13
Fully async FetchPage

icanhasjonas:
How about defaulting the behavior of rowset paging to async instead of the reverse? The implementation seems to be async all the way any way.

Here's a sample PR


jorgebay:
Thanks for your contribution.

It looks good, my only concern is that having to calls to the async method would result on duplicated rows in the `RowSet`.

This can be alleviated by removing the lock and using a control field like pageState that gets [compared-and-set](https://msdn.microsoft.com/en-us/library/system.threading.interlocked.compareexchange(v=vs.110).aspx) before issuing the new request and spin waiting on the actual task to finish.

The most common use case of paging is doing it by iterating through the rows. As iterating is a synchronous operation and the async solution is somehow involved, we decided to go with the sync + lock solution, but there is nothing that prevents us from doing the async paging the default: I created a [ticket](https://datastax-oss.atlassian.net/browse/CSHARP-335) for it.

Sorry for the delay in the answer, we are [sprinting for v2.7](https://datastax-oss.atlassian.net/browse/CSHARP-333?filter=-4&jql=project%20%3D%20CSHARP%20AND%20fixVersion%20in%20(2.7.0)%20ORDER%20BY%20status%20ASC%2C%20updated%20).

jorgebay:
Closing this pull request for no activity since the last comment.

PR #122, opened 2015-08-28
[CSHARP-332] Integration tests for per-host request timeout

kishkaru:
Integration tests for per-host request timeout


src/Cassandra.IntegrationTests/Cassandra.IntegrationTests.csproj
```diff
@@ -58,6 +58,7 @@
     <Compile Include="..\SharedAssemblyInfo.cs">
       <Link>Properties\SharedAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="Core\ClientTimeoutTests.cs" />
```
jorgebay: Nice catch! I forgot to include it on #114

PR #124, opened 2015-09-02
Bugfix: Select anonymous type containing a nullable column

oguimbal:
I ran into a quite anoying issue.

Suppose you have this schema

    create table mytable(id stringPRIMARY KEY, mydouble double);
    insert into mytable (id,mydouble) values ('id1',0);
    insert into mytable (id,mydouble) values ('id2',1);

Mapped to class:

    public class MyTable{ 
        public string Id{get;set;} 
        public double? MyDouble{get;set;}  // nb: This is nullable
    }

And you perform this request:

    myTable.Select(x=>new{x.Id,x.MyDouble}).ToArray();

Everything should be fine.
But now, insert a null value

    insert into mytable (id,mydouble) values ('id3',null);

...and perform the same request.

Exception thrown:
**Cannot convert null to System.Double because it is a value type, try using Nullable<System.Double>**

This commit fixes it

### DIFF
--- a/src/Cassandra.Tests/Mapping/FetchTests.cs
+++ b/src/Cassandra.Tests/Mapping/FetchTests.cs
@@ -300,5 +300,46 @@ namespace Cassandra.Tests.Mapping
             Assert.AreEqual("Come Away with Me", song.Title);
             Assert.AreEqual(DateTimeOffset.Parse("2002-01-01 +0").DateTime, song.ReleaseDate);
         }
+
+
+        [Test]
+        public void Fetch_Anonymous_Type_With_Nullable_Column()
+        {
+            var songs = FetchAnonymous(x => new { x.Title, x.ReleaseDate });
+            Assert.AreEqual("Come Away with Me", songs[0].Title);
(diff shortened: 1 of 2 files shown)


jorgebay:
Nice fix!

Could you clean your commit to avoid reformatting existing code to avoid messing up with history?

I created [Jira ticket CSHARP-346](https://datastax-oss.atlassian.net/browse/CSHARP-346) to track this issue.

oguimbal:
@jorgebay done

oguimbal:
Not sure to understand why this test fails on  Travis CI build: SetDown_Should_Get_The_Next_Delay_Once_The_Time_Passes

(... works on my machine)

jorgebay:
I asked travis to rebuild it, the test result is not conclusive.

oguimbal:
Not conclusive ?
It seems to pass now... doesnt it ?
In the test:
   
      Thread.Sleep(3000); 

That must be a timing/concurrency issue...

jorgebay:
Yes, it is.

I meant that the test failing does not prove that there is an issue: the test needs to be fixed not the code :smile:

PR #128, opened 2015-09-08
Updated doxyfile_csharp

clintascencio-datastax:
Updated for new parameters.


jorgebay:
Could you rebase it against master?

PR #135, opened 2015-09-29
Remove unnecessary copies from read path

alprema:
`OutputRow.ProcessRowItem()` and `TypeCodec.DecodeDouble()` were doing unnecessary copies of the byte buffers, slowing down with both the copy and the GC.

I ran a benchmark of a read intensive code sample ([described here](https://groups.google.com/a/lists.datastax.com/d/msg/csharp-driver-user/1dpco-NC5BE/B1BtJJijAgAJ)) and the speed was improved by 47%. I left an array copy in `TypeCodec.DecodeDouble()` to respect the previous behavior of not modifying the input parameter, but swapping its endianness in place would bring the improvement to 54%.

### DIFF
--- a/src/Cassandra/TypeCodec.cs
+++ b/src/Cassandra/TypeCodec.cs
@@ -1000,9 +1000,7 @@ namespace Cassandra
 
         public static object DecodeDouble(int protocolVersion, IColumnInfo typeInfo, byte[] value, Type cSharpType)
         {
-            var buffer = (byte[]) value.Clone();
-            Array.Reverse(buffer);
-            return BitConverter.ToDouble(buffer, 0);
+            return BitConverter.ToDouble(new [] { value[7], value[6], value[5], value[4], value[3], value[2], value[1], value[0] }, 0);
         }
 
         public static byte[] EncodeDouble(int protocolVersion, IColumnInfo typeInfo, object value)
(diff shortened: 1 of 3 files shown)


src/Cassandra/TypeCodec.cs
```diff
@@ -1000,8 +1000,9 @@ public static byte[] EncodeCounter(int protocolVersion, IColumnInfo typeInfo, ob
 
         public static object DecodeDouble(int protocolVersion, IColumnInfo typeInfo, byte[] value, Type cSharpType)
         {
-            var buffer = (byte[]) value.Clone();
-            Array.Reverse(buffer);
+            var buffer = new byte[8];
+            for (var i = 0; i < 8; ++i)
+                buffer[i] = value[7 - i];
             return BitConverter.ToDouble(buffer, 0);
```
jorgebay: It can be fixed:
```csharp
return BitConverter.ToDouble(
     new byte[] { value[7], value[6], value[5], value[4], value[3], value[2], value[1], value[0]}, 0);
```

jorgebay:
Apart from the minor comment, it looks good to me!

alprema:
I updated the patch according to your comment.

jorgebay:
Nice!
I was merging it at the same time :smile:

PR #140, opened 2015-10-05
Remove unneeded byte[] allocation from TypeCodec

ocoanet:
The last allocation remaining in `TypeCode.DecodeDouble()` after @alprema improvements can be removed using a conversion struct. The benchmark now runs 70% faster than the baseline instead of 54%.

`TypeCode.DecodeFloat()` was also updated to benefit from the same enhancements.

### DIFF
--- a/src/Cassandra/TypeCodec.cs
+++ b/src/Cassandra/TypeCodec.cs
@@ -23,6 +23,7 @@ using System.Net;
 using System.Reflection;
 using System.Text;
 using System.Linq;
+using System.Runtime.InteropServices;
 
 namespace Cassandra
 {
@@ -1008,7 +1009,11 @@ namespace Cassandra
 
         public static object DecodeDouble(int protocolVersion, IColumnInfo typeInfo, byte[] value, Type cSharpType)
         {
-            return BitConverter.ToDouble(new [] { value[7], value[6], value[5], value[4], value[3], value[2], value[1], value[0] }, 0);
+            if (value.Length < 8)
+                throw new ArgumentException("Specified array is too small", "value");
+
+            var converter = new DoubleConverter(value[7], value[6], value[5], value[4], value[3], value[2], value[1], value[0]);
+            return converter.DoubleValue;
         }
 
         public static byte[] EncodeDouble(int protocolVersion, IColumnInfo typeInfo, object value)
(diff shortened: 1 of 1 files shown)


ocoanet:
I signed the contribution license agreement. Is the label supposed to be automatically removed?

avalanche123:
Yes, it should have happened automatically, but you've triggered a bug. Should be fixed now.

alprema:
Those changes result from benchmarks that we ran using our read and write use cases.

Both are time series related (supposedly the most common pattern on Cassandra), and some of the changes we made were quite trivial, which makes me wonder how the driver is benched on Datastax's side. Do you do stress testing? On what kind of data? Do you compare the results to other drivers (our benchmarks ran between 2 and 3 times faster on the Java driver)?

More transparency on the performance of the driver would be appreciated.

jorgebay:
I created a micro benchmark of the 2 decoding strategies, the one that is proposed here and the current one.

https://gist.github.com/jorgebay/445d9ba9c0ad371c5796

It seems that there isn’t a significant improvement between the two. In terms of mean, both methods show similar values. In median, using physical layout shows some improvement. I think the readability of the code should not be compromised to deliver this small performance improvement (if any).

About driver benchmarking, we are working on it, it is high in our priority. We will share the results when we have them. We are now focusing on Cassandra 3.0 support.

As we discussed on the mailing list, there are 2 features that are implemented on the java driver that are not yet implemented on the C# side that significantly affects throughput:

- [JAVA-562](https://datastax-oss.atlassian.net/browse/JAVA-562) / [CSHARP-244](https://datastax-oss.atlassian.net/browse/CSHARP-244): Message coalescing can increase throughput from ~ 50% to ~ 100%. I've implemented it in a dev branch and the results are consistent with those figures (around 80% more throughput with coalescing enabled and tcp no delay ON). It reduces the payload overhead (tcp + ip headers) and the amount of sys calls. You can pull it and compare results.
- Buffer pooling / [CSHARP-270](https://datastax-oss.atlassian.net/browse/CSHARP-270): Netty 4 provides a buffer pooling implementation. Buffer pooling can reduce CPU usage, memory fragmentation and GC pressure.

alprema:
Hi,

I just ran your code as is and the result is about 15% in favor of the physical layout, changing the inner iteration count to 1000 to give more time for GC to factor in gives about 47% difference (benchmarking such code with 2000 iterations is clearly not representative).

Finally, removing Linq from the benchmark loop (as follows) gives about 80% difference.
```cs
                    for (var j = 0; j < values.Length; ++j)
                        DecodeFunc(values[j]);
```
instead of
```cs
                    values.Select(DecodeFunc).Sum();
```
(at those speeds, Linq accounts for about half the time of the measurement).

I understand your concern about readability of the code, but when writing serialization code that is called millions of times per seconds (and that is a bottleneck to a very common use case), performance can be chosen over readability.

I am at your disposal if you need more details on my results / tests.

jorgebay:
Showing that a Linq iteration of a couple of values affects the results of the microbenchmark is exactly what my point is: the improvement is small and gets diminished when paired with the rest of the tasks that need to be done within the driver. (ie: thread (Task) scheduling; managing streamIds / inflight requests).

Anyway, it's a valid possible improvement and I created the ticket [CSHARP-357](https://datastax-oss.atlassian.net/browse/CSHARP-357) for it were we can continue discussing and plan for it.

We should create 3 structures for (2 bytes, 4 bytes and 8 bytes) all numeric types, not just double and float.

PR #141, opened 2015-10-06
Expect column name 'system.dateof(timeuuid_sample)' instead of 'dateOf(timeuuid_sample)'

maxwellb:
With Cassandra 2.2.1 this integration test fails because the column name presented to the reader is `system.dateof`, not `dateOf`.

Alternatively, the query could be changed, or the driver could be modified to support column name mapping to the syntax provided in the select query, but that seems more complicated.

If the syntax described is not portable across versions (the format of the column named returned is different), this could be modified to inspect `row.GetValue<DateTimeOffset>(2)` by index.


jorgebay:
Nice catch, using the index makes sense.
Would u like to update your branch?

maxwellb:
Sure

maxwellb:
Let me know if you'd prefer I squash the three commits on my branch, or if you'll squash during merge if applicable.

jorgebay:
I will squash it, no prob.

Just FYI: your commit email is not linked to your github account: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user

jorgebay:
Squashed into 6f097898fd76416c127168b714411250af1ec49f and merged, Thanks @maxwellb !

PR #142, opened 2015-10-07
Dont ccm remove nontest clusters

maxwellb:
Having recently discovered the awesome `ccm` tool, it would be nice if running integration tests didn't blow away my current cluster at the start of the test.

Here's a pull request that will skip the initial removal if the cluster name doesn't look like `test_...`, by inspecting `ccm status` output.

### DIFF
--- a/src/Cassandra.IntegrationTests/TestBase/TestUtils.cs
+++ b/src/Cassandra.IntegrationTests/TestBase/TestUtils.cs
@@ -85,9 +85,11 @@ namespace Cassandra.IntegrationTests.TestBase
         public static readonly string SELECT_ALL_FORMAT = "SELECT * FROM {0}";
         public static readonly string SELECT_WHERE_FORMAT = "SELECT * FROM {0} WHERE {1}";
 
+	    public const string TEST_CLUSTER_NAME_FORMAT = "test_{0}";
+
         public static string GetTestClusterNameBasedOnTime()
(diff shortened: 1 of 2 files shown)


jorgebay:
For the tests, I think executing `ccm  remove` is safer than handling it ourselves: ccm will look into the current cluster, if any, and try to stop it and remove the files. I think we need to call something like that in case the previous run was aborted.

Thanks for contribution and keep them coming!

maxwellb:
Thanks - in this request - even if the previous run were aborted, `ccm remove` would still get called. This is because the previous run would create a cluster called, e.g. `test_123456789`. I agree completely that `ccm remove` is necessary.

However, it seemed to me that the initial `ccm remove` can be a bit greedy. If all tests pass successfully, and I still have another cluster hanging around -- `ccm create -v 2.2.2 somethingelse` then the `ccm remove` will remove the `somethingelse` cluster, which was not created by the tests.

For Jenkins, this probably isn't a problem, but can be a bit of a "gotcha" if nothing else. Running the integration tests in essence "interferes" with other potentially valid uses of `ccm`.

Alternative 1: make use of config-dir to have the tests not use the `~/.ccm` directory.

For now as a workaround I think I'll make sure that when I use `ccm` outside of integration tests, I'll use an alternate config dir for other clusters. Thanks for the feedback.

PR #144, opened 2015-10-07
Update BatchRequest.cs

axelbodo:
Also, I think that other short/byte type protocol fields lacks range checking, so some unit test would be good for edge cases

### DIFF
--- a/src/Cassandra/BatchStatement.cs
+++ b/src/Cassandra/BatchStatement.cs
@@ -113,6 +113,11 @@ namespace Cassandra
         /// <returns>The Batch statement</returns>
         public BatchStatement Add(Statement statement)
         {
+            if (_queries.Count >= ushort.MaxValue)
+            {
+                //see BatchMessage.codec field in BatchMessage.java in server code, and BatchRequest.GetFrame in this driver
+                throw new ArgumentOutOfRangeException(string.Format("There can be only {0} child statement in a batch statement accordung to the cassandra frame protocoll", ushort.MaxValue));
+            }
             _queries.Add(statement);
             return this;
         }
--- a/src/Cassandra/BatchStatement.cs
+++ b/src/Cassandra/BatchStatement.cs
@@ -111,7 +111,7 @@ namespace Cassandra
         /// </summary>
         /// <param name="statement">Statement to add to the batch</param>
         /// <returns>The Batch statement</returns>
-        /// <exception cref="System.ArgumentOutOfRangeException">Thrown when trying to add more than <c>short.MaxValue</c> Statements</exception>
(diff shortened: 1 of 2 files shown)


jorgebay:
We could limit the amount of queries inside a batch here: https://github.com/datastax/csharp-driver/blob/master/src/Cassandra/BatchStatement.cs#L116 and throw an `ArgumentOutOfRangeException` when trying to add more than 32K queries.

Would u like to update the commit on your branch?

axelbodo:
Ok, I'll do it.

axelbodo:
I've done the modification according to BatchStatement child stament count range check.

src/Cassandra/BatchStatement.cs
```diff
@@ -113,6 +113,11 @@ public BatchStatement SetRoutingValues(params object[] keys)
         /// <returns>The Batch statement</returns>
         public BatchStatement Add(Statement statement)
         {
+            if (_queries.Count >= ushort.MaxValue)
+            {
+                //see BatchMessage.codec field in BatchMessage.java in server code, and BatchRequest.GetFrame in this driver
+                throw new ArgumentOutOfRangeException(string.Format("There can be only {0} child statement in a batch statement accordung to the cassandra frame protocoll", ushort.MaxValue));
```
jorgebay: Minor typos:
`"There can be only {0} child statement in a batch statement according to the Cassandra native protocol"`

Also, the [batch length is a `short`](https://git-wip-us.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=doc/native_protocol_v3.spec) so it should be `short.MaxValue`

src/Cassandra/BatchStatement.cs
```diff
@@ -113,6 +113,11 @@ public BatchStatement SetRoutingValues(params object[] keys)
         /// <returns>The Batch statement</returns>
```
jorgebay: It should be nice to add:
```xml
<exception cref="System.ArgumentOutOfRangeException">Thrown when trying to add more than <c>short.MaxValue</c> Statements</exception>
```

axelbodo:
I used ushort as server deserializes the 2 byte length field as ushort, so the server is able to handle up to ushort.MaxValue child statement. Do I change  it anyway to short?

axelbodo:
I used ushort intentionally as per the server side unmarshalling code. It's ok to put a stricker constraint on batch statement count with short.MaxValue. Do I change it anyway to short?

jorgebay:
You should use `short` everywhere.

axelbodo:
Ok, changed to short.

jorgebay:
I've created [CSHARP-360](https://datastax-oss.atlassian.net/browse/CSHARP-360) to support big endian platforms.

jorgebay:
Squashed commit into 729c4fbd5e447faa8904c5228788069b31c29380 and merged!

PR #146, opened 2015-10-12
Fixes for some brittle tests

maxwellb:
Comments at https://github.com/maxwellb/csharp-driver/issues/1

This commit hopes to reduce false negatives in 8 integration tests.


maxwellb:
I'll close this to reduce drift, and resubmit if there continues to be quirky test runs. Thanks for the upstream updates!

PR #148, opened 2015-10-14
initiating [short] protocol field fix.

axelbodo:
In some earlier bugfix it turned out that the [short] noted protocol field has serialized with different semantic in the driver (according to native_protocol_v3.spec). The protocol sometimes interpretes it as signed integer (mainly to denote null value), however it can be identified in the driver as well.

### DIFF
--- a/src/Cassandra/Requests/BatchRequest.cs
+++ b/src/Cassandra/Requests/BatchRequest.cs
@@ -93,6 +93,8 @@ namespace Cassandra
                 wb.WriteBytesMap(Payload);
             }
             wb.WriteByte((byte) _type);
+            //TODO: it would be good if this, and other places where Int16 is used,  will be modified according to the cql binary protocol standard section 3. [short] 2 bytes unsigned integer
+            //TODO: however on some places [short] is interpreted as signed to denote null values.
             wb.WriteInt16((short) _requests.Count);
             foreach (var br in _requests)
             {


jorgebay:
Negative values for null and unset are handled in the driver. See `FrameWriter.WriteBytes()`, `FrameReader` and [`OutputRows`](https://github.com/datastax/csharp-driver/blob/master/src/Cassandra/Outputs/OutputRows.cs#L73) for example.

In the case you mention in this pull request, it is not possible to send a batch with a negative value of statements to represent no items, and `_requests.Count` is always positive.

Remember that you can use [JIRA](https://datastax-oss.atlassian.net/browse/CSHARP/issues) to post issues or the [mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!forum/csharp-driver-user) for discussions.

axelbodo:
yes, you are right, what I wanted to solve regarding to my earlier #143/144 request is in case of batch count and list count those fields are interpreted as insigned and not as signed so I just want to keep preciseity, I mean form the servers point of view 32768 is a valid count for collection types and for batch statement count, however according to your request I changed it back to shourt.MaxValue. I've tested in the phase when it was on my machine an ushort (however because of the bit compatibility it doesn't matter), and the server was able to handle up to 65535 batch statement, and up to 65535 element of collection type. What I wanted to reach, that driver should provade the same limits as the server, and let users/developers choose to use it or not, depending on what load is put on the server cluster.

PR #149, opened 2015-10-14
[CSHARP-286] Tests for modernized schema tables

kishkaru:


jorgebay:
Merged!

Remember that you can use `TestClusterManager.CreateNew()` to manually create the cluster. 
Anyway, in the future we can divide the metadata tests into short (related to table/func/indexes/... schema changes) that reuses the test cluster instance and long (related to topology changes).

PR #150, opened 2015-10-14
[CSHARP-359] Tests for clustering order representation

kishkaru:


src/Cassandra.IntegrationTests/Core/MetadataTests.cs
```diff
@@ -899,5 +899,43 @@ public void GetMaterializedView_Should_Refresh_View_Metadata_Via_Events()
...
+                      description text,
+                      price double,
+                      tags frozen<set<text>>,
+                      PRIMARY KEY(id, description, price)
+                      ) WITH COMPACT STORAGE
+                      AND CLUSTERING ORDER BY (description ASC, price DESC)";
+            session.Execute(cql);
+
+            var tableMeta = cluster.Metadata.GetKeyspace(keyspaceName).GetTableMetadata(tableName);
+            Assert.IsFalse(tableMeta.ColumnsByName["description"].IsReversed);
```
jorgebay: `IsReversed` is an internal property used for C* < 3.0, not part of the public API.
You should use `ClusteringKeys` that is an array of tuples containing the column and the clustering order, something like:
```csharp
CollectionAssert.AreEqual(new[] { "description", "price" }, table.ClusteringKeys.Select(c => c.Item1.Name));
CollectionAssert.AreEqual(new[] { SortOrder.Ascending, SortOrder.Descending }, table.ClusteringKeys.Select(c => c.Item2));
```

PR #153, opened 2015-10-21
Fix linked list implementation in HashedWheelTimer

GoldenCrystal:
The previous implementation of the Remove method could leave the linked list in an invalid state, which we noticed when using the C# driver intensively for benchmarking Cassandra. (The exception with the message "Next is null but it is not the tail" did get thrown)
This PR should fix the incorrect behavior we encountered.

### DIFF
--- a/src/Cassandra.Tests/TimeoutTests.cs
+++ b/src/Cassandra.Tests/TimeoutTests.cs
@@ -57,6 +57,48 @@ namespace Cassandra.Tests
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


jorgebay:
Great find! 

Looks good to me, sign the CLA and I'll merge it.

jorgebay:
Rebased into 2fc382a80168dedad42d2b3313ad96d2e838d3fe and merged, thanks!

PR #154, opened 2015-10-21
Buffer pooling

jorgebay:
Included 2 classes from `Microsoft.IO.RecyclableMemoryStream` under MIT, left copyright headers in files and added a notice file with the copyright information.

All requests in the write queue are written to a single stream. 
On the read path, the body from all frames received in the incoming buffer are stored in a single stream and then read in order. Instead of using 1 task to invoke each callback, 1 task is used to invoke all complete messages received.

Other changes
- Introduced a big endian converter `BeConverter`
- Renamed `BEBinaryWriter` and `BEBinaryWriter` into `FrameWriter` and `FrameReader`.
- Removed `ListBackedStream`.
- Introduced a thread-local reusable buffer for parsing row values (where allowed).

It's chained with #151.

There other improvements that can be made, but I think it is out of scope for now (we can create tickets for it):
- Improve the `IFrameCompressor` interface to support a count/length parameter to avoid extra copies.
- Improve TypeCodec to reuse buffers and support count/length parameter on decoding methods (see `DecodeVarchar()` for example).


jorgebay:
Calling @ocoanet @alprema @LukeTillman for review!

src/Cassandra/Connection.cs
```diff
@@ -600,87 +676,107 @@ public OperationState Send(IRequest request, Action<Exception, AbstractResponse>
...
-                //The request was not written, clear it from pending operations
-                RemoveFromPending(streamId);
-                //Callback with the Exception
-                state.InvokeCallback(ex);
+                //nothing to write
+                Interlocked.Exchange(ref _isWriteQueueRuning, 0);
+                return;
             }
+            //Write and close the stream when flushed
+            _tcpSocket.Write(stream.GetBuffer(), (int)stream.Length, () => stream.Close());
```
ocoanet: I think that `GetBuffer` should be avoided on a `RecyclableMemoryStream` because it generates large buffer allocations. It would be better to pass the stream to `TcpSocket.Write` and remove the `onBufferFlush` argument.

src/Cassandra/TcpSocket.cs
```diff
@@ -389,39 +392,47 @@ protected virtual void OnClosing()
...
             {
                 OnError(new SocketException((int)SocketError.Shutdown));
+                OnWriteFlushed();
                 return;
             }
             if (_sendSocketEvent != null)
             {
-                _sendSocketEvent.SetBuffer(buffer, 0, buffer.Length);
-                bool willRaiseEvent = false;
+                _sendSocketEvent.SetBuffer(buffer, 0, length);
```
ocoanet: I doubt it is a good pattern to generate a large buffer and replace the `SocketAsyncEventArgs` buffer on every send. Instead, the outgoing stream could copied into the `SocketAsyncEventArgs` buffer using `Stream.Read`. Multiple calls to `Read` and `SendAsync` might be required to send the whole stream though.

src/Cassandra/Outputs/OutputRows.cs
```diff
@@ -17,33 +17,30 @@
...
     internal class OutputRows : IOutput
     {
-        public readonly int RowLength;
-        private readonly Guid? _traceId;
-        private RowSetMetadata _metadata;
-        private byte _protocolVersion;
+        private readonly int _rowLength;
+        private readonly RowSetMetadata _metadata;
+        private readonly byte _protocolVersion;
+        private static readonly ThreadLocal<byte[]> ReusableBuffer = new ThreadLocal<byte[]>(() => new byte[16]);
```
ocoanet: The `OutputRows` already contains a lot of data (rows and columns) and it is instantiated only once for each response. The buffer could be a non static field instead of a `ThreadLocal`.

src/Cassandra/Connection.cs
```diff
@@ -453,113 +471,171 @@ private void ReadHandler(byte[] buffer, int bytesReceived)
...
+                foreach (var cb in operationCallbacks)
+                {
+                    cb(stream);
+                }
+                stream.Dispose();
+            }, CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default);
             return true;
-            //There isn't enough data to read the whole frame.
-            //It is already buffered, carry on.
         }
```
ocoanet: Ok, I know it is somewhat subjective but I think this method suffers from excessive complexity and excessive delegate usage. I would try to split the code in small methods and I would replace the `operationCallbacks` list with a list of `FrameToParse` that could be processed inside the `Task` action.
```cs
private class FrameToParse
{
    public readonly FrameHeader Header;
    public readonly Action<Exception, Response> StateCallback;
    // 
}
// replaces CreateCallbackAction, invoked inside the Task action,
private void ProcessFrame(FrameToParse frame, MemoryStream stream)
{
    //...
}
```

src/Cassandra/TcpSocket.cs
jorgebay: Remember that large buffers are pooled, meaning that few large buffers are created and then reused for writing across all connections.
`GetBuffer()` writes small chunks into a "final" large buffer that is used to send. It is indeed the same as writing to 1 buffer (the `SocketAsyncEventArgs` buffer) per connection, while avoiding the multiple system calls in case the amount of messages is large.

I think I found a solution that would avoid the block copying from stream to final buffer,  using [SocketAsyncEventArgs.BufferList](https://msdn.microsoft.com/en-us/library/system.net.sockets.socketasynceventargs.bufferlist(v=vs.110).aspx). I will try that when using the `SocketAsyncEventArgs` interface.

src/Cassandra/Outputs/OutputRows.cs
jorgebay: We need a buffer per thread, as the same thread can not be reading an output at the same time. There can be thousands of `OutputRows` instances at any given time, having a static thread local saves a lot of allocations.

jorgebay:
I've implemented the `TcpSocket.Write()` method using array segments (to the pooled blocks) to avoid block copies for the `SocketAsyncEventArgs` interface.

Moved the logic to invoke read callbacks out of `ReadParse()` to a new `InvokeReadCallbacks()` method.

alprema:
Hi,

First thank you for taking care of this, I can't wait to see how fast it can go :)

To be able to put numbers on those improvements I just created a repo with a [benchmark tool](https://github.com/Abc-Arbitrage/cassandra-csharp-benchmark) for read and writes inspired from a real life time-series application and ran it against the CSHARP-270 branch.

At the current version (31ce7146a95f6c3aaa7ac7c7374936bde3d507e0) the read benchmark is similar to the 2.7.2, so no regression but no improvement yet (did not test writes yet).

The benchmark is easy to run but don't hesitate to ask if you want me to run it against our cluster.

jorgebay:
I'm getting around twice the amount of ops per second, in an insert-based scenario: ~70% more throughput with just CSHARP-244 and an extra ~30% with CSHARP-270.

I will continue cover insert+select scenario and share my benchmarks tomorrow.

src/Cassandra/Connection.cs
```diff
@@ -600,87 +684,107 @@ public OperationState Send(IRequest request, Action<Exception, AbstractResponse>
...
-                    SendQueueProcessItem(state);
-                    return;
-                }
-                //Start a new task using the TaskScheduler for writing
-                Task.Factory.StartNew(() => SendQueueProcessItem(state), CancellationToken.None, TaskCreationOptions.None, _writeScheduler);
+                //there is another thread writing to the wire
+                return;
             }
-            else
+            if (useInlining)
```
jorgebay: Consider removing inlining (reusing the calling thread to start the writing operation) based on @alprema input. In this case, it would only improve latency for low concurrency scenarios but it doesn't make much sense to optimize for those scenarios.

jorgebay:
Thanks @ocoanet and @alprema for the input.

I've found a race condition that could cause the write queue to stop process more items (until next message arrives) and I've fixed it: https://github.com/datastax/csharp-driver/commit/b50454789129c5169a7e1494882e9f2f685198af#diff-1894c6df6681cb3e1204f1d16a5435faR755

About benchmarking, we are creating a benchmarking project (SUT) per driver, you can find the C# one here: https://github.com/riptano/csharp-driver-sut

The SUT is intended to be used as an http wrapper (WIP), but I've created a flag to run it as an script.
You can compare the results from master vs 2.7 branch by running:

```bash
> build master -c your-contact-point -h N
```

```bash
> build 2.7 -c your-contact-point -h N
```

You can tune the max outstanding requests and the level of parallelism. Use `--help` for more information:

```bash
> build master --help
```

PR #155, opened 2015-10-26
Handle exceptions in HostConnectionPool

alprema:
Some tasks did not access/log the `Exception` member and caused the exceptions to be handled by the finalizer. An application handling the `TaskScheduler.UnobservedTaskException` event would get exceptions that were left as unhandled even though they were.
(I'm not 100% sure about the location of the error handling regarding the pool semaphore).

### DIFF
--- a/src/Cassandra/HostConnectionPool.cs
+++ b/src/Cassandra/HostConnectionPool.cs
@@ -206,6 +206,10 @@ namespace Cassandra
                 else
                 {
                     _poolModificationSemaphore.Release();
+                    if (t.Exception != null)
+                    {
+                        t.Exception.Handle(e => true);
+                    }
                     Logger.Info("Reconnection attempt to host {0} failed", _host.Address);
                     _host.SetDown();
                 }
@@ -330,6 +334,10 @@ namespace Cassandra
             {
                 if (t.Status != TaskStatus.RanToCompletion)
                 {
+                    if (t.Exception != null)
+                    {
+                        t.Exception.Handle(e => true);
+                    }
                     //The first connection failed
                     //Wait for all to complete
                     return allCompleted;
@@ -370,6 +378,10 @@ namespace Cassandra
                     _connections.Add(t.Result);   
                 }
                 _poolModificationSemaphore.Release();
+                if (t.Exception != null)
+                {
+                    Logger.Error("Error during new connection attempt", t.Exception);
+                }
             }, TaskContinuationOptions.ExecuteSynchronously);
             return true;
         }


jorgebay:
Nice!
About the semaphore, I think doing the exception handling after releasing, to release it as soon as possible.

src/Cassandra/HostConnectionPool.cs
```diff
@@ -218,6 +218,8 @@ internal void AttemptReconnection()
                 }
                 else
                 {
+                    if (t.Exception != null)
+                        Logger.Error("Error during reconnection attempt", t.Exception);
```
jorgebay: Reconnection attempts that failed are not necessarily of log level `Error`: the `ReconnectionPolicy` marks the delay of the reconnection even if its notified as down via events.
We could add the exception string below, something like:
```csharp
Logger.Info("Reconnection attempt to host {0} failed: {1}", _host.Address, t.Exception);
```

jorgebay: Now that I think of it, it is not necessary to include the exception in the logs again. It was already logged on `CreateConnection()`.

src/Cassandra/HostConnectionPool.cs
```diff
@@ -218,6 +218,8 @@ internal void AttemptReconnection()
...
                     _poolModificationSemaphore.Release();
                     Logger.Info("Reconnection attempt to host {0} failed", _host.Address);
                     _host.SetDown();
```
jorgebay: Should handle `t.Exception` here also? `AttemptReconnection()` is `void` and it wont be possible to do it anywhere else.

src/Cassandra/HostConnectionPool.cs
```diff
@@ -343,6 +345,8 @@ internal Task<Connection[]> MaybeCreateCorePool()
             {
                 if (t.Status != TaskStatus.RanToCompletion)
                 {
+                    if (t.Exception != null)
+                        t.Exception.Handle(e => true);
```
jorgebay: small nit: can we add `{` and `}` for the statement, even if its a one-liner, like the rest of the file.

jorgebay:
Apart from the minor nits, looks good to me.

alprema:
I made all the suggested changes except for a potential missing `t.Exception` handling in `AttemptReconnection()` which I'm not sure I understood since it is handled in that method.

jorgebay:
It's OK, you handled the exception on `AttemptReconnection()`.
Looks good to me.

PR #158, opened 2015-11-02
Use LocalQuorum as default consistency

jorgebay:
Also
- Default timeout 12secs CSHARP-362
- Tcp no delay true by default CSHARP-259

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/BasicTypeTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/BasicTypeTests.cs
@@ -298,6 +298,16 @@ namespace Cassandra.IntegrationTests.Core
             Assert.Null(rs.Info.QueryTrace);
         }
 
+        /// Tests that the default consistency level for queries is LOCAL_QUORUM
+        /// 
+        /// LocalQuorum_Is_Default_Consistency tests that the default consistency level for all queries is LOCAL_QUORUM. It performs
+        /// a simple select statement and verifies that the result set metadata shows that the achieved consistenc level is LOCAL_QUORUM.
+        /// 
+        /// @since 3.0.0
+        /// @jira_ticket CSHARP-353
+        /// @expected_result The default consistency level should be LOCAL_QUORUM
+        /// 
+        /// @test_category consistency
         [Test]
         public void LocalQuorum_Is_Default_Consistency()
         {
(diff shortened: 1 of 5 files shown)


src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
```diff
@@ -178,5 +178,33 @@ public void Should_Throw_NoHostAvailableException_When_All_Hosts_Down()
...
+        ///
+        /// @since 3.0.0
+        /// @jira_ticket CSHARP-362
+        /// @expected_result A TimeoutException should be raised after 12 seconds.
+        ///
+        /// @test_category connection:timeout
+        [Test]
+        public void Should_Abort_After_Default_ReadTimeoutMillis()
+        {
+            var testCluster = TestClusterManager.GetNonShareableTestCluster(2, 1, true, false);
```
jorgebay: Nit: use `TestClusterManager.CreateNew()` to create test clusters, the other methods like `GetNonShareableTestCluster()` are deprecated.

src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
```diff
@@ -178,5 +178,33 @@ public void Should_Throw_NoHostAvailableException_When_All_Hosts_Down()
...
+            var builder = Cluster.Builder().AddContactPoint(testCluster.InitialContactPoint);
+            using (var cluster = builder.Build())
+            {
+                var session = cluster.Connect();
+                //warmup
+                TestHelper.Invoke(() => session.Execute("SELECT key FROM system.local"), 10);
+                testCluster.PauseNode(1);
+                testCluster.PauseNode(2);
+                Assert.Throws<TimeoutException>(() => session.Execute("SELECT key FROM system.local"));
```
jorgebay: `TimeoutException` is thrown at when the amount of time passed and the `Task` did not complete and can be tuned using `QueryAbortTimeout` (default: 20secs).

To test `SocketOptions.ReadTimeoutMillis`,  you should set a lower readtimeout (something that multiplied by the amount of hosts, will be lower than `QueryAbortTimeout`. This situation is already covered in 

- `Should_Throw_OperationTimedOutException_When_Retry_Is_False`: Without retrying the next host, it throws an `OperationTimedOutException`.
- `Should_Throw_NoHostAvailableException_When_All_Hosts_Down`: Retries to the next host, but also times out, it throws a `NoHostAvailableException` with `OperationTimedOutException` as (child) inner errors per host.

PR #161, opened 2015-11-04
add ISession.ChangeKeyspaceAsync

collinsauve:
I've added an `ExecuteAsync(string cqlQuery)` overload, as well as an async version of `ChangeKeyspace`.

`ExecuteAsync` I can implement as an extension method, but `ChangeKeyspaceAsync` cannot as `ISession.set_Keyspace` is not exposed.

### DIFF
--- a/src/Cassandra/ISession.cs
+++ b/src/Cassandra/ISession.cs
@@ -85,6 +85,12 @@ namespace Cassandra
         void ChangeKeyspace(string keyspaceName);
 
         /// <summary>
+        /// Switches to the specified keyspace asynchronously.
+        /// </summary>
+        /// <param name="keyspaceName">Case-sensitive name of keyspace to be used.</param>
+        /// <exception cref="InvalidQueryException">When keyspace does not exist</exception>
+        Task ChangeKeyspaceAsync(string keyspaceName);
+        /// <summary>
         ///  Creates new keyspace in current cluster.        
         /// </summary>
         /// <param name="keyspaceName">Case-sensitive name of keyspace to be created.</param>
--- a/src/Cassandra/Session.cs
+++ b/src/Cassandra/Session.cs
@@ -125,6 +125,16 @@ namespace Cassandra
         }
 
         /// <inheritdoc />
+        public Task ChangeKeyspaceAsync(string keyspace)
+        {
+            if (Keyspace != keyspace)
+            {
+                return ExecuteAsync(new SimpleStatement(CqlQueryTools.GetUseKeyspaceCql(keyspace)));
+            }
+            return TaskHelper.Completed;
+        }
+
+        /// <inheritdoc />
         public void CreateKeyspace(string keyspace, Dictionary<string, string> replication = null, bool durableWrites = true)
         {
             WaitForSchemaAgreement(Execute(CqlQueryTools.GetCreateKeyspaceCql(keyspace, replication, durableWrites, false)));


src/Cassandra/Tasks/TaskHelper.cs
```diff
@@ -195,6 +195,18 @@ private static Exception PreserveStackTrace(Exception ex)
         /// <summary>
         /// Smart ContinueWith
         /// </summary>
+        public static Task Continue<TIn>(this Task<TIn> task, Action<Task<TIn>> next)
```
jorgebay: This method is not necessary, it is used by `ChangeKeyspaceAsync()` but it can be safely removed as `task.Continue(rs => Keyspace = keyspace);` returns a `Task<string>` which is still a `Task`.

jorgebay:
I don't know if we should add the `ExecuteAsync(string)` overload method, the [java driver has it](http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/Session.html#executeAsync(java.lang.String)) but it was never added to the C# driver.
I think having the sync method `ExecuteAsync(string)` is good for newbies but we should encourage users to use statements (query / queryid + parameters + consistency + ...) instead.
I will let @al3xandru call this one. 

About the `ChangeKeyspaceAsync()` it makes sense for me to have it as we have a blocking counter part, even though users should not be changing keyspaces regularly within a session.

About the code, besides the small comment to avoid create a new `Continue()` method, looks good to me.

collinsauve:
Removed the `TaskHelper.Continue` overload.

collinsauve:
@al3xandru can this be merged?  If `ExecuteAsync(string)` is undesirable I can create a separate pull request with just `ChangeKeyspaceAsync` - that is the method that is truly necessary since it can't be implemented externally.

al3xandru:
@collinsauve firstly thanks for the PR (and sorry for completely missing the notifications about the questions)

As both of these are API additions, I'd like to better understand where/how they'd be useful.

1. while `ExecuteAsync` is present in a similar form in the Java driver, I find its usage really limited and thus I'd probably suggest not extending the API (it's also easily achievable with the existing API)
2. in the same vein in the above, I'm not sure I see where an async keyspace change is useful (and if I'm not mistaking that could also be done using the existing API if really needed)

collinsauve:
Hi @al3xandru, thanks for getting back to me.

1. The `ExecuteAsync(string)` is as useful as `Execute(string)`, except in an async context.  You are correct that this is easily achievable with the existing API.
1. The `ChangeKeyspaceAsync(string)` is necessary when you which to change keyspaces in an async context.  Unlike the above, this cannot be achieved with the existing API since there is no public setter on `ISession.Keyspace`.  

If it would help I can rework/resubmit the PR with only `ChangeKeyspaceAsync(string)`.

al3xandru:
@collinsauve I was under the impression that all `ChangeKeyspace` does is execute a `USE keyspace` statement. A quick check confirmed the first part, but also the fact the the `ISession` implementation maintains a reference to the keyspace.

Going back for a second to my previous question, could you give me the shortest example of where changing the keyspace asynchronously would be used? (basically what I'm trying to answer is how this method would be really used as to me it seems a bit weird in the sense that it can non-deterministically change the keyspace upon which non-fully qualified queries are getting executed---on the other hand it's kind of late on my side and I might be missing something)

collinsauve:
We only need to use it in a very small part of our app - during deployments we execute schema changes in serial.   Since they are serial, we are not concerned with parallel queries being interfered with.   

Perhaps I am misunderstanding, but are you implying that the non-Async `ChangeKeyspace` does not have a similar problem of interfering with parallel queries? 

Since [async is contagious](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx), we do almost everything in an async context.

> As you convert synchronous code to asynchronous code, you’ll find that it works best if asynchronous code calls and is called by other asynchronous code—all the way down (or “up,” if you prefer). Others have also noticed the spreading behavior of asynchronous programming and have called it “contagious” or compared it to a zombie virus. Whether turtles or zombies, it’s definitely true that asynchronous code tends to drive surrounding code to also be asynchronous. This behavior is inherent in all types of asynchronous programming, not just the new async/await keywords.

I've actually already modified some of our infrastructure to allow synchronous blocking functions higher in our stack in order to deal with [the lack of an async Connect function](https://datastax-oss.atlassian.net/browse/CSHARP-370). 

I could do the same for our schema changes.  Also this conversation has given me an idea, perhaps.  We could fully-qualify our schema changes.

jorgebay:
I'm +1 for adding `ChangeKeyspaceAsync()` even though is just a wrapper to `ExecuteAsync(new SimpleStatement("USE ksname"))`.

I'm -1 for `ExecuteAsync(string)`: We have to point users to use statements, specially `BoundStatement` and having an additional overload that accept a string is not a good idea. I think the reason that `Execute(string)` exists is to give users a "quick start" method to use.

collinsauve:
Modified PR to drop `ExecuteAsync(string)` and only include `ISession.ChangeKeyspaceAsync`.

For what its worth, I've changed our code to fully-qualify table names where we need to change keyspaces.

al3xandru:
Based on @jorgebay's vote and your explanations I'm going +1 too (with the only addition that it's still not extremely clear to me how the async keyspace change is actually useful, but I'm OK with going with the assumption I'm missing something)

src/Cassandra/Session.cs
```diff
@@ -107,6 +107,17 @@ public void ChangeKeyspace(string keyspace)
         }
 
         /// <inheritdoc />
+        public Task ChangeKeyspaceAsync(string keyspace)
+        {
+            if (Keyspace != keyspace)
+            {
+                var task = ExecuteAsync(new SimpleStatement(CqlQueryTools.GetUseKeyspaceCql(keyspace)));
+                return task.Continue(rs => Keyspace = keyspace);
```
jorgebay: You can safely remove the setting of the keyspace as this is done by `RequestExecution` class, after a [`set_keyspace` response](https://git-wip-us.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=doc/native_protocol_v4.spec;hb=refs/heads/cassandra-2.2).
This can be `return ExecuteAsync(...`

It is also not necessary on the [sync counterpart](https://github.com/datastax/csharp-driver/pull/161/files#diff-5137a268854a1ede5dc7e5c2ef5c5d3dL105), but was left there from a refactoring

collinsauve:
If it is the case that the implementation doesn't need to set the Session.Keyspace, then this PR isn't really necessary as changing the keyspace asynchronously can already be performed externally using the existing API.

collinsauve:
I've pushed an additional commit that removes the line that sets the `Session.Keyspace`.

joao-r-reis:
Hi @collinsauve, I've noticed this PR hasn't had any activity in a while... Are you still interested in going forward with this PR? If so, please update your branch to force a build on travis.

src/Cassandra/ISession.cs
```diff
@@ -84,6 +84,12 @@ public interface ISession: IDisposable
         /// <exception cref="InvalidQueryException">When keyspace does not exist</exception>
         void ChangeKeyspace(string keyspaceName);
 
+        /// <summary>
+        /// Switches to the specified keyspace asynchronously.
+        /// </summary>
+        /// <param name="keyspaceName">Case-sensitive name of keyspace to be used.</param>
+        /// <exception cref="InvalidQueryException">When keyspace does not exist</exception>
+        Task ChangeKeyspaceAsync(string keyspaceName);
```
joao-r-reis: Although this change makes sense since we already have a synchronous version of this method, it's still technically a breaking change and I'm worried that there are users out there creating fake classes that implement `ISession` for their tests. In the past there were some methods that were added to interfaces like `IStatement` but it's much less likely that users are creating fakes for these interfaces than `ISession` since this is the main interface of the driver.

collinsauve:
> Although this change makes sense since we already have a synchronous version of this method, it's still technically a breaking change and I'm worried that there are users out there creating fake classes that implement ISession for their tests.

If that's a concern then I only see a couple of ways forward:
1. [Interface segregation](https://en.wikipedia.org/wiki/Interface_segregation_principle)
2. Extension methods
3. Don't add this at all

Allow me to bring up an old conversation on this PR:

@jorgebay:

> You can safely remove the setting of the keyspace as this is done by RequestExecution class, after a set_keyspace response.

@collinsauve:

> If it is the case that the implementation (of a given ChangeKeyspaceAsync function) doesn't need to set the Session.Keyspace, then this PR isn't really necessary as changing the keyspace asynchronously can already be performed externally using the existing API.

After learning that, my work-around is just to do this externally.  I've been using this solution for a couple of years now.  I kept this PR because I believe the interface should keep async counterparts to its sync methods, but you will have to decide.  

P.S. Maybe if C# 8 ends up getting default interface methods (and old CLRs support it) that could be helpful.

src/Cassandra/ISession.cs
collinsauve: Responded [here](https://github.com/datastax/csharp-driver/pull/161#issuecomment-473984902)

joao-r-reis:
> > Although this change makes sense since we already have a synchronous version of this method, it's still technically a breaking change and I'm worried that there are users out there creating fake classes that implement ISession for their tests.
> 
> If that's a concern then I only see a couple of ways forward:
> 
>     1. [Interface segregation](https://en.wikipedia.org/wiki/Interface_segregation_principle)
> 
>     2. Extension methods
> 
>     3. Don't add this at all
> 

I generally agree with the first option but I don't think that is a solution here to prevent a breaking change. 

I'm +1 for 2. I don't generally like extension methods too much but I think that this is a scenario where an extension method fits nicely.

> P.S. Maybe if C# 8 ends up getting default interface methods (and old CLRs support it) that could be helpful.

Yeah default interface methods would solve this issue but since they might require a change to the runtime (last I checked) I'm not too sure if the target frameworks that this driver supports will support that feature (especially the `net452` target).

collinsauve:
Okay, I'll be fine with whatever you decide then.  I'm willing to re-write as an extension method, or close the PR if that's what you want to go with.  If you want interface segregation that's sounds like something I'd let you design :) LMK

joao-r-reis:
Seeing that you found a work-around for this issue and that you are okay with closing this for now, we can wait for the next major version of the driver to introduce this method. Thanks for the nice discussion regardless!

PR #162, opened 2015-11-06
Integration test fixes and stabilizations

kishkaru:


src/Cassandra.IntegrationTests/Mapping/Tests/Single.cs
```diff
@@ -193,14 +171,11 @@ public void Single_MissingPartitionKey()
             catch (InvalidQueryException e)
             {
                 string expectedErrMsg = null;
-                if (_session.BinaryProtocolVersion < 4)
-                {
-                    expectedErrMsg = "No indexed columns present in by-columns clause with Equal operator";
-                }
+                if (_session.BinaryProtocolVersion < 3)
```
maxwellb: The message difference here is not caused by binary protocol version - rather Cassandra version 2.1.3 changed the message.

PR #163, opened 2015-11-06
ConnectAsync

collinsauve:
For https://datastax-oss.atlassian.net/browse/CSHARP-370

Adds a `ConnectAsync` function.  I'm not entirely sure that cancellation/timeouts are working.  Will continue iterating next week.

Note that this uses .Net 4.5 (async/await) and C# 6 (await in catch blocks) functionality.  I'm actually surprised it passed your CI.


jorgebay:
We should still support NET 4.0.

Unit tests on travis passes because it uses Mono 4.0 which supports C# 6.0.

collinsauve:
Closing PR until I can rewrite with .Net 4.0 support.

PR #168, opened 2015-11-23
Use a private field for inFlight counting

jorgebay:
Avoid using `ConcurrentDictionary.Count` as it causes thread contention.

Do you have some time to review it @alprema ?

### DIFF
--- a/src/Cassandra/Connection.cs
+++ b/src/Cassandra/Connection.cs
@@ -69,6 +69,7 @@ namespace Cassandra
         private volatile byte _frameHeaderSize;
         private MemoryStream _readStream;
         private int _isWriteQueueRuning;
+        private int _inFlight;
         /// <summary>
         /// The event that represents a event RESPONSE from a Cassandra node
         /// </summary>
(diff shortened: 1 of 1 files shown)


alprema:
Hi,

It seems good but I am concerned about the log at [line 257](https://github.com/datastax/csharp-driver/pull/168/files#diff-1894c6df6681cb3e1204f1d16a5435faR257):
```cs
_logger.Info("Canceling pending operations {0} and write queue {1}", _pendingOperations.Count, _writeQueue.Count);
```

`_pendingOperations.Count` will be evaluated wether the log is at info level or not, meaning having another fight for the lock for nothing.

This maybe should be part of another issue where we'd go through all the logs to find similar effects.

jorgebay:
We can address it now using `_inFlight` field now, it only happens when the connection is closing so it's not a big deal.

I'm concerned about https://github.com/datastax/csharp-driver/commit/d40c4397ae643a0b505ee0224f1ea98a85aca0f8#diff-1894c6df6681cb3e1204f1d16a5435faL719 that occurs when the pressure is high.

I'll fix both of them.

jorgebay:
Pushed the changes based on @alprema comments and removed the call to `Count` when no stream ids are available.

Let me know if looks good to you.

alprema:
Looks good to me.

jorgebay:
Thanks @alprema for your time!

PR #175, opened 2015-12-03
Enable heartbeat by default

jorgebay:


jorgebay:
Contains improved `PoolingOptions` class summary, where the defaults per Cassandra version are detailed.

PR #178, opened 2015-12-11
[CSHARP-377] Tests for empty RowSet in VOID results

kishkaru:


jorgebay:
To test `VOID` result, you should use a insert/update query as query requests are always responded with `ROWS` result (even for 0 rows).
The same assertions should be valid with a insert query.

kishkaru:
Updated to use an INSERT statement.

PR #182, opened 2015-12-23
HostConnectionPool: Schedule next reconnection attempt without checking delay passed

jorgebay:
Removed pool semaphore.
Pool method to create first connection only and other to increase to core and max.
Reconnection test with multiple nodes: Created a new test fixture of reconnection tests for the category "short". Moved some existing test from PoolTests to the new test fixture.


jorgebay:
Rebased, solved conflicts against master.

PR #183, opened 2015-12-23
Add WithMaxProtocolVersion method to Builder

jorgebay:


landim:
:+1: as soon as CI passes

PR #184, opened 2015-12-30
Add support for CONTAINS keyword.

adamcogx:
Currently writing something like:

    songs.Where(x => x.Tags.Contains("Rock"))

throws an exception, when one would expect it to create CQL:

    SELECT * FROM table WHERE tags CONTAINS ?

I have fixed it so that it does work.

Also, I didn't affect:

    List<string> selectedArtists = new List<string> {
				"Artist 1",
				"Artist 2"
			};
    songs.Where(x => selectedArtists.Contains(x.Artist))


adamcogx:
Hmm.. Guess I lied.  I will look into fixing the issues.

adamcogx:
I updated it to support Dictionary / Map, and added tests for scenarios.  This support only works with basic types (no support for complex types, not even sure if that is possible at a Cassandra level...)

jorgebay:
Closing this in favour of #185 that includes both `CONTAINS` and `CONTAINS KEY` support.

PR #185, opened 2016-01-03
Key index support

kadamgreene:


jorgebay:
Looks really good, just a minor nits: 
- Use 4 spaces for indentation, new line for open curly brackets (`{`) and `else` statements. [Following C# coding conventions](https://msdn.microsoft.com/en-us/library/ff926074.aspx).
- Avoid supporting key indexes in the `Cassandra.Data.Linq.SecondaryIndex` as it is obsolete. Just supporting it on `Cassandra.Mapping.Attributes` is enough.
- Your efforts to avoid breaking changes are great (ie: multiple constructor definitions), there is just one binary breaking change though: existing `ColumnMap.WithSecondaryIndex()` should be kept. You can add a new public method `ColumnMap.WithSecondaryIndex(bool)`.

I know that Visual Studio is really lame to switch between tabs and space indentations but we should not allow different conventions as it messes up with git history.

jorgebay:
Closing this pull request for no activity.

Feel free to rebase it against master and submit a new pull request.

PR #190, opened 2016-02-05
Use wrapped stream for frame decompressing

jorgebay:


landim:
:+1: Looks good to me

PR #192, opened 2016-02-16
add overloads with ttl parameter to Insert functions on ICqlWriteClient and…

collinsauve:
For https://datastax-oss.atlassian.net/browse/CSHARP-395 (Mapper: Support TTL and timestamp)

When trying to implement this, I was looking at what `CqlQueryOptions` is used for.  These properties are typically used to decorate an `IStatement`.  In this case I believe the ttl is only put into the actual query, rather than serializing it into query_parameters.  

Also considering that:
* `CqlQueryOptions` is mainly **write-only**
* Currently the only way to read from it is to call `CopyOptionsToStatement(IStatement)`
* Ttl does not belong on the statement

I would need to break this quasi-write-only idiom in order to read from `Mapper` and `CqlBatch`.

If you still feel ttl should be on `CqlQueryOptions`, let me know and I'll file a new PR.

### DIFF
--- a/src/Cassandra/Mapping/CqlBatch.cs
+++ b/src/Cassandra/Mapping/CqlBatch.cs
@@ -36,8 +36,14 @@ namespace Cassandra.Mapping
 
         public void Insert<T>(T poco, bool insertNulls, CqlQueryOptions queryOptions = null)
         {
-            Insert(false, insertNulls, poco, queryOptions);
+            Insert(false, insertNulls, poco, queryOptions, null);
         }
+
+        public void Insert<T>(T poco, bool insertNulls, int? ttl, CqlQueryOptions queryOptions = null)
+        {
+            Insert(false, insertNulls, poco, queryOptions, ttl);
+        }
+
 
         public void InsertIfNotExists<T>(T poco, CqlQueryOptions queryOptions = null)
         {
@@ -49,7 +55,12 @@ namespace Cassandra.Mapping
             Insert(true, insertNulls, poco, queryOptions);
         }
 
-        private void Insert<T>(bool ifNotExists, bool insertNulls, T poco, CqlQueryOptions queryOptions = null)
+        public void InsertIfNotExists<T>(T poco, bool insertNulls, int? ttl, CqlQueryOptions queryOptions = null)
+        {
+            Insert(true, insertNulls, poco, queryOptions, ttl);
+        }
+
+        private void Insert<T>(bool ifNotExists, bool insertNulls, T poco, CqlQueryOptions queryOptions = null, int? ttl = null)
         {
             var pocoData = _mapperFactory.PocoDataFactory.GetPocoData<T>();
             var queryIdentifier = string.Format("INSERT ID {0}/{1}", pocoData.KeyspaceName, pocoData.TableName);
@@ -58,7 +69,7 @@ namespace Cassandra.Mapping
             var values = getBindValues(poco);
             //generate INSERT query based on null values (if insertNulls set)
             object[] queryParameters;
-            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists);
+            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl);
 
             _statements.Add(Cql.New(cql, queryParameters, queryOptions ?? CqlQueryOptions.None));
         }
--- a/src/Cassandra/Mapping/ICqlWriteAsyncClient.cs
+++ b/src/Cassandra/Mapping/ICqlWriteAsyncClient.cs
@@ -33,6 +33,27 @@ namespace Cassandra.Mapping
         Task InsertAsync<T>(T poco, bool insertNulls, CqlQueryOptions queryOptions = null);
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


collinsauve:
This might be a dupe of #90, however the comments there by @jorgebay indicate that this is undesired on `IMapper`/`CqlQueryOptions` and we should use the linq component instead.  This contradicts [CSHARP-395](https://datastax-oss.atlassian.net/browse/CSHARP-395).

I will review how to do this using the linq component.  Perhaps this is not necessary.

src/Cassandra/Mapping/CqlBatch.cs
```diff
@@ -29,14 +29,14 @@ public CqlBatch(MapperFactory mapperFactory, CqlGenerator cqlGenerator)
             _statements = new List<Cql>();
         }
 
-        public void Insert<T>(T poco, CqlQueryOptions queryOptions = null)
+        public void Insert<T>(T poco, CqlQueryOptions queryOptions = null, int? ttl = null)
```
jorgebay: Creating a new parameter (even if its with a default value) is a binary break, we must expose it as an overload if we want to deliver it in a minor / patch version: http://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net

src/Cassandra/Mapping/CqlBatch.cs
```diff
@@ -29,14 +29,14 @@ public CqlBatch(MapperFactory mapperFactory, CqlGenerator cqlGenerator)
...
         }
 
-        public void Insert<T>(T poco, bool insertNulls, CqlQueryOptions queryOptions = null)
+        public void Insert<T>(T poco, bool insertNulls, CqlQueryOptions queryOptions = null, int? ttl = null)
         {
-            Insert(false, insertNulls, poco, queryOptions);
+            Insert(false, insertNulls, poco, queryOptions, ttl);
         }
 
         public void InsertIfNotExists<T>(T poco, CqlQueryOptions queryOptions = null)
```
jorgebay: We should also add an overload for `InsertIfNotExists<T>()` method.

jorgebay:
Good call for making this only available for the insert methods, as those relies on query generation. Update methods can include it within the query.

Apart from the methods overload comment, looks good to me.

I will create a separate ticket for timestamp support in the mapper, that do belong at `CqlQueryOptions` level.

collinsauve:
Updated PR with new overloads.  If I overload and add it as optional I get a R# warning that

> Method with optional parameter is hidden by overload

Not exactly a fatal warning, but thought I'd take the hint and not overload with optionals.  I'll defer to your judgement if these signatures are correct.  I'm fine reworking them if you'd like.

collinsauve:
> Good call for making this only available for the insert methods, as those relies on query generation. Update methods can include it within the query.

@jorgebay I only excluded it because currently we do not have a requirement to set TTL on update.  

Thinking about it more closely I'm actually not sure I agree that it should be excluded :) (although maybe I'll do another PR to keep this one simpler)

Those that are using the mapper for the poco->Cql generation may wish to set TTL on update.  This is my reasoning for wanting TTL on insert.

jorgebay:
> Thinking about it more closely I'm actually not sure I agree that it should be excluded :)

Remember that users can specify the query in the `Update()` method of the mapper.

```csharp
mapper.Update("UPDATE tbl SET x = ? WHERE id = ? TTL ?", x, id, ttl);
```

src/Cassandra/Mapping/Mapper.cs
```diff
@@ -468,6 +483,11 @@ public AppliedInfo<T> InsertIfNotExists<T>(T poco, bool insertNulls, CqlQueryOpt
             return TaskHelper.WaitToComplete(InsertIfNotExistsAsync(poco, insertNulls, queryOptions), _queryAbortTimeout);
         }
 
+        public AppliedInfo<T> InsertIfNotExists<T>(T poco, bool insertNulls, int? ttl, CqlQueryOptions queryOptions = null)
+        {
+            throw new NotImplementedException();
```
jorgebay: Not implemented...

src/Cassandra/Mapping/ICqlWriteAsyncClient.cs
```diff
@@ -33,6 +33,27 @@ public interface ICqlWriteAsyncClient
...
+        /// <para>
+        /// Use <c>false</c> if you don't want to consider <c>null</c> values for the INSERT 
+        /// operation (recommended).
+        /// </para> 
+        /// <para>
+        /// Use <c>true</c> if you want to override all the values in the table,
+        /// generating tombstones for null values.
+        /// </para>
+        /// </param>
+        /// <param name="ttl">Ttl for entity</param>
```
jorgebay: Can you add a more descriptive xmldoc on all the methods that has a ttl parameter, like:
```
/// <param name="ttl">
/// Time to live (in seconds) for the inserted values. If set, the inserted values are automatically removed
/// from the database after the specified time.
/// </param>
```

collinsauve:
Pushed new commits to address the issues raised by @jorgebay.

jorgebay:
Added a few tests (58b448e40a8e18e4445b94066bb4662c761e71ce) and landed in 43c5a8e59d041e278749fd685dc4ae2bfefbb617.

Thanks!

PR #164, opened 2015-11-09
CreateKeyspaceAsync and CreateKeyspaceIfNotExistsAsync

collinsauve:
Built on top of https://github.com/datastax/csharp-driver/pull/163 and https://github.com/datastax/csharp-driver/pull/161

This adds `CreateKeyspaceAsync` and `CreateKeyspaceIfNotExistsAsync` to `ISession` and `Session` (async versions of `CreateKeyspace` and `CreateKeyspaceIfNotExists`).

You should really consider breaking apart the `ISession` interface and having the majority of these methods as extension methods.  You would need to expose the `Configuration` and `Logger` though.

PR #193, opened 2016-02-16
Create keyspace async

collinsauve:
Adds `CreateKeyspaceAsync` function to `ISession`.

Dependent on #161, which has been open for a long time.
Similar to #164, except no longer dependent on #163, and does not add `CreateKeyspaceIfNotExistsAsync`.

### DIFF
--- a/src/Cassandra/ISession.cs
+++ b/src/Cassandra/ISession.cs
@@ -105,6 +105,19 @@ namespace Cassandra
         /// <param name="durableWrites">Whether to use the commit log for updates on this keyspace. Default is set to <c>true</c>.</param>
         void CreateKeyspaceIfNotExists(string keyspaceName, Dictionary<string, string> replication = null, bool durableWrites = true);
         /// <summary>
+        ///  Creates new keyspace in current cluster asynchronously.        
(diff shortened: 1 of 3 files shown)


collinsauve:
FYI this is not strictly required - Creating a keyspace asynchronously can be achieved outside of this API.  I added it mostly because I did not want to write my own Cql for creating the keyspace.

I have since copied [CqlQueryTools](https://github.com/datastax/csharp-driver/blob/c891b57c9f3cf52a75ccb888bf76fe0dad452afd/src/Cassandra/CqlQueryTools.cs) and [Utils](ttps://github.com/datastax/csharp-driver/blob/10fc57339de9c4a93421a3248e5709e15b602d15/src/Cassandra/Utils.cs) into our code base to use.

jorgebay:
I'm against adding another DDL method to the Session. I think it was not a good idea to add `CreateKeyspace()` method in v1.

The reason is that the core driver should be "CQL agnostic": if the query syntax change in future versions of Cassandra, the method will likely become useless.

collinsauve:
Understood.  Closing PR.

PR #201, opened 2016-03-03
changes to src/Cassandra.sln.DotSettings that my VS/R# wants to make

collinsauve:
Hi guys, my R# keeps trying to change this `src/Cassandra.sln.DotSettings` file.  Just a minor annoyance in that I always have to reset this change.

Should this file even be tracked in git? If so, should this change to the file be merged?  

I'm using R# Ultimate 10.0.2 w/ Visual Studio 2015 Update 1


jorgebay:
You can use Resharper "Manage options", to set the location you want to store the change, in this case your 'This computer' or 'Solution personal'.

PR #202, opened 2016-03-03
CSHARP-420 TypeCodec.GetColumnTypeCodeInfo does not handle IEnumerabl…

collinsauve:
…e runtime types well


collinsauve:
The code for the other generic types (Nullable, ISet, IDictionary, Tuple) look like they also have this problem.

If my solution here is okay I can take a look at those types also, if you'd like.

jorgebay:
TypeCodec got revamped in #194, its now called `Serializer` (in the serialization namespace) and supports multiple `TypeSerializer<T>`.

Any feedback on #194 would be greatly appreciated. It's expected to be released in the next version.

See https://github.com/datastax/csharp-driver/blob/CSHARP-402/src/Cassandra/Serialization/Serializer.cs#L186

PR #203, opened 2016-03-10
Linq: Use a wrapper for client projections

jorgebay:


landim:
:+1:  Also added some tests just as described in issue CSHARP-414

PR #205, opened 2016-03-22
Change Logger to use TraceSource and support for wildcard configuration

batwad:
As discussed under https://datastax-oss.atlassian.net/browse/CSHARP-171

### DIFF
--- a/src/Cassandra/Logger.cs
+++ b/src/Cassandra/Logger.cs
@@ -16,119 +16,91 @@
 
 using System;
 using System.Diagnostics;
-using System.Text;
 
 namespace Cassandra
 {
     public class Logger
     {
-        private const string DateFormat = "MM/dd/yyyy H:mm:ss.fff zzz";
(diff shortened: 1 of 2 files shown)


jorgebay:
Looks really good, I will test it to see whats the behaviour that used the old pattern (`Trace.Listeners.Add()`) as soon as possible.

jorgebay:
The TraceSource implementation is more straightforward than the one currently in the driver and its definitely an improvement. Also, the wildcard source is cool workaround to allow some extensibility without depending on an external package.

For current users attaching a listener to `Trace`, merging this patch will cause their apps to stop receiving log events.

[There is more info in the ticket](https://datastax-oss.atlassian.net/browse/CSHARP-171?focusedCommentId=27936&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27936) but I think we should provide more extensibility (via NLog) without breaking the current functionality.

PR #206, opened 2016-03-23
CSHARP-409 - Mapper: Support timestamp at CqlQueryOptions level (Inserts only for now)

collinsauve:
From https://datastax-oss.atlassian.net/browse/CSHARP-409

### DIFF
--- a/src/Cassandra/Mapping/CqlBatch.cs
+++ b/src/Cassandra/Mapping/CqlBatch.cs
@@ -77,7 +77,8 @@ namespace Cassandra.Mapping
             var values = getBindValues(poco);
             //generate INSERT query based on null values (if insertNulls set)
             object[] queryParameters;
-            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl, queryOptions?.Timestamp);
+            var timestamp = queryOptions == null ? null : queryOptions.Timestamp;
+            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl, timestamp);
 
             _statements.Add(Cql.New(cql, queryParameters, queryOptions ?? CqlQueryOptions.None));
         }
--- a/src/Cassandra/Mapping/CqlBatch.cs
+++ b/src/Cassandra/Mapping/CqlBatch.cs
@@ -77,7 +77,7 @@ namespace Cassandra.Mapping
             var values = getBindValues(poco);
             //generate INSERT query based on null values (if insertNulls set)
             object[] queryParameters;
-            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl);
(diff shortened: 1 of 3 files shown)


src/Cassandra/Mapping/CqlBatch.cs
```diff
@@ -77,7 +77,7 @@ private void Insert<T>(bool ifNotExists, bool insertNulls, T poco, CqlQueryOptio
             var values = getBindValues(poco);
             //generate INSERT query based on null values (if insertNulls set)
             object[] queryParameters;
-            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl);
+            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ifNotExists, ttl, queryOptions?.Timestamp);
```
jorgebay: Null-conditional operator is a C# 6 language feature. Can you replace it with conditional operator?

We are still using C# 4.0, [will start using C# 5 soon](https://datastax-oss.atlassian.net/browse/CSHARP-426) but using C# 6.0 means requiring newer compiler versions (dropping msbuild 12).

collinsauve:
@jorgebay fixed.  Thanks.

src/Cassandra/Mapping/Mapper.cs
```diff
@@ -211,7 +211,7 @@ public Task InsertAsync<T>(T poco, bool insertNulls, int? ttl, CqlQueryOptions q
             var values = getBindValues(poco);
             object[] queryParameters;
             //generate INSERT query based on null values (if insertNulls set)
-            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ttl: ttl);
+            var cql = _cqlGenerator.GenerateInsert<T>(insertNulls, values, out queryParameters, ttl: ttl, timestamp: queryOptions?.Timestamp);
```
jorgebay: Another one :)

collinsauve:
Oops I missed that one.  Fixed.

Also see PR #208: Using lang features > C# 4.0 should be a compile error.

jorgebay:
Looks good, I will try to add a unit test for it and merge it.

jorgebay:
After a closer look, I'm not going to merge it as it should be implemented using protocol level timestamp: you can specify it at `Statement` level without affecting the query.

You can implementing by adding it to the following method:
https://github.com/datastax/csharp-driver/blob/a4afd10b15d33bde2606d657f0f59f59fb539389/src/Cassandra/Mapping/CqlQueryOptions.cs#L131

Something like:

```csharp
if (Timestamp != null)
{
    statement.SetTimestamp(Timestamp);
}
```

That later gets applied by the statement factory here: https://github.com/datastax/csharp-driver/blob/a4afd10b15d33bde2606d657f0f59f59fb539389/src/Cassandra/Mapping/Statements/StatementFactory.cs#L44

We appreciate your continuing effort to contribute and improve the driver, keep it coming!

Sorry for the misunderstanding

collinsauve:
Implemented according to your suggestions at PR #209

PR #207, opened 2016-03-25
No 'class' in the Replication collection. It should be StrategyClass.

shinya-terasaki:


jorgebay:
Landed in 02d69c7b169bb02fc5ad25391bbaec04548e9cef, thanks!

PR #208, opened 2016-03-28
set langversion=4

collinsauve:
Only C# 4.0 language features should be used.  Any language features above this should raise a compile error.


jorgebay:
Landed in a4afd10b15d33bde2606d657f0f59f59fb539389.
Thanks!

PR #209, opened 2016-03-30
CSHARP-409 - Mapper: Support timestamp at CqlQueryOptions level

collinsauve:
From https://datastax-oss.atlassian.net/browse/CSHARP-409

Implemented according to @jorgebay's suggestions on PR #206


jorgebay:
Looks good, once we create a test for it, it can be merged.

jorgebay:
Landed in 69f2386cfe0f5bcc2518be810b63e8e802bb1342 .

PR #210, opened 2016-03-31
Allow setting read timeout at statement level

jorgebay:


landim:
:+1:  Just added the test documentation to be more clear its behavior.

landim:
:+1:  Just added the test documentation to be more clear its behavior.

PR #212, opened 2016-04-04
Use QueryAbortTimeout for metadata sync queries

jorgebay:


landim:
:+1:  Looks good to me

PR #214, opened 2016-04-04
adding appveyor support

landim:


appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+#       Set-ExecutionPolicy Unrestricted
+
+#       Write-Host "Checking if test dll was created"
+
+#       If ((Test-Path "src/Cassandra.Tests/bin/Release/Cassandra.Tests.dll") -And (Test-Path "src/Cassandra.IntegrationTests/bin/Release/Cassandra.IntegrationTests.dll")) {
+#         Write-Host "Executing integration tests"
+#         msbuild build/build.proj /v:m /t:test
+#       } Else {
+#         Write-Host "Test dll not created"
+#       }
```
jorgebay: Remove the commented lines

appveyor.yml
```diff
@@ -0,0 +1,68 @@
+version: 3.0.5.{build}
```
jorgebay: We can use a notation like `version: {branch}-{build}`, that way we wont have to change it for releases.

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+cache:
+  - C:\ProgramData\chocolatey\bin                                                                                                     
+  - C:\ProgramData\chocolatey\lib
+  - C:\Users\appveyor\.ccm\repository
+  - C:\Users\appveyor\ccm
+os: Visual Studio 2015
+platform:
+  - Any CPU
+configuration:
+  - Debug
```
jorgebay: Should be `Release`

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+
+matrix:
+  fast_finish: true   
+
+install:
+  - ps: .\appveyor_install.ps1
+
+build:
+  project: src/Cassandra.sln
+
```
jorgebay: We can add under build:
```yaml
  verbosity: minimal
```

jorgebay:
Minor nits but otherwise looks good to me.

I will attempt to make the single node instance reusable in the integration tests.

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+  NUNIT_PATH: nunit-console
+  matrix:
+    - CASSANDRA_VERSION: 3.4
+
+    - CASSANDRA_VERSION: 3.0.4
+
+    - CASSANDRA_VERSION: 2.2.5
+
+matrix:
+  fast_finish: true
```
jorgebay: For now, we should remove the fail fast, in order to see all the range of possible failures (inconsistent)

jorgebay:
There is a parsing error from appveyor, but I can not tell what the error is...

jorgebay:
This patch looks good, I've been spending some time in trying to improve the reliability of some tests on #215 .
I will manually start a couple of more builds on appveyor against #215 and merge both.

Can you squash your commits?

landim:
Commits squashed.

jorgebay:
Appveyor jobs are looking good: https://ci.appveyor.com/project/DataStax/csharp-driver/build/master-62

I think that until we get more concurrent jobs, we could narrow down the matrix to 2: C* 2.2.x and C* 3.4, we have Jenkins for builds with all the relevant C* versions, what do you think?

PR #217, opened 2016-04-08
CSHARP-409 Test for Mapper: Support timestamp at CqlQueryOptions level

landim:
Simple test to check if the timestamp set at CQlQueryOptions was set in cassandra columns writetime.


jorgebay:
Looks good to me, just rebase to tip and squash your commits.

PR #220, opened 2016-04-29
CSHARP-451 Invalid CQL if first pocoValue is null and a later value is not

collinsauve:
See https://datastax-oss.atlassian.net/browse/CSHARP-451


jorgebay:
Nice catch @collinsauve, landed in 0200563

PR #222, opened 2016-05-09
TimeUuid compare by time representation

jorgebay:
- TimeUuid.CompareTo() method uses DateTimeOffset representation to compare.
- Support big-endian systems.
- Include retrieved values in the integration tests.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/TimeUuidSerializationTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/TimeUuidSerializationTests.cs
@@ -28,6 +28,10 @@ namespace Cassandra.IntegrationTests.Core
 
             var insertQuery = String.Format("INSERT INTO {0} (id, timeuuid_sample) VALUES (?, ?)", AllTypesTableName);
             var selectQuery = String.Format("SELECT id, timeuuid_sample, dateOf(timeuuid_sample) FROM {0} WHERE id = ?", AllTypesTableName);
+            if (CassandraVersion >= new Version(2, 2))
+            {
+                selectQuery = String.Format("SELECT id, timeuuid_sample, toTimestamp(timeuuid_sample) as timeuuid_date_value FROM {0} WHERE id = ?", AllTypesTableName);
+            }
             _insertPrepared = Session.Prepare(insertQuery);
             _selectPrepared = Session.Prepare(selectQuery);
         }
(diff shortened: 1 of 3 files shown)


eplowe:
Just happened to take a look at TimeUuid class. I think it might be
beneficial to remove the use of one static Random object and required
locking and move to a static class that wraps a ThreadLocal<Random>?

jorgebay:
@eplowe That is a good idea.
We use `ThreadLocal<T>` for some reusable buffers when reading: https://github.com/datastax/csharp-driver/blob/6f30810ed7fd0ad5f242cd792465d1824070b768/src/Cassandra/Outputs/OutputRows.cs#L29

In this case, we can add a simple counter to the [value factory](https://msdn.microsoft.com/en-us/library/dd642263(v=vs.110).aspx) to create a random specifying the seed value.

I will push a patch soon.

jorgebay:
After investigating a little further, it does not seem like a good idea.
Having a pool of random generators would be less random than using a single source: http://stackoverflow.com/a/36376889/208683

It is also clearly stated on MSDN: https://msdn.microsoft.com/en-us/library/system.random(v=vs.110).aspx#ThreadSafety

> Instead of instantiating individual Random objects, we recommend that you create a single Random instance to generate all the random numbers needed by your app

If the user is worried about contention caused by the random generation of the clock id and node id, those values can be specified in the other constructor (can use thread ids and counters for resolution higher than 1/10 microseconds). Moreover, generation of timeuuids can be made from a single thread and shared, if needed.

eplowe:
@jorgebay You're absolutely correct. These fantastic SO answers illustrates the problem excellently.

http://stackoverflow.com/questions/25390301/net-framework-random-number-generator-produces-repeating-pattern/25390969#25390969

http://stackoverflow.com/questions/36376888/seeding-multiple-random-number-generators

The problem stems from relying on a weak seeding method (A global Random.Next() or Environment.TickCount + 1, etc). 

In my testing/experience I've found that using the RNGCryptoServiceProvider (https://msdn.microsoft.com/en-us/library/system.security.cryptography.rngcryptoserviceprovider(v=vs.110).aspx) to provide a seed value for the per thread instance, brings the behavior in line with using a global random. 

```csharp
public static int RNGCryptoSeed()
{
	using (var rngCryptoServiceProvider = new RNGCryptoServiceProvider())
	{
		var buffer = new byte[4];
		rngCryptoServiceProvider.GetBytes(buffer);

		return BitConverter.ToInt32(buffer, 0);
	}
}
```

If you use that method to replace the seed of the of the "master" random in the first SO answer, you'll see the behavior.

PR #224, opened 2016-05-24
Add DSL-like construction for PoolingOptions

alprema:
Added the `DefaultOptions()` static method to allow the user to choose pooling options corresponding to his Cassandra version.
This is technically a breaking change since it hides the `PoolingOptions` constructor, but users currently using the public constructor might end up with settings that are not right for their Cassandra version.

Also, I used the `DefaultOptions()` factory method in all the integration tests and am not able to run them at the moment, so this might have side effects (that can be fixed by specifying a lower Cassandra version in the failing tests).

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
@@ -165,7 +165,7 @@ namespace Cassandra.IntegrationTests.Core
             var queryOptions = new QueryOptions().SetRetryOnTimeout(false);
             var builder = Cluster.Builder().AddContactPoint(testCluster.InitialContactPoint)
                 .WithSocketOptions(socketOptions)
-                .WithPoolingOptions(new PoolingOptions()
+                .WithPoolingOptions(PoolingOptions.DefaultOptions(Version.Parse("2.0"))
                     .SetCoreConnectionsPerHost(HostDistance.Local, 1)
                     .SetHeartBeatInterval(0))
                 .WithQueryTimeout(Timeout.Infinite)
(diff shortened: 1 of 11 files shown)


alprema:
Any news on an ETA?

jorgebay:
I'll try to review this one and #225 before the end of the week.

jorgebay:
The patch looks good on the technical side, I'm thinking on alternate ways to solve the current issues with the `PoolingOptions` API.

Understanding that:

A. The current API for changing 1 setting (like heartbeat interval) is cumbersome. If we were able to define it from scratch, we could use `null` settings as default that get overridden with the default values per protocol version once the protocol version is set.
B. The current pooling defaults, based on protocol version, are suitable for most users (except for the case of `HeartBeatInterval`, which is more likely to be tuned based on infrastructure information).

I think what we should do is to provide ways to easily override heartbeat interval without having to set the C* / protocol version. For the rest of the pooling options we could provide helper methods based on protocol version: pool size is an advance setting and the user should be aware of protocol versions / max concurrency / etc.

Let me comeback on this one tomorrow / Monday with a proposal.

jorgebay:
I propose to expose the other settings (`HeartBeatInterval`) at cluster builder level and provide helpers for creating `ProtocolOptions` instances based on protocol version.

Usage samples

Setting the hearbeat interval only:
```csharp
Cluster.Builder()
  .AddContactPoint(host1)
  .WithHeartBeatInterval(ms)
```

Setting some protocol options:
```csharp
// Use the pooling options for protocol 4 and override the max connections per host.
Cluster.Builder()
  .AddContactPoint(host1)
  .WithPoolingOptions(PoolingOptions.Create(4)
    .SetMaxConnectionsPerHost(HostDistance.Local, 3))
```

That way, we allow overriding of the heartbeat interval setting without having to define the pooling options and we provide helpers to obtain the default values for a given protocol version in case the user wants to change just 1 option (like the example above).

alprema:
It does seem like a better option in terms of API. Should I modify my patch or will you implement it yourself?

jorgebay:
I will close this pull request, thanks @alprema for bringing up this issue in the API, lets see if we can implement a solution without breaking changes. Feel free to send a new patch.

PR #225, opened 2016-05-25
Observe exceptions in AttemptReconnection

alprema:
When calling `AttemptReconnection()`, exceptions were not observed on `_creationTcs`'s Task and were handled by the finalizer (which can be caught in a client application).

This fix makes sure that the exceptions are always observed, but still sets `_creationTcs`'s exception in case `MaybeCreateFirstConnection()` is called at the same time.

### DIFF
--- a/src/Cassandra/HostConnectionPool.cs
+++ b/src/Cassandra/HostConnectionPool.cs
@@ -212,6 +212,13 @@ namespace Cassandra
                 {
                     t.Exception.Handle(e => true);
                     ex = t.Exception.InnerException;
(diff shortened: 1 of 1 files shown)


alprema:
Any news on an ETA?

src/Cassandra/HostConnectionPool.cs
```diff
@@ -212,6 +212,13 @@ internal void AttemptReconnection()
                 {
                     t.Exception.Handle(e => true);
                     ex = t.Exception.InnerException;
+                    //This makes sure that the exception is observed, but still sets _creationTcs' exception
```
jorgebay: Makes sense

PR #226, opened 2016-06-10
Consider racks on NetworkTopologyStrategy

jorgebay:
Skips hosts on the same rack until all racks have been included in the DC.
Memoize token building for keyspaces with the same network replication

### DIFF
--- a/src/Cassandra.Tests/TokenTests.cs
+++ b/src/Cassandra.Tests/TokenTests.cs
@@ -21,6 +21,7 @@ using System.Net;
 using System.Text;
 ﻿using Moq;
 ﻿using NUnit.Framework;
+using DatacenterInfo = Cassandra.TokenMap.DatacenterInfo;
 
 namespace Cassandra.Tests
 {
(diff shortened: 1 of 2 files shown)


landim:
:+1: LGTM

src/Cassandra.Tests/TokenTests.cs
```diff
@@ -239,15 +239,125 @@ public void TokenMap_Build_NetworkTopology_Adjacent_Ranges_Test()
...
+            };
+            foreach (var v in values)
+            {
+                var replicas = map.GetReplicas("ks1", new M3PToken(v.Item1));
+                CollectionAssert.AreEqual(v.Item2, replicas.Select(TestHelper.GetLastAddressByte));
+            }
+        }
+
+        [Test, Timeout(2000)]
+        public void TokenMap_Build_NetworkTopology_Quickly_Leave_When_Dc_Not_Found()
```
landim: :+1: good call on this method

PR #227, opened 2016-06-10
CSHARP-466 - DateTime.Kind should be set to DateTimeKind.Utc when des…

collinsauve:
Fixes https://datastax-oss.atlassian.net/browse/CSHARP-466


jorgebay:
This pull request only deals with Mapper component on the driver, introducing different behaviours in the core and in the Mapper/Linq.

I would advice you to use `DateTimeOffset` to represent CQL timestamps, as it is a more accurate representation than `DateTime`, but you can provide your own `TypeConverter` implementation on the mapping configuration if you still want to deal with UTC-based `DateTime` instances.

More info on the ticket https://datastax-oss.atlassian.net/browse/CSHARP-466

collinsauve:
Thanks @jorgebay, I will try to write my own `TypeConverter`.

PR #234, opened 2016-06-17
Csharp 461 tests

landim:


src/Cassandra.IntegrationTests/Policies/Tests/RetryPolicyTests.cs
```diff
@@ -226,6 +226,40 @@ private void RetryPolicyTest(ITestCluster testCluster)
...
+        /// 
+        /// @jira_ticket CSHARP-461
+        /// 
+        /// @test_category connection:retry_policy
+        /// </summary>
+        [Test, Category("short")]
+        public void IdempotenceAwareRetryPolicy_ShouldUseChildRetryPolicy_OnWriteTimeout()
+        {
+            const string keyspace = "idempotenceAwarepolicytestks";
+            var testCluster = TestClusterManager.GetNonShareableTestCluster(2);
```
jorgebay: `TestClusterManager.GetNonShareableTestCluster()` is deprecated see [xml doc](https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs#L96).

Use `CreateNew()` instead.

src/Cassandra.IntegrationTests/Policies/Tests/RetryPolicyTests.cs
```diff
@@ -226,6 +226,40 @@ private void RetryPolicyTest(ITestCluster testCluster)
...
+            testCluster.Session.Execute(string.Format(TestUtils.CreateKeyspaceSimpleFormat, keyspace, 2), ConsistencyLevel.All);
+            testCluster.Session.ChangeKeyspace(keyspace);
+            testCluster.Session.Execute(new SimpleStatement(string.Format("CREATE TABLE {0} (k int PRIMARY KEY, i int)", tableName)).SetConsistencyLevel(ConsistencyLevel.All));
+
+            testCluster.PauseNode(2);
+
+            //normally, it would throw Unavailable exception because of the ConsistencyLeve.ALL set, 
+            //but it will retry with ConsistencyLeve.ONE (IdempotenceAwareRetryPolicy) and should pass
+            //altought this query is not idempotent (just for test purpose)
+            Assert.DoesNotThrow(() => testCluster.Session.Execute(new SimpleStatement(string.Format("INSERT INTO {0}(k, i) VALUES (0, 0)", tableName))
```
jorgebay: It's OK to have this assertion, but it still evaluates `true` without `IdempotenceAwareRetryPolicy`.

Add another assertion to test the other possibility: non-idempotent statement:
- It should throw (write timeout or unavailable).
- It should NOT use the child policy: create a child policy similar to [this one](https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.Tests/PoliciesUnitTests.cs#L689) that counts the calls, use it as child policy. Assert that calls to write or unavailable are zero.

PR #237, opened 2016-06-28
CSHARP-476 - KeyNotFoundException from Cassandra Error Response

LukeTillman:
Not sure what the root cause of the unknown error code from Cassandra is, but this should provide a more helpful Exception if anyone runs into this in the future. Unit test included.

### DIFF
--- a/src/Cassandra.Tests/Cassandra.Tests.csproj
+++ b/src/Cassandra.Tests/Cassandra.Tests.csproj
@@ -119,6 +119,7 @@
     <Compile Include="Mapping\TestData\TestDataHelper.cs" />
     <Compile Include="Mapping\UpdateTests.cs" />
     <Compile Include="ObsoletedMemberTests.cs" />
+    <Compile Include="OutputErrorTests.cs" />
     <Compile Include="PoliciesUnitTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="RequestHandlerTests.cs" />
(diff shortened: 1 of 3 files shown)


jorgebay:
Nice catch!
+1 for the test.

TioLuiso:
I have been able to reproduce this (I believe it is the same case)

We have a type we are persisting into a cassandra table. That type has a property that is an enum, which is mapped as text in cassandra. If, by any chance, an invalid value ends up in that column (invalid for the specific enum type semantics anyway), when cassandra tries to map it to the enum, using EnumStringMapper.MapStringToEnum, it will fail miserably, because the StringToEnumCache dictionary has only valid enum values. So a KeyNotFoundException will be thrown.

What I would like very much is to have a more specific exception that informs me of this.

jorgebay:
@TioLuiso I think is not directly related to this issue, but let's look at it anyway:
You mean you are using the Mapper/Linq to map a string to an enum and because there is an invalid value stored in that column (not present in the enum), the Mapper throws `KeyNotFoundException` which is not specific, right?

If that is the case, I think the driver could throw a `MappingException` or something like that providing the info, wdyt?

TioLuiso:
Right you are.
Well, kindof. It is even funnier. we parse a file, and load an enum value from there (with Enum.TryParse). It is not defined in the enum. But the parse will not fail because of the nature of the enum in C#. It will have the invalid value. For example, if the enum has values 0, 1 and 2, it will accept value 3.
Then we store it in cassandra. And it won't fail. And you can see in the column the value "3"
Then we try to retrieve it, and this time it explodes. Because "3" isn't in the dictionary.

Yup, I believe MappingException would be really nice. If posible indicating the enum type and the value.

Thanks a huge lot

jorgebay:
Sounds good!

I've created a ticket for it: [CSHARP-517](https://datastax-oss.atlassian.net/browse/CSHARP-517).

PR #243, opened 2016-08-12
[fix] Use KeyspaceName for CqlGenerator

sumtiogo:
whithout this patch, the following code will failed due to `No keyspace
has been specified.`

```csharp
using Cassandra;
using Cassandra.Mapping;
namespace Test
{
    class Program
    {
        static void Main(string[] args)
        {
            Cluster cluster = Cluster.Builder()
                .AddContactPoint("localhost")
                .Build();
            Mapper mapper = new Mapper(cluster.Connect());
            MappingConfiguration.Global.Define<MyMapping>();
            User u = mapper.First<User>("WHERE id = ?", 1);
        }
        public class MyMapping : Mappings
        {
            public MyMapping()
            {
                For<User>()
                    .KeyspaceName("test")
                    .TableName("user")
                    .Column(u => u.id, cm => cm.WithName("id"));
            }
        }
        public class User
        {
            public int id { get; set; }
        }
    }
}
```

### DIFF
--- a/src/Cassandra/Mapping/Statements/CqlGenerator.cs
+++ b/src/Cassandra/Mapping/Statements/CqlGenerator.cs
@@ -48,7 +48,18 @@ namespace Cassandra.Mapping.Statements
             // If it's got the from clause, leave FROM intact, otherwise add it
             cql.SetStatement(FromRegex.IsMatch(cql.Statement)
                                  ? string.Format("SELECT {0} {1}", allColumns, cql.Statement)
-                                 : string.Format("SELECT {0} FROM {1} {2}", allColumns, Escape(pocoData.TableName, pocoData), cql.Statement));
+                                 : string.Format("SELECT {0} FROM {1} {2}", allColumns, GetEscapedTableName(pocoData), cql.Statement));
+        }
+
+        private static string GetEscapedTableName(PocoData pocoData)
+        {
+            string name = null;
+            if (!string.IsNullOrEmpty(pocoData.KeyspaceName))
+            {
+                name = Escape(pocoData.KeyspaceName, pocoData) + ".";
+            }
+            name += Escape(pocoData.TableName, pocoData);
+            return name;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


jorgebay:
Good catch! It looks like the issue affects all CQL generation using the Mapper.
Would you like to fix them all?
- Line 145
- Line 195
- Line 204
- Line 226

We could create a static method to get the table name and call it multiple times, similar to this:
```csharp
private static string GetEscapedTableName(PocoData pocoData)
{
    string name = null;
    if (Table.KeyspaceName != null)
    {
        name = Escape(Table.KeyspaceName, pocoData) + ".";
    }
    name += Escape(Table.Name, pocoData);
    return name;
}
```

Also, do you mind changing the commit message (avoid including a code snippet) and rebase?

src/Cassandra/Mapping/Statements/CqlGenerator.cs
```diff
@@ -45,10 +45,11 @@ public void AddSelect<T>(Cql cql)
             var pocoData = _pocoDataFactory.GetPocoData<T>();
             var allColumns = pocoData.Columns.Select(Escape(pocoData)).ToCommaDelimitedString();
 
+            var identifier = string.IsNullOrEmpty(pocoData.KeyspaceName) ? pocoData.TableName : string.Format("{0}.{1}", pocoData.KeyspaceName, pocoData.TableName);
```
jorgebay: The fully-qualified escaped table name should be of the form: `"Ks1"."MyTable"` (not `"Ks1.MyTable"`)

landim:
Looks good, whould you mind rebase to squash the commits into one*? 
* http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

sumtiogo:
Done. Please check if that is what you want. Thanks.

jorgebay:
Merged, thanks @sumtiogo !

sumtiogo:
You're welcome. Thanks for developing this package, it helps me a lot.

PR #244, opened 2016-08-12
Update build.yaml to use netcore and mono

landim:
Needed to change the build script. Also create an additional step to parse the generated
result xml to a xunit format, due unit3 format issues.
the solution of passing the format to nunit3 console (--result:file.xml;format=nunit2)
dont work when starting nunit3 console through "dotnet test" command. Thats why it is
using the xslt to parse the xml.

TODO: some tests are failing. Need investigation.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -6,31 +6,51 @@ cassandra:
   - 2.2
   - 3.0
   - 3.7
+env:
+  CSHARP_VERSION:
+    - mono
+    - netcore
 build:
   - script: |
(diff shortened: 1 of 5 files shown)


jorgebay:
Looks good to me, its running smoothly on Jenkins (linux) for netcore and mono. Removed the WIP in the name.

jorgebay:
Merged in 3.1 branch! Landed in 78b89ed87bbee662ada9c7ba6653fd13889a1edf.

PR #250, opened 2016-08-31
Updated DataTypeParser.ParseTypeName to handle case sensitive UDTs.

marcelvillet:
I had a table using a UDT with a case sensitive name:

``` CQL
CREATE TYPE "MyType"
(
...
);

CREATE TABLE "MyTable"
(
    "MyColumn"   FROZEN <"MyType">
...
);
```

When getting table schema information using `ICluster.Metadata.GetTable`, I received a `"System.ArgumentException"` exception: `Not a valid type "MyType"`. This update basically strips out the double quotes from the type name before calling the method that searches for the type.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,12 +2,12 @@ version: "{branch}-{build}"
 skip_tags: true
 shallow_clone: true
 cache:
-  - C:\ProgramData\chocolatey\bin                                                                                                     
-  - C:\ProgramData\chocolatey\lib
-  - C:\Users\appveyor\.ccm\repository
-  - C:\Users\appveyor\deps -> appveyor.ps1
-  - C:\Users\appveyor\ccm
-os: Visual Studio 2015
+  - C:\ProgramData\chocolatey\bin -> appveyor_install.ps1                                                                             
+  - C:\ProgramData\chocolatey\lib -> appveyor_install.ps1
+  - C:\Users\appveyor\.ccm\repository -> appveyor_install.ps1
+  - C:\Users\appveyor\deps -> appveyor_install.ps1
+  - C:\Users\appveyor\ccm -> appveyor_install.ps1
+os: Previous Visual Studio 2015
 platform:
   - Any CPU
 configuration:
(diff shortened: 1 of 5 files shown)


jorgebay:
I'm not sure if I understand the issue that this patch is fixing.

If you are using case sensitive UDTs (which I don't see a reason to use), you can use `SetIgnoreCase()` method when creating the `UdtMap`:
```csharp
UserDefinedTypes.Define(UdtMap.For<MyTypeClassName>("MyType").SetIgnoreCase(false))
````

If `SetIgnoreCase(false)` does not solve the issue you are facing, could you describe the steps to reproduce it?

marcelvillet:
My first post as shown above does describe the steps I followed. I created a case sensitive UDT using Datastax DevCenter and applied it to a table. The driver then crashed as indicated. I specifically created my UDT as case-sensitive because I chose to do so.

Whether or not it is good practice to create case sensitive UDTs is neither here nor there. The fact is that Cassandra allows one to do it and the driver cannot handle it. Hence the proposed fix.

jorgebay:
thanks @marcelvillet for the extra detail.

How are you making the mapping on your C# code? have you tried using `SetIgnoreCase(false)`?

marcelvillet:
I query the JSON out and use a JSON mapper to convert to my C# POCO class. The C# driver manual was not clear on whether prepared statements are in fact used behind the scenes with Mapper/LINQ. So I opted not to use the latter. Also, if you have a single POCO class that is used for multiple tables - it does not seem like the driver caters for that when using Mapper.

The call to `ICluster.Metadata.GetTable` was used to get primary key information from the table to facilitate the deletes. But as I mentioned the call failed as the table had a case-sensitive UDT defined on it.

marcelvillet:
Oh, sorry Jorge. Misunderstood. I am not defining it. The JSON conversion sorts it out.

marcelvillet:
So I basically do a straight SELECT JSON * and map the results to a POCO class using a JSON converter.

joao-r-reis:
Hi @marcelvillet,  I've noticed this PR hasn't had any activity in a while... Are you still interested in going forward with this PR? If so, I will try to reproduce this in order to better understand what's happening. Also, please change the base branch to `master` and update your branch to force a build on travis.

marcelvillet:
As far as I know these changes have been pulled in already.

jorgebay:
ok, I'll close it, thanks @marcelvillet !

PR #251, opened 2016-09-02
Fix: Reading while canceling in-flight requests can result in NRE

jorgebay:
Fix CSHARP-493: Avoid NRE on `Connection.ReadParse()`
Fix CSHARP-500: Concurrent calls to `OperationState.SetCompleted()` can result in deadlock

### DIFF
--- a/src/Cassandra/OperationState.cs
+++ b/src/Cassandra/OperationState.cs
@@ -38,7 +38,7 @@ namespace Cassandra
         private const int StateTimedout = 2;
         private const int StateCompleted = 3;
         private Action<Exception, Response> _callback;
-        private static readonly Action<Exception, Response> Noop = (_, __) => { };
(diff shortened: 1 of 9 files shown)


src/Cassandra/OperationState.cs
```diff
@@ -146,5 +153,24 @@ public void Cancel()
...
+        internal static void CallbackMultiple(IEnumerable<OperationState> ops, Exception ex)
+        {
+            Task.Factory.StartNew(() =>
+            {
+                foreach (var state in ops)
+                {
+                    var callback = state.SetCompleted();
+                    if (callback == Noop)
+                    {
+                        return;
```
jorgebay: It should be `continue`

landim:
:+1: lgtm

PR #254, opened 2016-09-07
Testing Ci

marcelvillet:
Canary pull request

### DIFF
--- a/src/Cassandra.Tests/IOUnitTests.cs
+++ b/src/Cassandra.Tests/IOUnitTests.cs
@@ -54,7 +54,7 @@ namespace Cassandra.Tests
                 {
                     () =>
                     {
-                        var timedout = state.SetTimedOut(
+                        var timedout = state.MarkAsTimedOut(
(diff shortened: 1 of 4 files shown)


jorgebay:
Hi!
If you are looking for submitting a pull request, you can create a new branch in your fork out of master (ie: `git checkout -b my-fix`), push it to your fork and then submitting a pull request.
Make sure to avoid having merge commits in your branch (ie: rebase against master).

marcelvillet:
Thanks for the advice. Point taken on the merge commits.

We made a new branch and resubmitted the merge request like you instructed but it still fails CI.

For some reason adding c559786899a5d31e93f8bd4e2fa041aa80dc00fe to our commit seems to fix the problem. Is master broken? See #254

jorgebay:
The failure in the [CI run](https://ci.appveyor.com/project/DataStax/csharp-driver/build/master-429) is caused by a race condition in the test, we should improve the stability of the tests to prevent that: its unrelated to your changes, so don't mind.
We could trigger a new build without sending a new change, just use master as base branch.

PR #255, opened 2016-09-07
Updated DataTypeParser.ParseTypeName to handle case sensitive UDTs.

marcelvillet:
I had a table using a UDT with a case sensitive name:

```CQL
CREATE TYPE "MyType"
(
...
);

CREATE TABLE "MyTable"
(
    "MyColumn"   FROZEN <"MyType">
...
);
```

When getting table schema information using `ICluster.Metadata.GetTable`, I received a `"System.ArgumentException"` exception: `Not a valid type "MyType"`. This update basically strips out the double quotes from the type name before calling the method that searches for the type.

### DIFF
--- a/src/Cassandra/Serialization/DataTypeParser.cs
+++ b/src/Cassandra/Serialization/DataTypeParser.cs
@@ -374,6 +374,7 @@ namespace Cassandra.Serialization
                 dataType.TypeCode = typeCode;
                 return TaskHelper.ToTask(dataType);
             }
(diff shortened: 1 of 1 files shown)


jorgebay:
I've created the ticket on JIRA for tracking: https://datastax-oss.atlassian.net/browse/CSHARP-504

Cherry picked your commit into #259, included an specific integration test and an additional fix.

Nice catch @marcelvillet!  you can follow the progress of your patch on #259 .

PR #257, opened 2016-09-09
CSHARP-456: Fixed default TimeUuid handling in Serializer.cs

LucVL:


LucVL:
Commit mentions the wrong JIRA issue

PR #258, opened 2016-09-09
CSHARP-503: Fixed default TimeUuid handling in Serializer.cs

LucVL:


jorgebay:
Thanks @LucVL for the detail of the issue on JIRA.

The default C# type for a CQL `timeuuid` is `Guid`, the reason is that we introduced `TimeUuid` in a minor version (2.1) and we didn't want to introduce a breaking change. The [documentation states that the default type is `TimeUuid`](https://github.com/datastax/csharp-driver/tree/master/doc/features/datatypes) when it should say that the default is `Guid`.

Similar to what happens with `list` and `set` types, where the default type is a `IEnumerable<T>` but you can retrieve it as a `List<T>` or `SortedSet<T>`, even if the default is `Guid` we want to allow `TimeUuid` to be used when writing and reading using the driver and it does, but, as you detail in the ticket there are 2 issues, when retrieving the values using `Row.GetValue<T>()`:

1. `Nullable<TimeUuid>` is not currently supported
2. `timeuuid` collections are retrieved as `IEnumerable<Guid>` when it should be, `IEnumerable<TimeUuid>` (when asked specified as target type). 

That said, the solution you propose is simple but its a breaking change.
I think the best solution would be to handle it at [row level, similar to the `Guid` to `TimeUuid` conversion](https://github.com/datastax/csharp-driver/blob/81fd36c006d07322ea5039409113a4122fb1185c/src/Cassandra/RowPopulators/Row.cs#L254). For example: 

```csharp
var isTargetTimeUuid = (targetType == typeof (TimeUuid) || targetType == typeof (TimeUuid?));
if (isTargetTimeUuid && !(value is TimeUuid))
{
  // ...
}
```
The reason we don't want to introduce a breaking change for expected C# types is that the exception at user level won't be very obvious to understand for users that are retrieving the values like this:

```csharp
object value = row.GetValue<object>("my_col_value");
```

jorgebay:
Would you like to modify your patch to do it at row level?

jorgebay:
Fixed as part of #280. Thanks identifying and proposing a solution @LucVL !

PR #260, opened 2016-09-14
Support type conversion for UDT fields

ElliotSis:
I tried to use UDTs and I found out that there was no implicit conversion for List, Enums and DateTime like there is in the mapper.
There are two tickets opened regarding that matter:
https://datastax-oss.atlassian.net/browse/CSHARP-381
https://datastax-oss.atlassian.net/browse/CSHARP-482
I tried to add that feature using the existing DefaultTypeConverter during the serialization/deserialization of UDTs.

I think I spotted a bug in the CreateDelegate() extension method while doing that (the generated Func type didn't include the return type of the method).

I also added a CQL generator to create UDTs in Cassandra with the driver.

### DIFF
--- a/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
+++ b/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
@@ -61,37 +61,29 @@ namespace Cassandra.Mapping.TypeConversion
             return converter(value);
         }
 
-        internal object ConvertObjectValue(Type valueType, Type dbType, object value)
+        /// <summary>
+        /// Converts a UDT field value (POCO) to to a destination type value for storage in C*.
+        /// </summary>
+        internal object ConvertToDbFromUdtFieldValue(Type valueType, Type dbType, object value)
         {
-            var converter = GetToDbConverter(valueType, dbType);
+            var converter = GetToDbConverter(dbType, valueType);
             if (converter == null)
             {
-                try
-                {
-                    return Convert.ChangeType(value, dbType);
-                }
-                catch (Exception)
-                {
-                    return value;
-                }
+                throw new InvalidTypeException(string.Format("Type {0} is not convertible to type {1}", valueType, dbType));
             }
 
             return converter.DynamicInvoke(value);
         }
 
-        internal object ConvertDbValue(Type dbType, Type valueType, object value)
+        /// <summary>
+        /// Converts a source type value from the database to a destination type value on a POCO.
+        /// </summary>
+        internal object ConvertToUdtFieldFromDbValue(Type dbType, Type valueType, object value)
         {
             var converter = GetFromDbConverter(dbType, valueType);
             if (converter == null)
             {
-                try
-                {
-                    return Convert.ChangeType(value, valueType);
-                }
-                catch (Exception)
-                {
-                    return value;
-                }
+                throw new InvalidTypeException(string.Format("Type {0} is not convertible to type {1}", dbType, valueType));
             }
 
             return converter.DynamicInvoke(value);
--- a/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
+++ b/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
@@ -66,10 +66,10 @@ namespace Cassandra.Mapping.TypeConversion
         /// </summary>
         internal object ConvertToDbFromUdtFieldValue(Type valueType, Type dbType, object value)
         {
-            var converter = GetToDbConverter(dbType, valueType);
+            var converter = GetToDbConverter(valueType, dbType);
             if (converter == null)
             {
-                throw new InvalidTypeException(string.Format("Type {0} is not convertible to type {1}", valueType, dbType));
+                throw new InvalidTypeException(string.Format("No converter is available from Type {0} is not convertible to type {1}", valueType, dbType));
             }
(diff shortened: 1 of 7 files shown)


jorgebay:
We were not able to review your pull request yet, we will try to catch up as soon as possible.

src/Cassandra/UdtMap.cs
```diff
@@ -83,7 +84,9 @@ public abstract class UdtMap
         protected readonly Dictionary<PropertyInfo, string> _propertyToFieldName;
         // ReSharper enable InconsistentNaming
         private Serializer _serializer;
-        protected const BindingFlags PropertyFlags = BindingFlags.FlattenHierarchy | BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance;
+        public const BindingFlags PropertyFlags = BindingFlags.FlattenHierarchy | BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance;
+
+        public TypeConverter TypeConverter { get; private set; }
```
jorgebay: Initially, it would be best not to expose the type converter in `UdtMap` class.
`TypeConverter` is an abstract class exposed on a higher namespace: `Cassandra.Mapping.TypeConversion` that was exposed for the Mapper and Linq components of the driver.

I think that we can internally reuse it, use a single instance for all `UdtMap` and if users in the future start asking for custom conversions outside of the default type converter (enums, collections, dates, ... are already covered), we could expose an interface that would be valid for both (udts and Mapper/Linq component).

jorgebay:
^ Do you mind leaving CQL generation out of this pull requests?

src/Cassandra/UdtMap.cs
ElliotSis: I'm thinking about creating a new class UdtFieldConverter that uses a TypeConverter passed in the constructor and that would expose two methods : ConvertToDbUdtFieldValue() and ConvertToUdtFieldDbValue(). 
An instance of that class that uses a DefaultTypeConverter would be exposed in the MappingConfiguration (just like the MapperFactory).
When ConvertTypesUsing(TypeConverter typeConverter) is called in the MappingConfiguration, it would change the TypeConverter used by the UdtFieldConverter as well.

Is that design better?

src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
```diff
@@ -61,6 +61,42 @@ protected TypeConverter()
...
         }
 
+        internal object ConvertObjectValue(Type valueType, Type dbType, object value)
+        {
+            var converter = GetToDbConverter(valueType, dbType);
+            if (converter == null)
+            {
+                try
+                {
+                    return Convert.ChangeType(value, dbType);
```
ElliotSis: We could throw an exception in the `TypeConverter`. 
When converting the db / udt field value, we would check if the conversion is necessary using `IsAssignableFrom()` and only perform it in that case.
If something goes wrong, the exception will be thrown and the user will know the cause of the error.

Is that what you were thinking about?

ElliotSis:
Sure, I'll remove the CQL generation in the next commit, and I'll make another PR with some unit tests for that feature during the week.

src/Cassandra/UdtMap.cs
jorgebay: I don't think we should expose another conversion API, at least for now, before understanding if there is a need for it.

If there is a need for it in the future, it would be nice to expose a single one for both uses (either a new interface or the existent `TypeConverter`).

ElliotSis: I see, I'll just use the `TypeConverter `exposed in the `MapperFactory `of the `MappingConfiguration `then.

jorgebay: You could use a single static instance of the default type converter as a internal field:

```csharp
    public class UdtMap
    {
         internal static TypeConverter TypeConverter = new DefaultTypeConverter();

         // Use it in the ToObject() method
         internal object ToObject(object[] values)
         {
              // ...
              TypeConverter.ConvertDbValue(/*...*/);
              // ...
          }
    }

    // Reuse the same instance in the UdtSerializer class
    public class UdtSerializer
    {
         public override byte[] Serialize(/*...*/)
         {
              // ...
              fieldValue = UdtMap.TypeConverter.ConvertObjectValue(/**/);
              // ...
          }
    } 
```

Avoid referring to `MappingConfiguration` as it's from the Mapper component.

ElliotSis: I'm on it, thanks!

src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
ElliotSis: Hum, it seems that removing that cast created a regression : List and Set still need to be cast to IEnumerable when uploaded to the db.

jorgebay:
LGTM!
We'll include some tests to cover the new functionality and merge it to master.

ElliotSis:
Great, thanks! Let me know if you run into some issues or if you need me to write the tests.

jorgebay:
Thanks for your contribution!
Landed in e83e578, you can follow #263 for the progress of the tests.

jorgebay:
Finally landed in master: b6e429469acf3c827681041e45b10210e59c9ee6 🎉

PR #262, opened 2016-09-24
I know I shouldnt do this, but i want this question answered really badly

akash3456:
For tuples in the Cassandra/DataStax - How do i process that in the C# API, as a UDT im assuming right?


LukeTillman:
The default .NET Type when dealing with a CQL Tuple is [`System.Tuple`](https://msdn.microsoft.com/en-us/library/system.tuple(v=vs.110).aspx).

akash3456:
An example of using the tuple in c# code would be very helpful as im not sure how to use it as a tuple would not fall into the IEnumerable object?

jorgebay:
You can ask questions on the mailing list: https://groups.google.com/a/lists.datastax.com/forum/#!forum/csharp-driver-user

Anyway, here you have an example how to retrieve a tuple value:
```
var row = session.Execute(query).First();
var tupleValue = row.GetValue<Tuple<string, int>>("my_col");
```

PR #267, opened 2016-11-07
Don't mark host down while one connection is active

jorgebay:
Refactor pool to allow partially connected pools and prevent nodes to be marked as down until all connections (from all connection pools) have been closed.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ConnectionTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ConnectionTests.cs
@@ -630,10 +630,10 @@ namespace Cassandra.IntegrationTests.Core
         }
 
         [Test]
-        public void Connection_Close_Faults_AllPending_Tasks()
+        public async Task Connection_Close_Faults_AllPending_Tasks()
         {
             var connection = CreateConnection();
-            connection.Open().Wait();
+            await connection.Open();
             //Queue a lot of read and writes
             var taskList = new List<Task<Response>>();
             for (var i = 0; i < 1024; i++)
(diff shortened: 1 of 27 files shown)


jorgebay:
@landim scassandra integration for tests is looking good. As the implementation of the new host pool involves a lot of changes, do you mind integrating scassandra in a different branch?

jorgebay:
Squashed commits.

jorgebay:
There is a failure in the [last appveyor build](https://ci.appveyor.com/project/DataStax/csharp-driver/build/CSHARP-274-651/job/4hjahtcvyw2c4uku) that surfaced an issue in the implementation of the logger (CSHARP-171).

I'm going to fix it on this branch.

jorgebay:
After several builds, there is another failure: https://ci.appveyor.com/project/DataStax/csharp-driver/build/CSHARP-274-708/job/qyeyrdfdc6dk8xrh

It looks like a driver issue, I'm looking into it.

jorgebay:
The issue [described here](https://github.com/datastax/csharp-driver/pull/267#issuecomment-266390177) is now fixed.

I've set the reconnection test to be repeated 20 times per build on f3a4069 and it passes (executed total of 400 times on Jenkins and 120 on appveyor).

I've restore to only execute it few times per build.

I've also added several test fixes. I will merge it once the last build finishes.

ablanchet:
Hi @jorgebay 

It seems that you merged the implementations of two properties on Host : `IsConsiderablyUp` and `IsUp` now return the same value (which is `IsUp`). 

I think it could be nice to mark the `IsConsiderablyUp` property as obsolete - and remove it from good for the next breaking changes release. 

I checked in the java implem and I can't find the same _considerably_ thing, only a  [`getIsUp` method](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/Host.java#L312)

jorgebay:
You are right about deprecating `IsConsiderablyUp`, it was previously used to determine if reconnection attempts can be made in case it was down.
I would soft-deprecate it, I'm very conservative about using `[Obsolete]` in a minor version: for users with "Warning as error" setting it will mean a breaking change.

We can now include a paragraph in the xmldoc, saying something like: `"This property is going to be removed in future versions, use IsUp instead"` and in the next major version, we can mark it as obsolete.

Would you like to send a pull request for it?

ablanchet:
Yes I will push a PR later today. 
And you're right about the obsolete attribute, I'll just add a comment and leave the breaking change for a future version.

PR #269, opened 2016-11-21
Alternate ILoggerProvider-based logging API	

jorgebay:


landim:
👍  Project used to test NLog integration > https://github.com/landim/csharp-driver-nlog-test

jorgebay:
Merged! +1 on the test project

PaulDMendoza:
We need to get rid of the package dependency this introduced because that package only supports .NET 4.5. It makes it impossible for .NET 4.5.1 and above projects to use the Nuget package for CassandraCSharpDriver. We've had to go back to using version 3.1.0.1 which is the last version where this worked.

jorgebay:
@PaulDMendoza [Microsoft.Extensions.Logging v1.0.0][1] and [Microsoft.Extensions.Logging.Abstractions v1.0.0][2] support .NET Standard 1.1+ [which includes .NET Framework 4.5][3], so it shouldn't be a problem.

We tested on a fresh Windows Server 2012 r2 with .NET Framework 4.5 SDK only and it works without any issues. Maybe you are using an older version of the nuget client?

[1]: https://www.nuget.org/packages/Microsoft.Extensions.Logging/1.0.0
[2]: https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/1.0.0
[3]: https://github.com/dotnet/standard/blob/master/docs/versions.md

PaulDMendoza:
.NET 4.5 is different and older than .NET 4.5.1. I was able to get it to work when I changed my assembly to use .NET 4.5 but I've got a bunch of applications that are already targeting .NET 4.5.1. 

I am using Visual Studio 2013 on Windows 8. But that shouldn't really make a difference.

PR #271, opened 2016-11-28
Mapper: Use TypeConverter for collection items

jorgebay:


landim:
👍 Included another test to cover HashSet type.

jorgebay:
Added `TimeUuid` collection conversion support at row level

src/Cassandra.IntegrationTests/Mapping/Tests/Fetch.cs
```diff
@@ -462,6 +462,27 @@ public void Fetch_TimeUuid_HashSet_Key()
         }
 
         [Test]
+        public void Fetch_TimeUuid_List()
+        {
+            _session.Execute("CREATE TABLE tbl_timeuuid_hashset (id int primary key, my_set set<timeuuid>)");
```
jorgebay: Don't reuse the same `tbl_timeuuid_hashset` table.

src/Cassandra.IntegrationTests/Mapping/Tests/Fetch.cs
```diff
@@ -462,6 +462,27 @@ public void Fetch_TimeUuid_HashSet_Key()
...
+            var map = new Map<PocoWithCollections<TimeUuid>>()
+                .Column(p => p.Id)
+                .Column(p => p.List, c => c.WithName("my_set"))
```
jorgebay: Mapping a set to a `List` would be allowed but it would cause sorting issues, like the failures in the build: https://ci.appveyor.com/project/DataStax/csharp-driver/build/CSHARP-511-671

landim:
Testing rebase at PR https://github.com/datastax/csharp-driver/pull/274. it passes I will push --force at this branch

PR #273, opened 2016-12-07
Csharp 274 scassandra

landim:


src/Cassandra.IntegrationTests/Cassandra.IntegrationTests.csproj
```diff
@@ -167,6 +167,7 @@
     <Compile Include="TestBase\Options.cs" />
     <Compile Include="CommonFixtureSetup.cs" />
     <Compile Include="TestClusterManagement\ITestCluster.cs" />
+    <Compile Include="TestClusterManagement\SCassandraManager.cs" />
```
jorgebay: I think you must also add the reference to `System.Net` (ie: line 52)

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -115,5 +117,59 @@ private Task<string[]> ExecuteMultiple(ITestCluster testCluster, Session session
...
+                SCassandraManager.DropConnection(ports[ports.Length - 1]).Wait();
+                Assert.AreEqual(2, pool.OpenConnections);
+                Assert.IsTrue(h1.IsUp);
+                //remove one connection
+                ports = SCassandraManager.GetListOfConnectedPorts().Result;
+                SCassandraManager.DropConnection(ports[ports.Length - 1]).Wait();
+                Assert.AreEqual(1, pool.OpenConnections);
+                Assert.IsTrue(h1.IsUp);
+                //remove one connection
+                ports = SCassandraManager.GetListOfConnectedPorts().Result;
```
jorgebay: We can also check here if the `session` is can still handle queries.

jorgebay:
You should not need to modify the `nuget.exe` file, can remove that change from the pr?

BTW: I've created a ticket to migrate the project to automatic restore without the `.nuget` folder.

src/.nuget/NuGet.targets
```diff
@@ -128,8 +128,9 @@
                     OutputFilename = Path.GetFullPath(OutputFilename);
 
                     Log.LogMessage("Downloading latest version of NuGet.exe...");
+                    Console.WriteLine("Downloading latest version of NuGet.exe...");
```
jorgebay: Changed to .nuget should not be needed...

.travis.yml
```diff
@@ -21,6 +21,7 @@ branches:
 
 solution: src/Cassandra.sln
 install:
+  - ./scassandra_install.sh
```
landim: Agree. Moving the files on next commits.

src/.nuget/NuGet.targets
landim: Reverting this line.

src/Cassandra.IntegrationTests/packages.config
```diff
@@ -2,4 +2,6 @@
 <packages>
   <package id="Moq" version="4.2.1402.2112" targetFramework="net40" />
   <package id="NUnit" version="3.4.1" targetFramework="net40" requireReinstallation="true" />
+  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" />
+  <package id="System.Net.Http" version="4.3.0" targetFramework="net45" />
```
jorgebay: I think this is not needed as its a .NET Framework 4.5+ assembly...

build/scassandra_install.sh
```diff
@@ -0,0 +1,28 @@
...
+  echo "creating scassandra dir $scassandra_path"
+  wget -O scassandra-server-1.0.10.tar.gz https://github.com/scassandra/scassandra-server/archive/1.0.10.tar.gz
+  tar -C "$dep_dir/" -zxf scassandra-server-1.0.10.tar.gz
+fi
+
+scassandra_build_path="$scassandra_path/server/build/libs"
+if [ ! -d $scassandra_build_path ]; then
+  echo "Building scassandra... pushd $scassandra_path"
+  pushd $scassandra_path
+  ./gradlew server:fatJar
```
jorgebay: Can we upload and reuse `scassandra-server_2.11-1.0.11-SNAPSHOT-standalone.jar` instead of building it every time?

It's adding ~5m to the build.

landim: the stanalone jar is 25mb. So we should choose between save space or save time.

jorgebay: I would say save time, also we can benefit from the appveyor cache.

.travis.yml
```diff
@@ -21,6 +21,7 @@ branches:
 
 solution: src/Cassandra.sln
 install:
+  - ./build/scassandra_install.sh
```
jorgebay: No need to install scassandra on Travis: We use scassandra for integration tests, which are not run on Travis.

build/scassandra_install.sh
jorgebay: After talking to @landim, I now understand that its going to be cached by appveyor after a successful run, so its not a problem to build it when there is a cache miss.
✅

src/Cassandra.IntegrationTests/TestClusterManagement/SCassandraManager.cs
```diff
@@ -0,0 +1,165 @@
...
+    public class SCassandraManager
+    {
+
+        private static Process _scassandraProcess;
+
+        public static void Start()
+        {
+            Stop();
+            _scassandraProcess = new Process();
+            var jarPath = Environment.GetEnvironmentVariable("SCASSANDRA_JAR");
```
landim: done

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -115,5 +117,63 @@ private Task<string[]> ExecuteMultiple(ITestCluster testCluster, Session session
...
+                //remove one connection
+                var ports = SCassandraManager.GetListOfConnectedPorts().Result;
+                SCassandraManager.DropConnection(ports[ports.Length-1]).Wait();
+                Thread.Sleep(500); //wait few miliseconds to update state of driver
+                Assert.AreEqual(3, pool.OpenConnections);
+                Assert.IsTrue(h1.IsUp);
+                //remove one connection
+                ports = SCassandraManager.GetListOfConnectedPorts().Result;
+                SCassandraManager.DropConnection(ports[ports.Length - 1]).Wait();
+                Thread.Sleep(500); //wait few miliseconds to update state of driver
```
jorgebay: As in Jenkins is still failing, you can try using `TestHelper.WaitUntil()` to see if its received after a long period of time.
If it isn't, try enabling the heartbeat (to something like 2 seconds).

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -152,27 +152,30 @@ public void MarkHostDown_PartialPoolConnection()
...
                 TestHelper.WaitUntil(() => pool.OpenConnections == 1);
                 Assert.AreEqual(1, pool.OpenConnections);
                 Assert.IsTrue(h1.IsUp);
                 //remove one connection
                 ports = SCassandraManager.GetListOfConnectedPorts().Result;
                 SCassandraManager.DropConnection(ports[ports.Length - 1]).Wait();
+                Assert.Throws<NoHostAvailableException>(() => session.Execute("SELECT * FROM system.local"));
                 TestHelper.WaitUntil(() => pool.OpenConnections == 0);
                 Assert.AreEqual(0, pool.OpenConnections);
-                Assert.Throws<NoHostAvailableException>(() => session.Execute("SELECT * FROM system.local"));
```
jorgebay: I think this is a valid assertion, we should not remove it.

jorgebay: I see you move it up, sorry: nevermind

src/Cassandra.IntegrationTests/TestClusterManagement/SCassandraManager.cs
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+
+        public static void Stop()
+        {
+            if (_scassandraProcess == null) return;
+
+            try
+            {
+                _scassandraProcess.Kill();
```
jorgebay: Instead of using `Kill` and `WaitForExit()`, I think it should be `process.Dispose()`.

jorgebay:
I think we were not considering the connection made by the `ControlConnection`, that leaves a total of 5 connections: 4 from the pool and 1 from the `ControlConnection`.

The solution would be to either use 2 hosts and use the host not used by the control connection or take that connection into account.

I've made a concept of the second approach here: https://github.com/datastax/csharp-driver/tree/scassandra2

src/Cassandra.IntegrationTests/TestClusterManagement/SCassandraManager.cs
landim: Test it with Dispose() but it is not killing the process (I am using windows 10). The only way to kill to process was using Kill()

jorgebay:
Closing this as it's not intended for master any more.

PR #277, opened 2017-02-07
Mapper: Use a field and atomic increments to count prepared queries

jorgebay:


landim:
The code looks good to me. I will rebase with mster as soon as we have the appveyor build fix.

PR #282, opened 2017-02-13
Enable cache for appveyor build

landim:


jorgebay:
It looks like the test issue could be related to https://issues.apache.org/jira/browse/CASSANDRA-9646.
Splitting the test and expecting more than 1 event for `CREATE TABLE` should fix it. 

For the ccm related issue: it looks like you are using the `deps` dir that is cached for cloning ccm. Using another directory for ccm that is not cached should fix it.

PR #283, opened 2017-02-13
ProtocolVersion enum and decrease protocol versions

jorgebay:
Introduce `ProtocolVersion` enum.

After CASSANDRA-11464, the server can respond with the same unsupported version as provided. In that case, decrease the protocol version manually.
Tested against C* versions lower and higher than 3.0.9.


jorgebay:
Includes doc (CSHARP-459)

landim:
👍

PR #285, opened 2017-02-20
Parse custom type names

jorgebay:


landim:
LGTM, only added some checks regarding the other properties of the class ColumnDesc

PR #287, opened 2017-03-01
Timestamp generator

jorgebay:


src/Cassandra.IntegrationTests/Core/TimestampTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                CREATE TABLE {0}(
+                id uuid PRIMARY KEY,
+                label text);", "tbl_timestamp_test"));
+                var now = DateTimeOffset.Parse("2017-01-01T00:00:00.000+00:00");
+                var id = Guid.NewGuid();
+                var stmt = new SimpleStatement("INSERT INTO tbl_timestamp_test (id, label) VALUES (?,?) USING TIMESTAMP ?",
+                    id, "test label", now.Ticks);
+                session.Execute(stmt);
+                // The driver should use the generator against C* 2.1+
+                Assert.AreEqual(CassandraVersion < new Version(2, 1) ? 0 : 4, generator.GetCounter());
```
landim: In that case, which "USING TIMESTAMP" at cql, the timestamp generator must tick or just ignore?

jorgebay: If the protocol version is lower than 3, the driver will not generate a new timestamp.
The driver doesn't parse the query and it's unaware if the query contains `USING TIMESTAMP` (provided by the user only).

src/Cassandra.IntegrationTests/Core/TimestampTests.cs
```diff
@@ -33,6 +34,59 @@ public void Cluster_Uses_Provided_Timestamp_Generator()
             }
         }
 
+        [Test]
+        public void Should_Use_Statement_Timestamp_Precedence_Over_Cluster_Timestamp_Generator()
```
jorgebay: Nice!

src/Cassandra.IntegrationTests/Core/TimestampTests.cs
jorgebay: We could retrieve the stored timestamp using `writetime()` CQL function to know that `USING TIMESTAMP ?` is the value that was used but that would be a test for the server behaviour (not the driver).

I say we should remove this test.

PR #288, opened 2017-03-03
Linq: Avoid compiling lambdas whenever possible

jorgebay:
- Avoid compiling lambdas by using property and field methods.
- Set string builders capacity for Linq to CQL query generation.

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs
@@ -259,6 +259,69 @@ namespace Cassandra.Tests.Mapping.Linq
             Assert.AreEqual(@"SELECT ""id"" FROM ""ks1"".""tbl1"" WHERE ""id"" = ?", query);
             CollectionAssert.AreEqual(parameters, new object[] { 199 });
         }
+
+        private int _instanceField = 18;
+        private bool InstanceProperty { get; set; }
+
+        private class ClassWithPublicField
+        {
+            public Guid GuidField;
+            public static decimal DecimalStaticField = 10M;
+            public const double DoubleConstant = 3.1D;
+        }
+
+        [Test]
+        public void Select_With_All_Possible_Member_Expression_In_Where_Clause()
+        {
+            string query = null;
+            object[] parameters = null;
+            var session = GetSession((q, v) =>
+            {
+                query = q;
+                parameters = v;
+            }, TestDataHelper.CreateMultipleValuesRowSet(new[] { "id" }, new[] { 5000 }));
+            var map = new Map<AllTypesEntity>()
+                .ExplicitColumns()
+                .Column(t => t.DoubleValue, cm => cm.WithName("c0"))
+                .Column(t => t.StringValue, cm => cm.WithName("c1"))
+                .Column(t => t.IntValue, cm => cm.WithName("c2"))
+                .Column(t => t.BooleanValue, cm => cm.WithName("c3"))
+                .Column(t => t.DateTimeValue, cm => cm.WithName("c4"))
+                .Column(t => t.UuidValue, cm => cm.WithName("c5"))
+                .Column(t => t.DecimalValue, cm => cm.WithName("c6"))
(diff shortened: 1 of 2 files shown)


src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs
```diff
@@ -259,6 +259,69 @@ public void Select_With_Keyspace_Defined()
...
+            var linqQuery = from t in table
+                            where
+                                t.Int64Value == fieldWithProp.Int64Value &&
+                                t.DoubleValue == ClassWithPublicField.DoubleConstant &&
+                                t.StringValue == fieldWithProp.IntValue.ToString() &&
+                                t.IntValue == _instanceField &&
+                                t.BooleanValue == InstanceProperty &&
+                                t.DateTimeValue == date &&
+                                t.UuidValue == fieldWithField.GuidField &&
+                                t.DecimalValue == ClassWithPublicField.DecimalStaticField
```
landim: Nice coverage of differente types of fields, properties and visibility

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -72,15 +86,19 @@ internal class CqlExpressionVisitor : ExpressionVisitor
         private int _limit;
         private readonly List<Tuple<PocoColumn, object, ExpressionType>> _projections = new List<Tuple<PocoColumn, object, ExpressionType>>();
         private readonly List<Tuple<string, bool>> _orderBy = new List<Tuple<string, bool>>();
-        private readonly List<string> _selectFields = new List<string>();
+        private readonly List<string> _selectFields = new List<string>(DefaultClauseParameterCapacity);
```
landim: Nice to set an initial value, this will prevent the internal array of the list to be resized. Would make sense to use the number of columns of PocoData received at constructor of CqlExpressionVisitor instead of fixed value?

jorgebay: It could be a good idea but we don't know how many of those fields are going to be selected until we evaluate the query.

src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs
```diff
@@ -259,6 +259,122 @@ public void Select_With_Keyspace_Defined()
...
+                            select new { Age = (t.DoubleValue * 2), Id = t.Int64Value, DBool = !t.BooleanValue };
```
landim: I think it would be good to have an validation to ensure that this is not possible to do. Not possible in CQL queries.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
landim: I agree it has to be a guess. So, it comes down to: allocate more memory that would not be used or resizing the array if the select has more than 8 columns. I am 👍 with the fixed value.

src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs
jorgebay: That's good that you mention it, I was going to bring it up: I would like to refactor the Linq expression visitor to something similar to the one introduced in #291 but it's hard to know which expressions are supported and which aren't.

We could create a separate ticket to test all the currently supported expressions and the ones that aren't. It would be really nice to have some coverage on that side...

landim: 👍 on creating another ticket. I will add a full coverage on ticket's test, and put this test on ignore, sounds good?

jorgebay: I'd say we remove it as it's a type of expression that it's not supported.

landim: done

landim:
👍 . After build passes its good to go.

PR #289, opened 2017-03-03
Update build matrix to use C* 3.10

landim:


jorgebay:
Makes sense to update it to 3.10, can you also modify the `appveyor.yml` to point to 3.10?

Another thing, I don't see this branch on the Jenkins builds...

PR #290, opened 2017-03-09
Extended retry policy

jorgebay:


src/Cassandra/Requests/RequestHandler.cs
```diff
@@ -71,12 +71,12 @@ public RequestHandler(ISession session, Serializer serializer, IRequest request,
...
             Statement = statement;
             Policies = _session.Cluster.Configuration.Policies;
-            RetryPolicy = session.Cluster.Configuration.Policies.RetryPolicy;
+            RetryPolicy = Policies.ExtendedRetryPolicy;
             if (statement != null)
             {
                 if (statement.RetryPolicy != null)
                 {
-                    RetryPolicy = statement.RetryPolicy;   
+                    RetryPolicy = statement.RetryPolicy.Wrap(Policies.ExtendedRetryPolicy);
```
landim: Minor suggestion, to avoid nested ifs. Could it be 
`if (statement != null && statement.RetryPolicy != null)`

landim:
Included a PR with tests on https://github.com/datastax/csharp-driver/pull/292 using scassandra to check if ExtendedRetryPolicy is capturing the others server exceptions: OverloadedException, ServerErrorException and IsBootstrapingException

src/Cassandra/Requests/RequestHandler.cs
jorgebay: Makes sense, I will change it.

jorgebay:
Rebased against master.

jorgebay:
I'll merge this and focus on #293.

PR #291, opened 2017-03-17
Linq GROUP BY support

jorgebay:
Using #288 as base branch.

### DIFF
--- a/src/Cassandra.IntegrationTests/Linq/LinqMethods/GroupByTests.cs
+++ b/src/Cassandra.IntegrationTests/Linq/LinqMethods/GroupByTests.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Cassandra.Mapping;
+using Cassandra.Data.Linq;
+using Cassandra.IntegrationTests.TestBase;
+using Cassandra.Tests.Mapping.Pocos;
+using NUnit.Framework;
+
+namespace Cassandra.IntegrationTests.Linq.LinqMethods
+{
+    [Category("short"), TestCassandraVersion(3, 10)]
+    public class GroupByTests : SharedClusterTest
+    {
+        protected override string[] SetupQueries
+        {
+            get
+            {
+                return new []
+                {
+                    "CREATE TABLE sensor_data (sensor_id text, bucket text, timestamp timeuuid, value double," +
+                    " PRIMARY KEY ((sensor_id, bucket), timestamp))",
+                    "INSERT INTO sensor_data (sensor_id, bucket, timestamp, value) VALUES ('sensor1', 'bucket1', now(), 1.5)",
+                    "INSERT INTO sensor_data (sensor_id, bucket, timestamp, value) VALUES ('sensor1', 'bucket1', now(), 2)",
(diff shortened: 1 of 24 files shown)


src/Cassandra.IntegrationTests/Linq/LinqMethods/GroupByTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                    Min = g.Min(i => i.Value),
+                    Max = g.Max(i => i.Value),
+                    g.Key.Bucket
+                });
+            var results = linqQuery.Execute().ToArray();
+            Assert.AreEqual(1, results.Length);
+            var aggregation = results[0];
+            Assert.AreEqual("sensor1", aggregation.Id);
+            Assert.AreEqual("bucket1", aggregation.Bucket);
+            Assert.AreEqual(aggregation.Avg, 2);
```
landim: Use the expected value as first parameter of 
`Assert.AreEqual( int expected, int actual );`

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
```diff
@@ -67,6 +67,10 @@ private static void LoadCassandraVersion()
             //in case there is a version label like rc1 / beta1
             versionText = versionText.Split('-')[0];
             _cassandraVersion = Version.Parse(versionText);
+            if (_cassandraVersion.Build == -1)
+            {
+                _cassandraVersion = new Version(_cassandraVersion.Major, _cassandraVersion.Minor, 0);
```
landim: Nice way to preventing build number mistakes

src/Cassandra/Data/Linq/CqlQueryExtensions.cs
```diff
@@ -95,6 +106,34 @@ public static CqlQuery<TSource> Where<TSource>(this CqlQuery<TSource> source, Ex
...
+        ///  Returns a CqlQuery which after execution returns grouped sequence of values based on a predicate.
+        ///  To execute this CqlQuery use <c>Execute()</c> method.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of source.</typeparam>
+        /// <typeparam name="TKey">The type of the value returned by selector.</typeparam>
+        /// <param name="source">The CqlQuery&lt;TSource&gt; to filter.</param>
+        /// <param name="predicate">A function to test each element for a condition.</param>
+        /// <returns>a CqlQuery&lt;TSource&gt; which after execution will return an IEnumerable&lt;TSource&gt;
+        /// that contains elements from the input sequence that satisfy the condition.</returns>
+        public static CqlQuery<IGrouping<TKey, TSource>> GroupBy<TKey, TSource>(this CqlQuery<TSource> source,
```
landim: just thinking out loud:
It could be possible to use another class instead of IGrouping (maybe ICassandraIGrouping) only implementing the aggregation methods supported by cassandra. So it would have an verification at compile time. But it could be an huge effort to implement it.

landim:
There is an integration test failing, but its not related (intermittent fail).
But everything looks great. Included an test for exception cases.

src/Cassandra/Data/Linq/CqlQueryExtensions.cs
jorgebay: I see what you mean... if `ICassandraGrouping` inherited from `IGrouping` the extension methods would still apply, if `ICassandraGrouping` didn't inherit it would be a hell of an effort to implement those methods...

PR #294, opened 2017-03-25
Add Get method to allow checks before addition of new mappings.

satyajitps:
Currently, it is not possible to check whether a mapping for a POCO exists in a given MappingConfiguration.

For example, when updating the Global instance, consumers are forced to implement their own tracking to prevent exceptions that get thrown when adding a mapping that already exists.

This PR will simplify a check-before-add scenario.

### DIFF
--- a/src/Cassandra.Tests/Mapping/MappingConfigurationTests.cs
+++ b/src/Cassandra.Tests/Mapping/MappingConfigurationTests.cs
@@ -1,9 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using Cassandra.Mapping;
+﻿using Cassandra.Mapping;
 using Cassandra.Mapping.TypeConversion;
+using Cassandra.Tests.Mapping.FluentMappings;
+using Cassandra.Tests.Mapping.Pocos;
 using NUnit.Framework;
 
 namespace Cassandra.Tests.Mapping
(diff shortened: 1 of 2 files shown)


src/Cassandra.Tests/Mapping/MappingConfigurationTests.cs
```diff
@@ -22,5 +20,22 @@ public void ConvertTypesUsing_Creates_Uses_MapperFactory_Instance()
             //New instance of the mapper factory
             Assert.AreNotSame(originalMapperFactory, config.MapperFactory);
         }
+
+        [Test]
+        public void Get_FromGlobalConfig_Returns_Mapping_IfExists()
+        {
+            var userMapping = new FluentUserMapping();
+            MappingConfiguration.Global.Define(userMapping);
```
jorgebay: Can you use a local instance of `MappingConfiguration`? We should avoid using the global instance on the tests to avoid messing up the other test fixtures.

jorgebay:
The patch is straightforward and besides the minor nit, the tests look good!

src/Cassandra.Tests/Mapping/MappingConfigurationTests.cs
satyajitps: Makes complete sense. Updated the PR.

landim:
Looks good. Integration tests are not needed, since the unit tests, in this case, already checks the behavior.

jorgebay:
Landed in d1ce3c5f7dd2d6e72c3aa430f17ab520d745368c, thanks @satyajitps !

PR #297, opened 2017-04-21
Cluster.Init(): Use connect and read timeouts wrapped in NoHostAvailableException

jorgebay:
Make `ControlConnection` initialization fully asynchronous and avoid using abort timeout as a mechanism to control socket connect timeout and STARTUP read timeout.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
@@ -3,8 +3,6 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Security.Cryptography;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Cassandra.IntegrationTests.Policies.Util;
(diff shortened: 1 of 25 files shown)


src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -176,5 +174,48 @@ public void MarkHostDown_PartialPoolConnection()
...
+            var cluster = Cluster.Builder()
+                                 .AddContactPoint(testCluster.InitialContactPoint)
+                                 .Build();
+            //initially it will throw as there is no node reachable
+            Assert.Throws<NoHostAvailableException>(() => cluster.Connect());
+
+            // wait for the node to be up
+            testCluster.Start(1);
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
```
landim: Could use Stopwatch.StartNew() to create instance and starts. Also we need to stopped after the succeeded connection and define a reasonable time to assert around it.

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -176,5 +174,48 @@ public void MarkHostDown_PartialPoolConnection()
...
+        [Test]
+        public void Cluster_Initialization_Recovers_From_NoHostAvailableException()
+        {
+            var testCluster = TestClusterManager.CreateNew();
+            testCluster.StopForce(1);
+            var cluster = Cluster.Builder()
+                                 .AddContactPoint(testCluster.InitialContactPoint)
+                                 .Build();
+            //initially it will throw as there is no node reachable
+            Assert.Throws<NoHostAvailableException>(() => cluster.Connect());
```
landim: Regarding the issue on jira, I think we should add a test which includes N attempts to connect and monitor the memory increase. What do you think?

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
jorgebay: I've left it unintentionally, I'll remove it.

src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
jorgebay: I've just included a test for something similar to what the user reported.

landim:
Tests at https://github.com/datastax/csharp-driver/pull/301

src/Cassandra/TcpSocket.cs
```diff
@@ -200,21 +195,24 @@ public Task<bool> Connect()
...
                 () => new TimeoutException("The timeout period elapsed prior to completion of SSL authentication operation."));
-#pragma warning disable 4014
+
             sslStream.AuthenticateAsClientAsync(targetHost,
-                                                SSLOptions.CertificateCollection,
-                                                SSLOptions.SslProtocol,
-                                                SSLOptions.CheckCertificateRevocation)
+                         SSLOptions.CertificateCollection,
+                         SSLOptions.SslProtocol,
+                         SSLOptions.CheckCertificateRevocation)
```
landim: Identation only change

jorgebay: I'll restore the previous formatting

PR #298, opened 2017-04-24
Mapper: support options at batch level

jorgebay:


landim:
I will squash my commits as soon as build fixes

landim:
The jenkins is breaking because at mono the exception throw is UnavailableException, but at windows it throws a WriteTimeoutException.

jorgebay:
I think the test can be simplified: if you use a single node cluster, create a keyspace with replication factor 3 and try to write with consistency > 1, it will throw an unavailable exception.

PR #301, opened 2017-04-28
Cluster Init/Connect : Memory leak test

landim:
Attempt to automate some memory profiling tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,10 +27,16 @@ environment:
     - TARGET: net452
       CASSANDRA_VERSION: 2.2.7
       CI_TYPE: INTEGRATION
+    - TARGET: net452
+      CASSANDRA_VERSION: 3.10
+      CI_TYPE: MEMORY
 
 install:
   - ps: .\build\appveyor_install.ps1
   - ps: .\build\scassandra_install.ps1
+
+before_build:
+  - ps: dotnet restore
 
 build:
   project: src/Cassandra.sln
@@ -47,9 +53,16 @@ test:
       - duration
 
 test_script:
-  - dotnet restore
-  - IF "%CI_TYPE%" == "UNIT" dotnet test src\Cassandra.Tests -c Release -f %TARGET% --result=TestResult.xml
-  - IF "%CI_TYPE%" == "INTEGRATION" dotnet test src\Cassandra.IntegrationTests -c Release -f %TARGET% --where "cat=short" --result=TestResult.xml
+  - ps: >-
+      If ("$($env:CI_TYPE)" -eq "UNIT") {
+        dotnet test src\Cassandra.Tests -c Release -f $env:TARGET --result=TestResult.xml
+      }
+      If ("$($env:CI_TYPE)" -eq "INTEGRATION") {
+        dotnet test src\Cassandra.IntegrationTests -c Release -f $env:TARGET --where "cat=short" --result=TestResult.xml
+      }
+      If ("$($env:CI_TYPE)" -eq "MEMORY") {
+        .\build\memory_tests.ps1
+      }
 
 after_test:
   - ps: >-
--- a/build/appveyor_install.ps1
+++ b/build/appveyor_install.ps1
@@ -79,10 +79,17 @@ If (!(Test-Path $sslPath)) {
   Copy-Item "$($env:CCM_PATH)\ssl" -Destination $sslPath -Recurse
 }
 
-$sslPath="$($env:HOMEPATH)\ssl"
-If (!(Test-Path $sslPath)) {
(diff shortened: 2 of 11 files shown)


jorgebay:
Rebased after merging #297.

appveyor.yml
```diff
@@ -47,18 +53,28 @@ test:
...
         $wc = New-Object 'System.Net.WebClient'
-        $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit3/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\TestResult.xml))
+        $wc.UploadFile("https://ci.appveyor.com/testresults/nunit3/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\TestResult.xml))
       }
 
 on_failure:
   - ps: >-
       Write-Host "Build failed"
+
+#on_finish:
```
jorgebay: Left commented out :)

build/appveyor_install.ps1
```diff
@@ -79,10 +79,17 @@ If (!(Test-Path $sslPath)) {
   Copy-Item "$($env:CCM_PATH)\ssl" -Destination $sslPath -Recurse
 }
 
-$sslPath="$($env:HOMEPATH)\ssl"
```
jorgebay: Fixes duplication introduced by me here: https://github.com/datastax/csharp-driver/commit/f21280e4a1bb376596278ddd023e735d7029a996#commitcomment-22069993

build/memory_tests.ps1
```diff
@@ -0,0 +1,6 @@
+$nunitrunner = "C:/Tools/NUnit3/nunit3-console.exe";
```
jorgebay: It would be nice to use an env variable for nunit 3 console path, to be able to reuse this script (ie: locally) regardless of the path.

src/Cassandra.IntegrationTests/Core/MemoryLeakTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                TestContext.WriteLine("Before cluster creation");
+                var cassandraPackage = memory.GetObjects(presetQueryCassandraObjects);
+                var socketObjects = memory.GetObjects(where => where.Type.Is<Socket>());
+                TestContext.WriteLine("=== Cassandra driver Object count: " + cassandraPackage.ObjectsCount);
+                TestContext.WriteLine("=== Number of Socket objects: " + socketObjects.ObjectsCount);
+            });
+
+            var socketOptions = new SocketOptions().SetReadTimeoutMillis(1).SetConnectTimeoutMillis(1);
+            var builder = Cluster.Builder()
+                                    .AddContactPoint("1.1.1.1")
```
jorgebay: Use `TestHelper.UnreachableHostAddress` when possible.

jorgebay:
I think having this test automated is awesome. I've added a few nits, but overall looks good.
The test covers the situation described by the user.

In the future, we could add a one against a real cluster but I think maybe a more suitable place would be the endurance test: just recording memory growth somewhere, wdyt?

build/memory_tests.ps1
```diff
@@ -0,0 +1,14 @@
+<# dotMemoryUnit requires the pull path of nunit3-console.
```
jorgebay: The script below wasn't making much sense to me until I read the comment specifying that `dotMemoryUnit` requires the full path. 👍

jorgebay:
There is a remaining issue related to the `JetBrains.DotMemoryUnit` dependency, that is causing [TravisCI to fail](https://travis-ci.org/datastax/csharp-driver/builds/231299627).
You need to add the reference in the `Cassandra.IntegrationTests.csproj` file as well, in the path `..\packages\JetBrains.DotMemoryUnit.2.3.20160517.113140\Lib\dotMemory.Unit.dll`

I know its a nightmare to add new references in the project thanks to the dual toolchain, but it should be straightforward once we migrate to VS2017 (see CSHARP-548).

src/Cassandra.IntegrationTests/Core/MemoryLeakTests.cs
```diff
@@ -0,0 +1,102 @@
+﻿#if !(NETCORE || __MonoCS__)
```
jorgebay: Avoid using `__MonoCS__` as it's obsolete once mono uses Roslyn.

jorgebay:
Fails on TravisCI under mono 5 but not on Jenkins... 😡

jorgebay:
I'm not able to reproduce on OSX under Mono 5...

landim:
Job Creator build failures are intermittent, we need to create an test item regarding this behavior.

jorgebay:
ok, I've created CSHARP-571 with some of the failures we've encountering.

PR #305, opened 2017-05-31
Fix CqlExpressionVisitor when RuntimePropertyInfo

landim:
Linq evaluation throws exception when CqlExpressionVisitor tries to cast member to FieldInfo.
Reported on CSHARP-577

### DIFF
--- a/src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
+++ b/src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
@@ -1118,19 +1118,32 @@ namespace Cassandra.Data.Linq
 
         private static object GetFieldValue(MemberExpression node)
         {
-            var fieldInfo = (FieldInfo)node.Member;
+            Func<object, object> getValueDel;
+            switch (node.Member.MemberType)
+            {
+                case MemberTypes.Field:
+                    var fieldInfo = (FieldInfo)node.Member;
+                    getValueDel = fieldInfo.GetValue;
+                    break;
+                case MemberTypes.Property:
+                    var propertyInfo = (PropertyInfo) node.Member;
+                    getValueDel = propertyInfo.GetValue;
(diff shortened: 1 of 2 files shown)


src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -1118,19 +1122,32 @@ private static object GetClosureValue(MemberExpression node)
...
+                case MemberTypes.Field:
+                    var fieldInfo = (FieldInfo)node.Member;
+                    getValueDel = fieldInfo.GetValue;
+                    break;
+                case MemberTypes.Property:
+                    var propertyInfo = (PropertyInfo) node.Member;
+                    getValueDel = propertyInfo.GetValue;
+                    break;
+                default:
+                    throw new NotSupportedException(node.Member.MemberType.ToString());
```
jorgebay: Use `CqlLinqNotSupportedException` instead.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -1118,19 +1122,32 @@ private static object GetClosureValue(MemberExpression node)
 
         private static object GetFieldValue(MemberExpression node)
         {
-            var fieldInfo = (FieldInfo)node.Member;
+            MemberOrPropertyGetValue getValueDel;
```
jorgebay: If you want to avoid creating the delegate, you can reuse the generic one: `Func<object, object>`

jorgebay:
Besides the minor nits, the fix and test looks good!
It would be nice to could include tests for expressions with static properties, nested constants, ...

landim:
There is an error while instantiating an object at where clause. I am looking into it.

jorgebay:
Fixed in CSHARP-577.

PR #306, opened 2017-06-01
use integral calculation to prevent wrong fractional rounding

Kran440:
Hi, i meet a bug when try to filter values by DateTimeOffset. Say we have a class 

```csharp
[Table("buckets")]
public class Bucket
{
	[PartitionKey]
	public int Index { get; set; }
	
	[ClusteringKey]
	public DateTimeOffset Date { get; set; }
}
```

insert some values

```csharp
// Table<Bucket> table defined above
var date = new DateTimeOffset(2017, 1, 1, 0, 0, 0, 0, TimeSpan.Zero); // 01.01.2017 0:00:00 +00:00
table.Insert(new Bucket { Index = 1, Date = date }).Execute();
table.Insert(new Bucket { Index = 1, Date = date.AddDays(1) }).Execute();
```

Now we have two records.

```csharp
var endOfDay = date.AddTicks(TimeSpan.TicksPerDay - 1); // last tick of date
var result = table.Where(bucket => bucket.Index == 1 && bucket.Date <= endOfDay).Execute(); // returns both values!
```

So my changes eliminate this problem.

### DIFF
--- a/src/Cassandra/Serialization/Primitive/DateTimeOffsetSerializer.cs
+++ b/src/Cassandra/Serialization/Primitive/DateTimeOffsetSerializer.cs
@@ -26,12 +26,14 @@ namespace Cassandra.Serialization.Primitive
 
         internal static DateTimeOffset Deserialize(byte[] buffer, int offset)
         {
-            return UnixStart.AddMilliseconds(BeConverter.ToInt64(buffer, offset));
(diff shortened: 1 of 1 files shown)


jorgebay:
Nice catch @Kran440 ! makes perfect sense.

I'll wait for CI builds to finish and I'll merge it.

PR #307, opened 2017-06-08
Fix TaskCompletionSourceWithTimeout

jorgebay:
Fixes an issue that's very unlikely to appear in the wild but that's showing up on Jenkins after `Should_Not_Leak_Connections_When_Node_Unreacheable_Test` was introduced:

```
System.NullReferenceException
    at Cassandra.Tasks.TaskHelper.<>c__DisplayClass19_0`1.<TaskCompletionSourceWithTimeout>b__0(Object _)
    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
```
The timer was elapsing before the field was set. After the fix, I've manually started the build 7 times to make sure it's gone.

### DIFF
--- a/src/Cassandra/Tasks/TaskHelper.cs
+++ b/src/Cassandra/Tasks/TaskHelper.cs
@@ -344,6 +344,10 @@ namespace Cassandra.Tasks
             {
                 var t = timerWrapper[0];
                 Interlocked.MemoryBarrier();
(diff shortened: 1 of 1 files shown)


landim:
Its a good catch (those time conflict issues are always hard to identify) and neat fix! 👍

PR #308, opened 2017-06-13
Fix missing ConfigureAwait on Reconnect

LukeTillman:
For background, see this StackOverflow question:

https://stackoverflow.com/questions/44531557/datastax-driver-3-3-0-deadlocking-on-connect-to-cluster

I'm not 100% sure this is what's causing the issue reported in the SO question, but I suspect as a library, we should be using `ConfigureAwait(false)` pretty much everywhere.

### DIFF
--- a/src/Cassandra/ControlConnection.cs
+++ b/src/Cassandra/ControlConnection.cs
@@ -257,7 +257,7 @@ namespace Cassandra
                     //Control connection is being disposed
                 }
             }
-            return await tcs.Task;
+            return await tcs.Task.ConfigureAwait(false);
         }
 
         internal async Task Refresh()


jorgebay:
Nice catch! I've forgot to add it. 

About the stackoverflow question, it looks like the task returned by`ControlConnection.Init()` is not getting completed, I'll try to ask the user for more info to see what could be happening.

lordfist:
This did not solve the issue with deadlocking. I've tried to build from your source and same thing happens. Driver gets stuck with 5 waiting tasks that are blocked by awaiting call for Cassandra.ControlConnection.Init()
imgdeadlock22

PR #309, opened 2017-06-15
Use ConfigureAwait() for all awaited Tasks

jorgebay:
Add remaining ConfigureAwait() for ControlConnection and TcpSocket.
Include a test that checks the source code and looks for missing ConfigureAwait() calls.

Fixes CSHARP-579.


jorgebay:
cc @LukeTillman

landim:
I tested locally on a Windows Forms application (using the solution on https://github.com/landim/csharp-driver-winform-test) and the fix worked. 👍

PR #310, opened 2017-06-15
Add Allow filtering on scalar Count() method.  

francischie:
Add Allow filtering on scalar Count() method.  ALLOW FILTERING has unpredictable performance but if the partition key is already known not to grow that much should still be allowed instead performing on the client side. CQL already allowed this.

example

This query is allowed on CQL but can't be done on current LINQ.

SELECT Count (*) FROM sometable WHERE partition_column = ? AND non_primary_key = ? ALLOW FILTERING;

LINQ
Session.Table<SomeTable>().Where(a => a.PartitionColumn == {some value} && a.NonPrimaryKey == {some value}). AllowFiltering().Count();

The fix will generate equivalent CQL above.


jorgebay:
lgtm! I'll add a few more tests on Monday and merge it.

jorgebay:
I've added a test on #312, you can continue tracking your contribution there.

PR #311, opened 2017-06-16
Migrate to dotnet 1.0.1+ tools

jorgebay:
Migrate the solution and projects files to dotnet tools v1.0.1+.
Update Moq library used for test to a newer version that does support .NET Core.
Adapt travis and appveyor builds.


jorgebay:
Jenkins image have to be updated to include latest dotnet tools.

appveyor.yml
```diff
@@ -35,43 +36,18 @@ install:
...
+  - ps: dotnet build src\Cassandra.sln -c Release
 
 test_script:
   - ps: >-
-      If ("$($env:CI_TYPE)" -eq "UNIT") {
-        dotnet test src\Cassandra.Tests -c Release -f $env:TARGET --result=TestResult.xml
-      }
-      If ("$($env:CI_TYPE)" -eq "INTEGRATION") {
-        dotnet test src\Cassandra.IntegrationTests -c Release -f $env:TARGET --where "cat=short" --result=TestResult.xml
+      If ("$($env:CI_TYPE)" -ne "MEMORY") {
```
landim: Nice way to avoid the old 3 "if"s

appveyor.yml
```diff
@@ -35,43 +37,21 @@ install:
...
-
-after_test:
-  - ps: >-
-      if($true)
-      {
-        $wc = New-Object 'System.Net.WebClient'
-        $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit3/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\TestResult.xml))
-      }
-
+      xml tr tools\trx-to-junit.xslt TestResult.xml > ResultsJUnit.xml
```
jorgebay: This should be moved to the section `on_finish`.
On appveyor, the build is shown as successful or failed depending on the exit code of the `ps` script.

TLDR: The `ps` script is evaluated block by block and when the last statement is executed, the exit code for the script is returned. In this case, the last statement would be `$wc.UploadFile()`: as it succeeds, the build is shown as successful, even though `dotnet test` exited with code != 0. See https://ci.appveyor.com/project/DataStax/csharp-driver/build/master-1166/job/510r3hd6flkxh26i for example.

jorgebay:
Good progress on the Jenkins image and appveyor builds @landim, kudos!

Even though I've removed from TravisCI (getting the mono runtimes was too much for each build on Travis), on Jenkins I think we should still build with mono. We could update to mono 5 and the commands used should be:
- `msbuild src/Cassandra.sln /v:m /p:Configuration=Release`
- `nuget install NUnit.ConsoleRunner -Version 3.6.1 -OutputDirectory testrunner`
- `mono ./testrunner/NUnit.ConsoleRunner.3.6.1/tools/nunit3-console.exe src/Cassandra.Tests/bin/Release/net452/Cassandra.Tests.dll --where ...`

Note that it should point to `Release/net452` assemblies for mono and that `xbuild` has been deprecated in favor of `msbuild`.

jorgebay:
Its looking good, Jenkins reports for mono and dotnet are great!

I'm merging it.

PR #312, opened 2017-06-19
CSHARP-582: Linq: Add ALLOW FILTERING on scalar Count() method

jorgebay:
Includes an additional test for #310.


landim:
Rebased with master

src/Cassandra.IntegrationTests/Linq/LinqMethods/Count.cs
```diff
@@ -66,6 +83,18 @@ public void LinqCount_Where_Async()
             Assert.AreEqual(1, count);
         }
 
+        #pragma warning disable 618
```
jorgebay: Instead of declaring a new entity, reuse one if possible and define mappings per tests using the fluent mapping (not the attributes).

On the other hard, we should stop using Linq attributes and use `Cassandra.Mapping.Attributes` instead.

landim:
I will squash those test commits before merge.

jorgebay:
Looks good to me!
I'll do a merge commit to acknowledge the 3 different commits from 3 different authors :)
Thanks @francischie !

PR #313, opened 2017-06-19
Refactored TcpSocket.Connect to use BeginConnect/EndConnect for Mono.

sotodel:
[This old Mono bug](https://bugzilla.xamarin.com/show_bug.cgi?id=23700) causes a fatal problem when the current `TcpSocket` implementation times out while attempting to connect.

When the `TaskCompletionSource` is completed with a timeout, the `Connection` code attempts to `Dispose` the corresponding `Socket`. This triggers the `SocketAsyncEventArgs` callback on an unmanaged thread. The first thing it does is check if the `Socket` was disposed and throws an exception if it is. Application code cannot intercept and handle that exception. It goes unhandled and kills the process.

Mono was patched, but for those that can't upgrade this problem remains. 

Instead, I propose this workaround that should exhibit the same _happy-path_ behavior but doesn't die under timeout. With this implementation, the `ObjectDisposedException` would be thrown on `EndConnect` which is handled by the `FromAsync` infrastructure (and propagated/ignored in the `Task` returned).

### DIFF
--- a/src/Cassandra/TcpSocket.cs
+++ b/src/Cassandra/TcpSocket.cs
@@ -113,34 +113,17 @@ namespace Cassandra
         /// <exception cref="SocketException">Throws a SocketException when the connection could not be established with the host</exception>
         public async Task<bool> Connect()
         {
-            var tcs = TaskHelper.TaskCompletionSourceWithTimeout<bool>(
-                Options.ConnectTimeoutMillis, 
-                () => new SocketException((int) SocketError.TimedOut));
-            var socketConnectTask = tcs.Task;
-            var eventArgs = new SocketAsyncEventArgs
+            var connectTask = Task.Factory.FromAsync(
+                                  (ipEndPoint, callback, state) => ((Socket)state).BeginConnect(ipEndPoint, callback, state),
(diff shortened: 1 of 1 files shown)


sotodel:
I'm going to need some help. I don't know these `dotnet` tools at all. The `BeginConnect` and `EndConnect` methods I tried to use seem to be unsupported in the `dotnet` version you've configured for build, `1.0.0-preview2-003131`. I tried to update that to something more recent, but there are other issues with the CLI tools (eg. not supporting `CodeTaskFactory`).

jorgebay:
> The `BeginConnect` and `EndConnect` methods I tried to use seem to be unsupported in the dotnet version you've configured for build, `1.0.0-preview2-003131`

APM methods (`Begin` / `End`) are not supported under .NET Standard 1.5, I forgot about it when we were discussing about this.
We can fix it by using conditional compilation but I would like to avoid it whenever possible, as duplicating code in such a sensitive part of the driver will make maintenance more difficult...

jorgebay:
We can look into ways to support older versions of Mono in the future.

PR #314, opened 2017-06-21
CSHARP-581 Linq Batch: allow setting the batch type

collinsauve:
Fixes https://datastax-oss.atlassian.net/browse/CSHARP-581

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqBatchStatementUnitTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqBatchStatementUnitTests.cs
@@ -1,0 +1,96 @@
+﻿using System;
+using System.Linq;
+using Cassandra.Data.Linq;
+using Cassandra.Mapping;
+using Cassandra.Tests.Mapping.Pocos;
+using NUnit.Framework;
+
+namespace Cassandra.Tests.Mapping.Linq
+{
+    public class LinqBatchStatementUnitTests : MappingTestBase
+    {
+
+        [Test]
+        [TestCase(null)]
+        [TestCase(BatchType.Logged)]
+        [TestCase(BatchType.Unlogged)]
+        public void DeleteBatch(BatchType? batchType)
+        {
+            BatchStatement statement = null;
+            var session = GetSession<BatchStatement>(new RowSet(), stmt => statement = stmt);
+            var map = new Map<AllTypesEntity>()
+                .ExplicitColumns()
+                .Column(t => t.StringValue, cm => cm.WithName("val"))
+                .Column(t => t.UuidValue, cm => cm.WithName("id"))
+                .PartitionKey(t => t.UuidValue)
+                .TableName("tbl1");
+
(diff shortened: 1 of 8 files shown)


jorgebay:
We need to add a test to cover the parameter.

collinsauve:
@jorgebay I added a couple Linq-to-Cql tests, however 2 things to note:

* I think I fixed a bug where it was previously writing "BEGIN COUNTERBATCH" and "APPLY COUNTERBATCH" with no space.  Can you verify that I've done the right thing here?
* Are there any Linq-to-Statement style tests where I can just add another?

jorgebay:
There are tests for [bound statements in the LinqToCqlSelectUnitTest class](https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.Tests/Mapping/Linq/LinqToCqlSelectUnitTests.cs#L47-L65) but there isn't for batch statements.

In this case, it should be something like (pseudo code):

```csharp
        protected ISession GetSession(Action<BatchStatement> callback)
        {
            var sessionMock = new Mock<ISession>(MockBehavior.Strict);
            sessionMock.Setup(s => s.Cluster).Returns((ICluster)null);
            sessionMock
                .Setup(s => s.ExecuteAsync(It.IsAny<BatchStatement>()))
                .Returns(() => TaskHelper.ToTask(new RowSet))
                .Callback(callback)
                .Verifiable();
            sessionMock
                .Setup(s => s.PrepareAsync(It.IsAny<string>()))
                .Returns<string>(query => TaskHelper.ToTask(GetPrepared(query)))
                .Verifiable();
            return sessionMock.Object;
        }

        [Test]
        public void Select_With_ConsistencyLevel()
        {
            BatchStatement statement = null;
            var session = GetSession(stmt => statement = stmt);
            var map = new Map<AllTypesEntity>()
                // ... TODO: define mappings
                .TableName("tbl1");
            var table = GetTable<AllTypesEntity>(session, map);
            //TODO: execute batch and assert state of statement field
        }
```

collinsauve:
I added a test that creates a linq batch with delete operations.   I believe this is enough to test the changes I'm actually making, but seems to lead me to create more tests for other operations.  If you like the path I'm on with this test and want me to create more I wouldn't mind adding them.

jorgebay:
Nice! `DeleteBatch` unit test is clear and concise, sure it would be nice to have more of them!

collinsauve:
Okay :) I added tests around updates as well.  I originally thought I'd be adding insert, but I guess that doesn't make sense in a linq batch.

jorgebay:
Great tests! I'll merge it.

> I originally thought I'd be adding insert, but I guess that doesn't make sense in a linq batch.

I'm not sure what you mean, it should be possible to add multiple insert statements in a Linq batch, after all they are all upserts.

jorgebay:
The functionality looks more than covered from the unit test side, @landim do you want to add integration tests?

landim:
@jorgebay I will add them

PR #315, opened 2017-06-21
CSHARP-577 InvalidCastException in Cassandra.Data.Linq

collinsauve:
Fixes https://datastax-oss.atlassian.net/browse/CSHARP-577


jorgebay:
Good fix @collinsauve , that way it can be nested arbitrarily.

PR #263, opened 2016-09-29
Type conversion for UDT fields and tests

jorgebay:
To track and include tests for #260 by @ElliotSis (CSHARP-381).

CC @landim

PR #317, opened 2017-06-22
UTDMap Change visibility and  virtual  some methods

tfscosta:
After having the same issue like the user in https://github.com/datastax/csharp-driver/pull/260 
And since the tests failed I need in my use case map some object to enum so i changed some methods to protected and add some virtual methods in the UTD class so I can make my own mapper

### DIFF
--- a/src/Cassandra/UdtMap.cs
+++ b/src/Cassandra/UdtMap.cs
@@ -14,12 +14,12 @@
 //   limitations under the License.
 //
 
-﻿using System;
+using System;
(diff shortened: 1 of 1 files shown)


jorgebay:
Thanks for your contribution, your patch has a straightforward approach. I think what you are proposing can be achieved by implementing your own `UdtSerializer`.
#263 tries to cover a generic conversion approach that could help users without the need to override all default mapping behaviour.

PR #319, opened 2017-06-26
Fix unit tests

jorgebay:

PR #320, opened 2017-06-26
Fix appveyor mem test report

landim:


landim:
@jorgebay  After this pass, I will merge it into https://github.com/datastax/csharp-driver/pull/319

jorgebay:
We can merge it into master directly.

landim:
Reopening to include memory leak tests report at appveyor

jorgebay:
It looks like there is some kind of issue with dotmemoryunit: https://ci.appveyor.com/project/DataStax/csharp-driver/build/fix_appveyor_test_report-1664/job/t38l2pwjnpanwti5

I was scratching my head with: `System.Net.Sockets.Socket.Receive()` in the stack trace, as we don't use the sync API, but the error shows that the issue on the dotmemoryunit side:

`No connection could be made because the target machine actively refused it 127.0.0.1:1644`

We don't use that port number :)

The job result is still success (should be red).

PR #321, opened 2017-06-26
CSHARP-586 Flip the WithDefaultTimestamp flag when using a timestamp …

collinsauve:
…generator

### DIFF
--- a/src/Cassandra/QueryProtocolOptions.cs
+++ b/src/Cassandra/QueryProtocolOptions.cs
@@ -161,6 +161,25 @@ namespace Cassandra
             var protocolVersion = wb.Serializer.ProtocolVersion;
             var flags = GetFlags();
 
+            // Determine timestamp
+            long? timestamp = null;
+            if (Timestamp.HasValue)
+            {
+                timestamp = TypeSerializer.SinceUnixEpoch(Timestamp.Value).Ticks / 10;
+            }
+            else if (protocolVersion.SupportsTimestamp() && _policies != null)
+            {
+                timestamp = _policies.TimestampGenerator.Next();
+                if (timestamp.Value == long.MinValue)
+                {
+                    timestamp = null;
+                }
+                else
+                {
+                    flags = flags | QueryFlags.WithDefaultTimestamp;
+                }
+            }
+
             if (protocolVersion != ProtocolVersion.V1)
             {
                 wb.WriteUInt16((ushort)Consistency);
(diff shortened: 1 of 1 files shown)


jorgebay:
Again, great findings @collinsauve

jorgebay:
I think we could simplify it by replacing the last constructor parameter of the `QueryProtocolOptions` to be the timestamp, instead of the policies.
That way, the the constructor caller should provide either the user timestamp or the generated timestamp.
`GetFlags()` should accept the protocol version and add the timestamp flag if the protocol version supports it and it was provided (user or generated), set the flag.

I've created a gist for that approach: https://gist.github.com/jorgebay/7f04eed5b68c2ca2d47507bc759e6ba0
Let me know what you think.

I've been reluctant to refactor `QueryProtocolOptions` class because it's part of the public API since v1.0 and removing it would be a breaking change, but I think its clear we must do so in the next major revision.

collinsauve:
That approach would work as well, and does look cleaner and more straightforward.  Would you consider merging my approach on a 3.x release and then switching to yours on a 4.x release?

jorgebay:
The gist I proposed does not include a breaking change as the constructor is internal.

jorgebay:
Just to clarify, I said

> I've been reluctant to refactor QueryProtocolOptions class because it's part of the public API since v1.0 and removing it would be a breaking change, but I think its clear we must do so in the next major revision.

To give some background and to assert that we should remove it from the public API whenever possible.

Do you think its a good idea to use the approach I've proposed in your patch?

collinsauve:
Yes I think it is a good idea to use your approach instead of mine - it is more straightforward.

jorgebay:
I've just submitted a pull request containing a test for the message generated, making sure it contains the timestamp by default #327.

Can you adapt your patch to use the approach suggested so I can merge it and rebase #327 against it.

jorgebay:
Sorry to put pressure on you but we need to close this one by tomorrow, do you think you can adapt your patch?
If you are on your summer vacations or anything like that, nevermind, we can create the patch, we just wanted to acknowledge that the original fix was coming from you :)

collinsauve:
Sorry @jorgebay I was on vacation last week.  Just got back today and catching up and probably won't be able to get to this today.  I can try to do later this week if today was not a hard deadline.

jorgebay:
No problem, I understand. I'll create a patch for this and include the original author in the commit message.

Thanks for identifying the issue and for the patch!

jorgebay:
Closed in favor of #328 .

PR #326, opened 2017-07-05
LocalDate and LocalTime Parse() implementation

jorgebay:


src/Cassandra/LocalTime.cs
```diff
@@ -100,6 +97,32 @@ public LocalTime(int hour, int minute, int second, int nanosecond)
                 nanosecond;
         }
 
+        /// <summary>
+        /// Creates a new <see cref="LocalDate"/> instance based on the string representation.
```
landim: typo: should be LocalTime

src/Cassandra/LocalTime.cs
```diff
@@ -100,6 +97,32 @@ public LocalTime(int hour, int minute, int second, int nanosecond)
...
+        public static LocalTime Parse(string value)
+        {
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+            var parts = value.Split(':');
+            if (parts.Length < 2 || parts.Length > 3)
+            {
+                throw new FormatException("LocalDate format is invalid");
```
landim: same: LocalDate instead of LocalTime

src/Cassandra/LocalTime.cs
jorgebay: Nice catch!

jorgebay:
Fixed xmldoc and exception typos, I'll merge it as soon as unit tests on CI builds finish.

PR #329, opened 2017-07-14
- Allow linq non-redundant boolean expression.

francischie:
Current driver only allows redundant boolean expression. 
Ex:  Where(a => a.Visible == true) . Resharper will always say this is redundant code. Yeah,  I know I can add this to ignore list but I rather not to. 

This changes will allow non-redundant:
Ex:   Where(a => a.Visible)  and will generate CQL  WHERE visible = true instead of throwing an error. 

Note: Negate still not handle and still won't work 
Ex: Where (a => !a.Visible)  which is fine cause this can be rewritten with a.Visible == false and it's also non-redundant.

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUnitTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUnitTests.cs
@@ -702,6 +702,8 @@ APPLY BATCH".Replace("\r", ""));
             Assert.AreEqual(@"SELECT ""x_f1"" FROM ""x_t"" ALLOW FILTERING", query.ToString());
         }
 
+
+
         [Test]
         public void Select_OrderBy_Columns()
         {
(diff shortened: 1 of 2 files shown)


landim:
Hi @francischie , thanks for your contribution. Your PR was not forgotten, I will start a review as soon as possible.

src/Cassandra.Tests/Mapping/Linq/LinqToCqlUnitTests.cs
```diff
@@ -702,6 +702,8 @@ public void Select_One_Columns()
             Assert.AreEqual(@"SELECT ""x_f1"" FROM ""x_t"" ALLOW FILTERING", query.ToString());
         }
 
+
+
```
landim: Avoid excess of empty new lines

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -588,7 +591,7 @@ private Expression AddProjection(Expression node, PocoColumn column = null)
             object value;
             if (node is MemberExpression)
             {
-                value = GetClosureValue((MemberExpression) node);
+                value = GetClosureValue((MemberExpression)node);
```
landim: Avoid changing whitespaces format

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -989,6 +999,10 @@ private Expression FillCondition(MemberExpression node, StringBuilder clause, Li
                         "Trying to order by a field or property that is ignored or not part of the mapping definition.");
                 }
                 clause.Append(Escape(columnName));
+
+                //clause.Append()
```
landim: Delete commented code

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -635,34 +638,34 @@ private bool EvaluateMethod(string name, MethodCallExpression node)
...
-                        .Append(placeHolders)
-                        .Append(")");
-                    return true;
-                }
+                        clause
+                            .Append(" IN (")
+                            .Append(placeHolders)
+                            .Append(")");
+                        return true;
+                    }
```
landim: Avoid code format changes.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
francischie: Oops. I forgot to remove it. :)

jorgebay:
We've merged #351 because the issues from the review were not addressed on time. Anyway, thanks for your contribution @francischie !

PR #330, opened 2017-07-24
CSHARP-370 Add Cluster.ConnectAsync() and Cluster.ShutdownAsync to the API

collinsauve:
Continuing conversations from [mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!topic/csharp-driver-user/W1GUUqVIzUg), [CSHARP-369](https://datastax-oss.atlassian.net/browse/CSHARP-369), and [CSHARP-370](https://datastax-oss.atlassian.net/browse/CSHARP-370).  Note that despite [CSHARP-579](https://datastax-oss.atlassian.net/browse/CSHARP-579) I have some evidence  indicating this problem is actually worse in v3.3.2 compared to v3.2.1 - as soon as we upgraded our QA cluster started producing `TimeoutException`s on Connect where it did not previously (may be coincidence)

Anyways,  here's a fix that includes a `ConnectAsync` and `ShutdownAsync` that seems to work within an `AsyncContext.Run`.  

Two things:

* Needs some integration tests.  Should I just duplicate some existing tests somewhere?
* I've included a copy of  `AsyncLock` from https://github.com/neosmart/AsyncLock.  This is under MIT license.  Is that compatible with this project?  I have included the MIT copyright notice at the top of that file.

### DIFF
--- a/src/Cassandra/Cluster.cs
+++ b/src/Cassandra/Cluster.cs
@@ -39,7 +39,8 @@ namespace Cassandra
         private readonly ControlConnection _controlConnection;
         private volatile bool _initialized;
         private volatile Exception _initException;
-        private readonly object _initLock = new object();
+        private readonly SemaphoreSlim _initLock = new SemaphoreSlim(1, 1);
+
         private readonly Metadata _metadata;
         private readonly Serializer _serializer;
 
@@ -119,7 +120,7 @@ namespace Cassandra
         {
             get
             {
-                Init();
+                TaskHelper.WaitToComplete(Init());
                 return _metadata;
             }
         }
@@ -146,14 +147,15 @@ namespace Cassandra
         /// <summary>
         /// Initializes once (Thread-safe) the control connection and metadata associated with the Cluster instance
         /// </summary>
-        private void Init()
+        private async Task Init()
         {
             if (_initialized)
             {
                 //It was already initialized
                 return;
             }
-            lock (_initLock)
+            await _initLock.WaitAsync().ConfigureAwait(false);
+            try
             {
                 if (_initialized)
                 {
@@ -171,7 +173,8 @@ namespace Cassandra
                     var initialAbortTimeout = Configuration.SocketOptions.ConnectTimeoutMillis * 2 *
                                               _metadata.Hosts.Count;
                     initialAbortTimeout = Math.Max(initialAbortTimeout, ControlConnection.MetadataAbortTimeout);
-                    TaskHelper.WaitToComplete(_controlConnection.Init(), initialAbortTimeout);
+                    await _controlConnection.Init().WaitToCompleteAsync(initialAbortTimeout).ConfigureAwait(false);
+
                     // Initialize policies
                     Configuration.Policies.LoadBalancingPolicy.Initialize(this);
                     Configuration.Policies.SpeculativeExecutionPolicy.Initialize(this);
@@ -203,6 +206,10 @@ namespace Cassandra
                 _metadata.Hosts.Added += OnHostAdded;
                 _metadata.Hosts.Removed += OnHostRemoved;
             }
+            finally
+            {
+                _initLock.Release();
+            }
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


landim:
Hi @collinsauve ,
first of all, thanks for the contribution. I will talk with legal about the copyright. Also, I will start the review of this PR as soon as possible.
About the integration tests, I did not understood your question? What do you mean by duplicate existing?

collinsauve:
Re: Integration tests - I'm thinking there should probably be some tests for these new `ConnectAsync` and `ShutdownAsync` functions.   Are there tests already for `Connect` and `Shutdown` that I can duplicate for the async versions?

landim:
Hi @collinsauve you can take a look at src/Cassandra.IntegrationTests/Core/ConnectionTests.cs. Although, basically all tests use Connect method on TestSetup.

collinsauve:
HI @landim I explorered the tests more and found `ClusterTests`.  I've parameterized those tests to also attempt connect/shutdown using the new `ConnectAsync` and `ShutdownAsync` functions.  Let me know if this is wrong/I should add more tests.

P.S. I will be on vacation next week so will not be very responsive, if at all. :)

jorgebay:
Enjoy your vacations! When you come back, we can discuss if we can avoid using `AsyncLock` in favor of `SemaphoreSlim.WaitAsync()`.
I think its possible to use just `await semaphore.WaitAsync()` and then `semaphore.Release()` wrapped in a `try...finally` block, wdyt?

collinsauve:
I believe the benefit of this `AsyncLock` implementation is that it is re-entrant safe.  So long as we do not need that, `SemaphoreSlim` should work just fine.

src/Cassandra/Cluster.cs
```diff
@@ -39,7 +39,7 @@ public class Cluster : ICluster
         private readonly ControlConnection _controlConnection;
         private volatile bool _initialized;
         private volatile Exception _initException;
-        private readonly object _initLock = new object();
+        private readonly AsyncLock _initLock = new AsyncLock();
```
jorgebay: Change it for a `SemaphoreSlim` instance.

collinsauve:
I've changed it to a `SemaphoreSlim`, and squashed my commits.

jorgebay:
I saw you made the change @collinsauve, I'll look at it in the next few days, sorry I can not look at it earlier... :)

jorgebay:
Somehow related: I've added [CSHARP-598](https://datastax-oss.atlassian.net/browse/CSHARP-598) to simplify `PreserveStackTrace()` implementation.

src/Cassandra/Cluster.cs
```diff
@@ -39,7 +39,8 @@ public class Cluster : ICluster
         private readonly ControlConnection _controlConnection;
         private volatile bool _initialized;
         private volatile Exception _initException;
-        private readonly object _initLock = new object();
+        private readonly SemaphoreSlim _initLock = new SemaphoreSlim(0, 1);
```
jorgebay: It should be `new SemaphoreSlim(1, 1)` as `initialCount` refers to initially available (its confusing setting!)

collinsauve: Thanks @jorgebay, fixed.

collinsauve:
Hi @jorgebay.  Just circling back to this in case it got lost/forgotten about.   I believe I've made all the changes you requested.  Is there anything else I can do to help move this along?

collinsauve:
Btw it appears as though your appveyor is broken, multiple PRs (including this one) are failing with the same errors.  See the last 3 builds in [appveyor history](https://ci.appveyor.com/project/DataStax/csharp-driver/history).

jorgebay:
Sorry for the long delay, thanks for bringing this up, I think I will be able to give a final look at this patch this week.

jorgebay:
I've done final review, focusing on finding possible deadlocks and lgtm.

Regarding the broken builds, I think its due to the several changes in the build tools we made in the past months, I've created a merge commit on a separate branch in order to trigger CI builds (appveyor and private Jenkins): https://ci.appveyor.com/project/DataStax/csharp-driver/build/connect-async-merge-1336

I've created 2 tickets to follow up on this API change: CSHARP-607 (DSE Driver) and CSHARP-608 (for `ICluster` interface, which is a breaking change).

Regarding tests, we should add some more tests for `ConnectAsync()` with paused nodes and stuff like that, cc @landim 

I'll wait for the CI builds to finish and merge it. Thanks again @collinsauve for a great contribution and for your patience!

PR #331, opened 2017-08-07
Change Access Level Modifier of Clear() method from MappingConfigurat…

carloscosta08:
The main purpose of this pull request is to modify the access level of Clear() method from MappingConfiguration.cs.

It will be of great value to have this method exposed, allowing to clear the mappings defined for an instance.


jorgebay:
Sorry for the delay in the answer, we haven't forgot about your contribution.
I'll look at it in the next few days.

PR #334, opened 2017-08-29
Correcting .NET Standard support in README.md

rwkarg:
.NET Standard 1.5 target pulled from https://github.com/datastax/csharp-driver/blob/master/src/Cassandra/Cassandra.csproj#L10


jorgebay:
We support .NET Core platform through .NET Standard 1.5, as you intended to describe but we only support .NET Framework and .NET Core platforms, not [any other platforms that uses the .NET Standard spec](https://docs.microsoft.com/en-us/dotnet/standard/net-standard). For example: we don't officially support Mono 4.6, even though Mono 4.6 adheres to .NET Standard 1.5 spec.

Thanks for your contribution @rwkarg, in this case we are leaving the current message specifying the platforms.

PR #340, opened 2017-09-29
Use ExceptionDispatchInfo to preserve stack trace

jorgebay:

PR #341, opened 2017-09-29
Change the message for batch log write failures

jorgebay:
Use a differentmessage for batch log write failures.
Also fixed WriteFailureException message.

### DIFF
--- a/src/Cassandra/Exceptions/NoHostAvailableException.cs
+++ b/src/Cassandra/Exceptions/NoHostAvailableException.cs
@@ -17,7 +17,9 @@ using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.Net;
+#if !NETCORE
 using System.Runtime.Serialization;
+#endif
 using System.Text;
 
 namespace Cassandra
--- a/src/Cassandra/Exceptions/NoHostAvailableException.cs
+++ b/src/Cassandra/Exceptions/NoHostAvailableException.cs
@@ -18,6 +18,7 @@ using System.Linq;
 using System.Collections.Generic;
 using System.Net;
 using System.Runtime.Serialization;
+using System.Text;
 
 namespace Cassandra
 {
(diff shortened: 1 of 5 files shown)


jorgebay:
Fixes both CSHARP-590 and CSHARP-528.

src/Cassandra/Exceptions/NoHostAvailableException.cs
```diff
@@ -18,6 +18,7 @@
 using System.Collections.Generic;
 using System.Net;
 using System.Runtime.Serialization;
```
landim: Remove unused import

src/Cassandra/Exceptions/NoHostAvailableException.cs
```diff
@@ -18,6 +18,7 @@
...
+using System.Text;
 
 namespace Cassandra
```
landim: Wdyt about change the namespace to Cassandra.Exceptions (to match the folder structure)? I don't have a strong options about it, but I think it could be clearer to user when using code completion

landim:
I will refactor the test once rebased after https://github.com/datastax/csharp-driver/pull/340 is merged

src/Cassandra/Exceptions/NoHostAvailableException.cs
jorgebay: Will do, thanks!

src/Cassandra/Exceptions/NoHostAvailableException.cs
jorgebay: It would be nice to have the exceptions on the correct namespace, unfortunately the exceptions were originally exposed on this namespace and changing it would be a breaking change...

src/Cassandra/Exceptions/NoHostAvailableException.cs
jorgebay: ahh, I've just checked: `System.Runtime.Serialization` is unused for .NET Core target but needed for building against .NET Framework (see `SerializableAttribute` below): I'll surround it with `#if !NETCORE` directive to make it more clear.

jorgebay:
I've just pushed a commit to address the issue, let me know if it looks good to you (we can create additional tests if needed on a different branch).

src/Cassandra/Exceptions/NoHostAvailableException.cs
landim: Good solve!

src/Cassandra/Exceptions/NoHostAvailableException.cs
landim: Agree, it would have to be a long term refactoring (create inherited classes on "right namespace" and mark the current ones as deprecated, etc...)
👍 on keeping the current namespace.

PR #342, opened 2017-09-29
Provide a way to build Pooling options based on protocol version

jorgebay:


src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -214,5 +214,40 @@ public void Connect_With_Ssl_Test()
...
+                TestHelper.WaitUntil(() =>
+                    pool.OpenConnections == coreConnectionLength);
+                var ports = sCluster.GetConnectedPorts();
+                //coreConnectionLength + 1 (the control connection) 
+                Assert.AreEqual(coreConnectionLength + 1, ports.Count);
+                
+                TestHelper.WaitUntil(() =>
+                    {
+                        //Force pool to resize connections to max
+                        pool.ConsiderResizingPool(int.MaxValue);
```
jorgebay: We should avoid using internal methods in integration tests.
Using core connections to test that the settings are being honored is enough for me.

src/Cassandra.Tests/BuilderTests.cs
```diff
@@ -184,23 +184,18 @@ public void WithMaxProtocolVersion_Validates_Greater_Than_Zero()
         }
 
         [Test]
-        public void PoolingOptions_Create_Based_On_Protocol_Version()
+        [TestCase(ProtocolVersion.MaxSupported, 1, 2)]
+        [TestCase(ProtocolVersion.V2, 2, 8)]
+        public void PoolingOptions_Create_Based_On_Protocol_Version(ProtocolVersion protocolVersion,
```
jorgebay: +1 for the refactoring.

landim:
Removed the max connection check on integration (avoiding to have to call internal method)

jorgebay:
This needs to be rebased in order to work, I'll do it.

jorgebay:
Test fixes look good to me.
I've rebased it and squashed @landim  commits into 1.

PR #344, opened 2017-10-05
Connect async tests (CSHARP-370)

landim:
Refactored some tests to run using ConnectAsync method.

### DIFF
--- a/src/Cassandra.IntegrationTests/TestBase/TestGlobals.cs
+++ b/src/Cassandra.IntegrationTests/TestBase/TestGlobals.cs
@@ -196,7 +196,8 @@ namespace Cassandra.IntegrationTests.TestBase
                 }
                 finally
                 {
-                    cluster?.ShutdownAsync().Wait();
+                    var shutdownAsync = cluster?.ShutdownAsync();
+                    if (shutdownAsync != null) await shutdownAsync;
                 }
             }
             else
--- a/src/Cassandra.IntegrationTests/TestBase/TestGlobals.cs
+++ b/src/Cassandra.IntegrationTests/TestBase/TestGlobals.cs
@@ -1,5 +1,5 @@
 ﻿//
-//      Copyright (C) 2012-2014 DataStax Inc.
+//      Copyright (C) 2012-2017 DataStax Inc.
 //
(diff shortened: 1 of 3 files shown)


landim:
By running the MemoryLeak tests using ConectAsync and throwing NoHostAvailableExcetions it failed

>   Increased the amount of new objects
>  Expected: less than or equal to 1
>  But was:  2999

jorgebay:
This is failing due to simulacron integration: CSHARP-609.

src/Cassandra.IntegrationTests/Core/MemoryLeakTests.cs
```diff
@@ -47,8 +50,14 @@ public void Monitor_Should_Not_Leak_Connections_Test()
 
                 for (var i = 0; i < length; i++)
                 {
-                    var ex = Assert.Throws<NoHostAvailableException>(() => cluster.Connect());
-                    Assert.AreEqual(1, ex.Errors.Count);
+                    try
+                    {
+                        TestGlobals.Connect(cluster, asyncConnect, session => {}).ConfigureAwait(false);
```
jorgebay: Forgot to `await` on the returned task...

src/Cassandra.IntegrationTests/TestBase/TestGlobals.cs
```diff
@@ -184,6 +185,29 @@ public static bool VersionMatch(TestCassandraVersion versionAttr, Version execut
...
+            if (asyncConnect)
+            {
+                try
+                {
+                    var session = await cluster.ConnectAsync();
+                    action(session);
+                }
+                finally
+                {
+                    cluster?.ShutdownAsync().Wait();
```
jorgebay: I don't think its a good idea to reuse this method for the memory tests.
If awaiting for `ConnectAsync()`, we should `await` for `ShutdownAsync()` (instead of wait).

landim: I am removing this dotMemoryUnit test as it is not possible to run this test async (dotMemory loose profiling).

landim:
Addressed the usage of `await` instead of `Wait()` and some code fixes at TestGlobals: unused code, comments, etc.

PR #345, opened 2017-10-10
Support prepare on all nodes and re-prepare on UP

jorgebay:
- Introduce `PrepareHandler`.
- Simplify `RequestHandler`, which now only handles `QUERY`, `EXECUTE` and `BATCH` requests.
- Test query preparation using simulacron.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/CustomPayloadTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/CustomPayloadTests.cs
@@ -76,8 +76,12 @@ namespace Cassandra.IntegrationTests.Core
         [Test, TestCassandraVersion(2, 2)]
         public void Prepare_Payload_Test()
         {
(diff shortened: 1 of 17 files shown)


jorgebay:
There is a failure related to timeouts using simulacron, I'm not sure the underlying reason (does `delay_in_ms` works on simulacron?)...

landim:
It should work, I will make some tests

landim:
It should work, I'll take a look at simulacron code

landim:
Rebased with master. Your test looks good. The issue is on simulacron side (as you suspected. Prepare command is handled in a special way and it is ignoring the delayInMs. I already created a bug: https://github.com/datastax/simulacron/issues/23 and also pushed the fix. I will keep an eye on it and update the simulacron version on our CI as soon as it is released.

jorgebay:
I've rebased it against master and add ignore attribute for test related to datastax/simulacron#23.

landim:
looks good to me. 👍

PR #347, opened 2017-10-11
Simulacron startup (Csharp 609)

landim:


jorgebay:
I've changed the title to reference CSHARP-609 (the reference to CSHARP-608 was unintended, right?).

About the ticket on JIRA, @landim can you reopen it again until merged?

src/Cassandra.IntegrationTests/TestClusterManagement/SimulacronManager.cs
```diff
@@ -30,7 +30,7 @@ public static Uri BaseAddress
 
         public void Start()
         {
-            if (_started)
+            if (IsUp())
```
jorgebay: I think we should use a different flag here, like `_init` or whatever, checking that it's started just once.
If the process has exited, it will fail on the next `POST` or `GET` request, which is the desired behaviour.

src/Cassandra.IntegrationTests/TestClusterManagement/SimulacronManager.cs
```diff
@@ -104,7 +104,7 @@ public void Stop()
 
         public bool IsUp()
         {
-            return _started;
+            return _started && _simulacronProcess != null && !_simulacronProcess.HasExited;
```
jorgebay: I think using the `_started` flag as it was, it was easier to understand (started or manually stopped and can be restarted).

Maybe it would be best to remove this method (not used outside this class) and add it if needed in the future when we define a purpose for it.

landim:
thanks for all inputs. I all work on them and push asap.

src/Cassandra.IntegrationTests/TestClusterManagement/SimulacronManager.cs
```diff
@@ -9,31 +9,24 @@ public class SimulacronManager
...
-        public static Uri BaseAddress
-        {
-            get { return new Uri("http://127.0.0.1:8188"); }
-        }
+        public static Uri BaseAddress => new Uri("http://127.0.0.1:8188");
 
         public void Start()
         {
-            if (_started)
+            lock (this)
```
jorgebay: I think this is not necessary as we are not supporting calling `Start()` from multiple threads (otherwise we would have to use some kind of concurrency control to await until it was started).

landim: The integration tests are not run in parallel yet, but as it is something in radar, so I thought it would not hurt to put it already.

PR #348, opened 2017-10-16
Linq: Substract assign for map keys

jorgebay:


src/Cassandra/Data/Linq/CqlOperator.cs
```diff
@@ -25,11 +25,38 @@ public sealed class CqlOperator
...
         /// </summary>
         public static T SubstractAssign<T>(T value) where T: IEnumerable
         {
             return default(T);
         }
+
+        /// <summary>
+        /// Represents the CQL operator to remove an item from a map (col1 = col1 - ?).
+        /// </summary>
+        public static Dictionary<TKey, TValue> SubstractAssign<TKey, TValue>(this Dictionary<TKey, TValue> map,
```
landim: Using SubstractAssign (whith S between B ant T) is not a typo, but ExpressionType use SubtractAssign, which can be confusing. 
We have to keep retro compatibility with released versions, so I would not consider an must do, but rather a 'would be nice to do' (override the method and marking depreacted).

jorgebay: Nice catch, I was unable to see it because of spanish is my native language... https://english.stackexchange.com/questions/3640/is-substract-versus-subtract-a-proper-word

jorgebay:
Rebased to resolve conflicts with master and squashed commits.

jorgebay:
There was a failure on the Jenkins build (no link on this pull request / integration looks disabled), as subtract assign for maps is available only on C* 2.1+.

src/Cassandra/Data/Linq/CqlOperator.cs
landim: I know what you mean, mine is portuguese. I only found because of code completion (was typing subt.. and wasn't finding the method).

landim:
Thanks for that Cassandra version check.

PR #349, opened 2017-10-16
Linq: Support TTL and Timestamp methods on CqlConditionalCommand

jorgebay:


jorgebay:
> I also notive that _timestamp and _ttl (on Linq/CqlCommand) are protected and, according the name convention, should be renamed to Timestamp and TTL. Wdyt?

As those are `protected`, those are part of the public API :/

PR #351, opened 2017-10-17
Linq: Support boolean expressions without equality operators

jorgebay:


francischie:
Is this same?  https://github.com/datastax/csharp-driver/pull/329

jorgebay:
Yes, it's a fix for the same issue , which is [CSHARP-512](https://datastax-oss.atlassian.net/browse/CSHARP-512). On #329, the issues that @landim identified were never addressed, I've thought it was stale.

CSHARP-512 is scheduled for the next release, so I've created an alternate fix.

If you still want #329 to be the one merged, you can address the feedback and add support for filter for boolean members in any of the following forms: `x => x.BooleanValue == falseValue`, `x => !x.BooleanValue`, `x => x.BooleanValue`, `x => x.BooleanValue == trueValue`.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -768,6 +768,14 @@ protected override Expression VisitUnary(UnaryExpression node)
             {
                 var clause = _currentCondition.Get().Item1;
                 var parameters = _currentCondition.Get().Item2;
+                if (node.NodeType == ExpressionType.Not && node.Operand.NodeType == ExpressionType.MemberAccess)
```
landim: Good catch. The NOT cases are covered as well.

src/Cassandra/Data/Linq/CqlExpressionVisitor.cs
```diff
@@ -867,19 +892,33 @@ protected override Expression VisitBinary(BinaryExpression node)
...
                         return node;
                     }
                 }
                 else if (!CqlUnsupTags.Contains(node.NodeType))
                 {
                     var val = Expression.Lambda(node).Compile().DynamicInvoke();
-                    _currentCondition.Get().Item2.Add(val);
-                    _currentCondition.Get().Item1.Append("?");
+                    parameters.Add(val);
+                    clause.Append("?");
```
landim: There are some occurrences of the string hardcoded "?", and as it carry a strong meaning (parameter value) I would refactor to create a constant.

jorgebay:
What do you mean by:

> I would include another test in which the boolean variable

Do you plan to do that later on integration tests or do you want me to do it?

landim:
> Do you plan to do that later on integration tests or do you want me to do it?
Actually it is a suggestion/debate. I was thinking about having the different scope variables of Linq at unit tests.

PR #352, opened 2017-10-18
Linq: Fix CqlQuery and CqlCommand QueryTrace

jorgebay:


src/Cassandra/Data/Linq/CqlCommand.cs
```diff
@@ -59,7 +64,19 @@ public override object[] QueryValues
...
+        /// After being executed, it retrieves the trace of the CQL query.
+        /// <para>Use <see cref="IStatement.EnableTracing"/> to enable tracing.</para>
+        /// <para>
+        /// Note that enabling query trace introduces server-side overhead by storing request information, so it's
+        /// recommended that you only enable query tracing when trying to identify possible issues / debugging. 
+        /// </para>
+        /// </summary>
+        public QueryTrace QueryTrace
+        {
+            get => Volatile.Read(ref _queryTrace);
```
landim: If the purpose is to avoid memory cache of this property, why not using volatile keyword on property?

jorgebay: Its not possible to mark a property as `volatile`. A field can be marked as `volatile` but to mark that the access to that value should be volatile, so the new methods `Volatile.Read()` and `Volatile.Write()` have the correct meaning (a certain address is not volatile, its how you access to it).

landim: Sorry about the confusion: I typo prop/field. 
I was just suggesting, this:
`
volatile _queryTrace;
public QueryTrace QueryTrace
{ 
    get {
        return _queryTrace;
    }
    set{
     _queryTrace = value;
    }
}
`

jorgebay: ok, np. Regarding `volatile` fields, read and writes should to be marked as volatile, instead of the field (address). So `Volatile.Read()` and `Volatile.Write()` should be recommended instead of volatile field declarations.

PR #353, opened 2017-10-20
Mapper: Support nested types with conversion

jorgebay:
Support serialization and deserialization of collections with conversion

### DIFF
--- a/src/Cassandra.Tests/Mapping/Pocos/HairColor.cs
+++ b/src/Cassandra.Tests/Mapping/Pocos/HairColor.cs
@@ -2,12 +2,12 @@
 {
     public enum HairColor
     {
-        Black,
-        Blonde,
-        Brown,
-        Gray,
-        Red,
-        White,
-        Other
+        Black = 0,
+        Blonde = 1,
+        Brown = 2,
+        Gray = 3,
+        Red = 4,
+        White = 5,
+        Other = 6
     }
 }
--- a/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
+++ b/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
@@ -54,11 +54,11 @@ namespace Cassandra.Mapping.TypeConversion
         private static readonly MethodInfo ConvertToArrayFromDbMethod = typeof(TypeConverter).GetTypeInfo()
             .GetMethod(nameof(ConvertToArrayFromDb), PrivateInstance);
 
-        private static readonly MethodInfo ConvertToIEnumerableToDbTypeMethod = typeof(TypeConverter)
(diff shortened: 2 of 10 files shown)


src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
```diff
@@ -302,22 +332,91 @@ internal Delegate GetToDbConverter(Type pocoType, Type dbType)
...
+                        $"Specified cast is not valid: from {a.GetType()} to {typeof(TDatabase)}", ex);
+                }
+            }
+
+            return ChangeType;
         }
 
-        private Dictionary<TKeyResult, TValueResult> ConvertToDictionaryWithCast<TKeySource, TValueSource, TKeyResult, TValueResult>
-            (IDictionary<TKeySource, TValueSource> mapFromDatabase)
+        private IEnumerable<TResult> ConvertToIEnumerableToDbType<TSource, TResult>(IEnumerable<TSource> items)
```
landim: I think the name Convert**From**IEnumerableToDbType would be more clear

src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
```diff
@@ -302,22 +332,91 @@ internal Delegate GetToDbConverter(Type pocoType, Type dbType)
...
+        {
+            return items?.Select(TryFindToDbConverter<TSource, TResult>());
+        }
+
+        private IDictionary<TResultKey, TResultValue> ConvertToIDictionaryToDbType<TSourceKey, TSourceValue, TResultKey,
```
landim: Same From/To rule would be more clear

src/Cassandra.Tests/Mapping/Pocos/HairColor.cs
```diff
@@ -2,12 +2,12 @@
...
     public enum HairColor
     {
-        Black,
-        Blonde,
-        Brown,
-        Gray,
-        Red,
-        White,
-        Other
+        Black = 0,
```
landim: Maybe I am overseeing something, but why set the default value?

jorgebay: I've added some tests that check the value of the enum based on their int representation. Its just to prevent issues in the future when adding a new member in the enum.

src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
jorgebay: Nice catch!

jorgebay:
Fixed the issue around method names:, ie:`ConvertIEnumerableToDbType`.

PR #354, opened 2017-10-23
Fix: Support ColumnAttribute with no name

jorgebay:


landim:
The build failure on jenkins is not related to this PR, I will re-run it. as soon as the appveyor build finishes I am +1 on merge

PR #357, opened 2017-10-25
Fix: Include generated timestamp in batch request

jorgebay:


src/Cassandra/Requests/BatchRequest.cs
```diff
@@ -65,13 +65,26 @@ public BatchRequest(ProtocolVersion protocolVersion, BatchStatement statement, C
                 _batchFlags |= QueryProtocolOptions.QueryFlags.WithSerialConsistency;
                 _serialConsistency = statement.SerialConsistencyLevel;
             }
-            if (_timestamp != null)
+            if (protocolVersion.SupportsTimestamp())
```
landim: An alternative logic to avoid nested ifs is:
``` 
bool timestampSpecified;
timestampSpecified = statement.Timestamp != null;
if (!protocolVersion.SupportsTimestamp())
{
    if (timestampSpecified)
    {
        throw new NotSupportedException("Timestamp for BATCH request is supported in Cassandra 2.1 or above.");
    }
    return;
}
if (timestampSpecified)
{
    _batchFlags |= QueryProtocolOptions.QueryFlags.WithDefaultTimestamp;
    _timestamp = TypeSerializer.SinceUnixEpoch(statement.Timestamp.Value).Ticks / 10;
}
else
{
    var timestamp = policies.TimestampGenerator.Next();
    if (timestamp != long.MinValue)
    {
        _batchFlags |= QueryProtocolOptions.QueryFlags.WithDefaultTimestamp;
        _timestamp = timestamp;
    }
}
```

jorgebay: Nice improvement, will do!

jorgebay: hm... now that I think about using `return` statement in a constructor can be unexpected. For example, functionality added below in the future will never be executed.

I'll move timestamp logic to a different method which should be easier to maintain.

landim:
+1 on merge. the issue on appveyor is related to ccm.

PR #358, opened 2017-10-25
Use default serializer for mockable PreparedStatement instances

jorgebay:
Fixes CSHARP-547.


jorgebay:
The default constructor is present only to allow mocking.
We should introduce an `IPreparedStatement` interface and deprecate this constructor ✅ .

PR #364, opened 2017-11-02
Test for details on authentication error message

landim:
Included a verification on error message when authentication fails:

1. Wrong authentication
2. No authentication specified


jorgebay:
Lgtm!

PR #368, opened 2017-11-06
Remove UpdateIfNotExists() Linq extension method

jorgebay:
Remove method introduced in #350 that is not supported in CQL.


landim:
+1 to merge

PR #374, opened 2017-11-09
Test Session.GetState when cluster disposes

landim:


jorgebay:
+1 on the coverage, lgtm!

PR #381, opened 2017-12-20
CSHARP-640 - Allow Nullable<T> values in UDTs

collinsauve:
For https://datastax-oss.atlassian.net/browse/CSHARP-640

This still needs some tests around it...  investigating where to put them.

### DIFF
--- a/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
+++ b/src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
@@ -96,6 +96,15 @@ namespace Cassandra.Mapping.TypeConversion
         /// </summary>
         internal object ConvertToDbFromUdtFieldValue(Type valueType, Type dbType, object value)
         {
+            if (valueType.GetTypeInfo().IsGenericType && valueType.GetGenericTypeDefinition() == typeof(Nullable<>))
+            {
+                var nullableType = typeof(Nullable<>).MakeGenericType(dbType);
+                if (valueType == nullableType)
+                {
+                    return value;
+                }
+            }
+
             var converter = GetToDbConverter(valueType, dbType);
             if (converter == null)
             {
--- a/src/Cassandra.IntegrationTests/Core/UdtMappingsTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/UdtMappingsTests.cs
@@ -34,7 +34,7 @@ namespace Cassandra.IntegrationTests.Core
             base.OneTimeSetUp();
 
             const string cqlType1 = "CREATE TYPE phone (alias text, number text, country_code int, verified_at timestamp, phone_type text)";
-            const string cqlType2 = "CREATE TYPE contact (first_name text, last_name text, birth_date timestamp, phones set<frozen<phone>>, emails set<text>)";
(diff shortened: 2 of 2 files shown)


jorgebay:
Nice findings!

About the implementation, what do you think about moving the logic to `ConvertToDbFromUdtFieldValue()`?

collinsauve:
Np, moved it.

collinsauve:
Hey @jorgebay are there any other tests that exercise `Mapper.Insert` with a table using UDTs?   If so can you point me at them so I can use them as an example?

jorgebay:
I think this is not directly related to the mapper (conversion for udts is made by the default `TypeConverter` from the Mapper), so I think the integration tests should go into: [UdtMappingsTests.cs](https://github.com/datastax/csharp-driver/blob/3.4.0/src/Cassandra.IntegrationTests/Core/UdtMappingsTests.cs).

In any case, there are some unit and integration tests for UDTs with the Mapper and Linq: 
- https://github.com/datastax/csharp-driver/blob/3.4.0/src/Cassandra.Tests/Mapping/InsertTests.cs#L134
- https://github.com/datastax/csharp-driver/blob/3.4.0/src/Cassandra.IntegrationTests/Linq/LinqTable/LinqUdtTests.cs
- https://github.com/datastax/csharp-driver/blob/3.4.0/src/Cassandra.IntegrationTests/Mapping/Tests/Fetch.cs#L342

There is a doc on running integration tests locally: https://github.com/datastax/csharp-driver/wiki/Building-and-running-tests
If you don't want to setup the env for running the integration tests locally, feel free to use our CI to check your tests results (just let us know when its ready for review).

collinsauve:
Sorry for the delay, just managed to get some free time today to work on this.

I've changed UdtMappingTests to perform an insert with a `long?` field in the Contact UDT.   If you want a different name for the field let me know, happy to change it (I'm not very creative).

collinsauve:
Test failure is in LoadBalancingPolicyShortTests... probably not related to my changes.  Is it possible this is an intermittent failure @jorgebay  ?

jorgebay:
The test failure is caused by ccm not able to create a test C* cluster, not related to this change 👍

src/Cassandra/Mapping/TypeConversion/TypeConverter.cs
```diff
@@ -96,6 +96,12 @@ protected TypeConverter()
         /// </summary>
         internal object ConvertToDbFromUdtFieldValue(Type valueType, Type dbType, object value)
         {
+            var nullableType = typeof(Nullable<>).MakeGenericType(dbType);
+            if (valueType == nullableType)
+            {
+                return value == null ? null : nullableType.GetTypeInfo().GetProperty("Value").GetValue(value);
```
jorgebay: You can use just `return value` for the nullable type regardless of the underlying value.

collinsauve:
@jorgebay thanks for those suggestions, I've made the changes requested.  Can you review again?

jorgebay:
Thanks for the quick turnaround, I'll look at it today.

jorgebay:
Thanks @collinsauve !
I've merge it, I'll add a more tests to cover using udts with nulls as parameters.

collinsauve:
Thanks @jorgebay.

Do you have a timeline for 3.4.1 ?  Just wondering when to expect this fix (bug is blocking our upgrade to 3.4.x).

jorgebay:
We don't have a target date yet...

PR #382, opened 2018-01-15
 Move response warnings and custom payload parsing to Result

jorgebay:
Fixes CSHARP-643.


jorgebay:
Thanks @landim, fixed it.

src/Cassandra/Responses/Response.cs
```diff
@@ -25,12 +26,22 @@ internal class Response
...
-        protected internal Guid? TraceId { get; set; }
+        protected internal Guid? TraceId { get; }
+
+        /// <summary>
+        /// Warnings from the server which were generated by the server to go along with this response.
+        /// </summary>
+        public string[] Warnings { get; }
+
+        /// <summary>
+        /// Gets or sets the incoming custom payload.
```
mpenick: The comment says, "Gets or sets", but this is no longer publicly settable.

jorgebay: +1 I'll change that

PR #385, opened 2018-02-02
Test: Insert UDT with null properties

jorgebay:
Test for CSHARP-640.

Revamped equality comparison for UDTs on the test.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/UdtMappingsTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/UdtMappingsTests.cs
@@ -20,6 +20,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Cassandra.Tests;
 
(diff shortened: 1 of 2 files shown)


landim:
If you rebase with master (with your mono build fix) it will pass on jenkins build. 
The fail build of appveyor might need a deeper look.

jorgebay:
Regarding the Linq/Mapper failures: we are using a `KeyedCollection<,>` so ordering is guaranteed during the instance lifetime, the problem is that it maintains the order it was added which depends on reflection `GetFields()` method, which doesn't guarantee order.

This behaviour should be expected, its not a driver issue, it generate queries like: `" WHERE Id2 = ? AND Id1 = ?"` or `" WHERE Id1 = ? AND Id2 = ?"` where the parameter order provided is in alignment with the position of the markers in the query.

We have to fix the tests (in a separate PR) to look for either one or the other.

PR #386, opened 2018-02-06
Fix Mapping tests by using regex

landim:
Changing the validation of generated queries using regex instead of string comparison.

### DIFF
--- a/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUpdateUnitTests.cs
+++ b/src/Cassandra.Tests/Mapping/Linq/LinqToCqlUpdateUnitTests.cs
@@ -235,6 +235,9 @@ namespace Cassandra.Tests.Mapping.Linq
                 .Select(t => new AllTypesDecorated { DateTimeValue = dateTimeValue })
                 .UpdateIf(t => t.IntValue == 100)
                 .Execute();
+            TestHelper.VerifyUpdateCqlColumns("atd", query, new []{@"""string_VALUE"""}, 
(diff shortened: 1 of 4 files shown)


landim:
The insert tests have been fixed. I will move to fix update linq tests.

landim:
The `AtomicMonotonicTimestampGenerator` tests are failing. Wdyt about use `Assert.GreaterThan` instead of `AreEquals` as the result can be unpredictable?

jorgebay:
Nevermind the TimestampGenerator tests, we could fix them in a separate PR.

landim:
I've rerun the tests and looks its ready for review!

src/Cassandra.Tests/Mapping/Linq/LinqToCqlUpdateUnitTests.cs
```diff
@@ -69,8 +69,9 @@ public void Update_Multiple_Where_Test()
                 .Select(t => new AllTypesEntity { StringValue = "Billy the Vision" })
                 .Update()
                 .Execute();
-            Assert.AreEqual("UPDATE tbl1 SET string_val = ? WHERE id = ? AND val2 > ?", query);
-            CollectionAssert.AreEqual(new object[] { "Billy the Vision", id, 20M }, parameters);
+            TestHelper.VerifyUpdateCqlColumns(@"tbl1", query, new []{@"string_val"},
```
jorgebay: No need for verbatim string literal, `@"something"`

landim:
Addressed all the suggestions

jorgebay:
that was fast!

jorgebay:
lgtm! +1 :shipit:

PR #387, opened 2018-02-13
Linq: Use a list of conditions

jorgebay:
Introduces `IConditionItem` that represents a [CQL relation](http://cassandra.apache.org/doc/latest/cql/dml.html#grammar-token-relation) in Linq. This patch separates Linq expression parsing and CQL generation, to allow more complex conditions in the future.

Fixes CSHARP-498 by introducing conversions when generating parameters.

Its a large changeset but helps to make `CqlExpressionVisitor` more readable where possible (expression tree evaluation is usually nested, so there's that limitation to consider).


jorgebay:
I've added a small fix to invert the condition of "yoda" expressions and a small fix for the test.

PR #392, opened 2018-02-19
Test ControlConnection failover using Simulacron

jorgebay:
Use simulacron instead of ccm to test failover mechanism of the `ControlConnection`.


src/Cassandra.IntegrationTests/Core/PoolShortTests.cs
```diff
@@ -239,5 +239,115 @@ public void PoolingOptions_Create_Based_On_Protocol(ProtocolVersion protocolVers
...
+                // Drop all connections to hosts
+                foreach (var connection in serverConnections)
+                {
+                    await testCluster.DropConnection(connection);
+                }
+
+                TestHelper.WaitUntil(() => !cluster.GetHost(ccAddress).IsUp);
+
+                // All host should be down by now
+                TestHelper.WaitUntil(() => cluster.AllHosts().All(h => !h.IsUp));
```
landim: As the HeartBeatInterval its been set, wdyt about also Assert the amount of time it took to mark the node up and down?

jorgebay: Its a good idea but I would like to avoid being too specific around time as we the precision is not guaranteed (there are huge difference between runs on AppVeyor specially).

PR #399, opened 2018-05-08
Remove C* 2.0 from build matrix

landim:


jorgebay:
There seems to be an issue with the image used by Jenkins related to mono, but there isn't any mono installation on the image setup scripts (like `apt install mono-devel`) 🤔, should we [add it](http://www.mono-project.com/download/stable/#download-lin)?

Do you know how it used to work? (probably some base image?)

landim:
I am checking if that was missing on last image update, thanks for the heads up.

jorgebay:
+1

PR #400, opened 2018-05-08
Remove C* 2.0 from CI matrix

tolbertam:
Just a small change to remove C* 2.0 from the build matrix as unsupported C* versions will no longer be part of our CI env.


tolbertam:
Oh woops, I see #399 takes care of this, closing :D

PR #403, opened 2018-05-29
StatementFactory: Detect cached faulted Task and attempt to re-prepare + cache key fixes

jorgebay:
Fixes: CSHARP-660 and CSHARP-667.

Check for cached faulted Task and attempt to re-prepare on the `StatementFactory`.
Additionally, use a composed cache key,

### DIFF
--- a/src/Cassandra/Mapping/Statements/StatementFactory.cs
+++ b/src/Cassandra/Mapping/Statements/StatementFactory.cs
@@ -12,7 +12,7 @@ namespace Cassandra.Mapping.Statements
     /// </summary>
     internal class StatementFactory
     {
-        private readonly ConcurrentDictionary<string, Task<PreparedStatement>> _statementCache;
+        private readonly ConcurrentDictionary<CacheKey, Task<PreparedStatement>> _statementCache;
         private static readonly Logger Logger = new Logger(typeof(StatementFactory));
         private int _statementCacheCount;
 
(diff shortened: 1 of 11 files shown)


jorgebay:
cc @landim

src/Cassandra/Mapping/Statements/StatementFactory.cs
```diff
@@ -39,25 +40,49 @@ public StatementFactory()
...
-                               "to misuse of the driver, you should use parameter markers for queries. You can " +
-                               "configure this warning threshold using " +
-                               "MappingConfiguration.SetMaxStatementPreparedThreshold() method.", count);
+                // The exception was caused from awaiting upon a Task that was previously cached
+                // It's possible that the schema or topology changed making this query preparation to succeed
+                // in a new attemp
+                prepareTask = session.PrepareAsync(query);
+                ps = await prepareTask.ConfigureAwait(false);
+                // AddOrUpdate() returns a task which we already waited upon, its safe to call Forget()
+                _statementCache.AddOrUpdate(psCacheKey, prepareTask, (k, v) => prepareTask).Forget();
```
landim: Why not use `TryUpdate` instead of AddOrUpdate? from what I understand, the key is already there always

jorgebay: `TryUpdate()` will only update after a comparison. In this case, I want to update no matter what (last one wins), that is why I use `(k, v) => prepareTask` lambda.

jorgebay:
Thanks for the review @landim

PR #405, opened 2018-06-01
Use OptionsRequest for heartbeats

jorgebay:
Use OPTIONS request for heartbeats and convert few tests to async
execution.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ConnectionTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ConnectionTests.cs
@@ -690,34 +690,36 @@ namespace Cassandra.IntegrationTests.Core
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


jorgebay:
cc @landim

landim:
The errors at CI builds are not related to this PR

jorgebay:
> One side note. I think that simulacron would be beneficial to verify the amount of heartbeat request sent to server. wdyt?

Its a good idea, does simulacron logs it? How can we retrieve it?

landim:
You can get all the requests received by the cluster/dc/node using this method: https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/TestClusterManagement/Simulacron/SimulacronBase.cs#L93

PR #408, opened 2018-06-04
EC2 multi-region address translator

jorgebay:
An implementation for multi-region EC2 deployments, where clients are also deployed in EC2, in order to optimizes network costs, as Amazon charges more for communication over public IPs.


jorgebay:
I'm testing it on EC2 now.

jorgebay:
I've tested on EC2 and reverse+forward resolution in the address translator does the trick, ie:
`52.91.x.x` -> `ip-172-31-20-118.ec2.internal` -> `172.31.20.118`
(public ip -> name -> private ip).

jorgebay:
cc @landim

PR #414, opened 2018-06-27
Support setting consistency level for CqlConditionalCommand

ultimaweapon:
Currently, If we call to `SetSerialConsistencyLevel` or `SetConsistencyLevel` after a call to conditional method (e.g. `IfNotExists`) like the following code:

```cs
var result = await table
    .Insert(new Model { Column = value }, false)
    .IfNotExists()
    .SetSerialConsistencyLevel(ConsistencyLevel.Serial)
    .SetConsistencyLevel(ConsistencyLevel.All)
    .ExecuteAsync();
```

The result that returned from `ExecuteAsync` will not be `AppliedInfo`. Thus, no way to check if conditional operation is succeed or not. This PR will fix this issue.

### DIFF
--- a/src/Cassandra/Data/Linq/CqlConditionalCommand.cs
+++ b/src/Cassandra/Data/Linq/CqlConditionalCommand.cs
@@ -52,6 +52,18 @@ namespace Cassandra.Data.Linq
             return TaskHelper.WaitToComplete(task, config.ClientOptions.QueryAbortTimeout);
         }
 
+        public new CqlConditionalCommand<TEntity> SetConsistencyLevel(ConsistencyLevel? consistencyLevel)
+        {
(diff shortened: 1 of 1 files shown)


ultimaweapon:
Actually we can't do that since the return value of `SetSerialConsistencyLevel` or `SetConsistencyLevel` is `CqlCommand`, not `CqlInsert`. AFAIK, the only class that have `IfNotExists` is `CqlInsert`.

Anyway, I already signed CLA on last Thursday. After looking at #404 it seem like it will take some time to complete at DataStax side.

jorgebay:
hm... the CLA bot should have notified back that is was signed, I'll check it manually. Once I get that confirmation, I'll merge it.

PR #415, opened 2018-07-02
Constructor overloading for the Class LocalDate

sandeepslive:
Will add one more public constructor to handle c# date time object, for convenience to the programmer 
      **(Cassandra\LocalDate.cs)**
public LocalDate(DateTime datetime)
{
this. LocalDate(datetime.year, datetime.month, datetime.day)
}


jorgebay:
Hi @sandeepslive !
It looks like your pull request doesn't contain a change to the `LocalDate`, feel free to submit a new pull request or change your base branch on this one.

Regarding including a new constructor, I think it makes more sense to create a static method `LocalDate.From(DateTime)` and `ToDateTime()`.

PR #416, opened 2018-08-14
Blacklisted DC policy

slapenko:
This policy ensures that any hosts from the provided data centers list will never be used. 
It might be useful if some maintenance should be done in data centers.


jorgebay:
Thank you for your contribution.

You can check out our issue tracker to look for existing features and bugs to contribute or to propose a new feature: https://datastax-oss.atlassian.net/projects/CSHARP/issues

PR #421, opened 2018-11-05
RowSet should support multiple enumerations.

aalmada:
`RowSet` only supports one enumeration because all its `IEnumerator` instances dequeue from the same instance of `ConcurrentQueue`. I know this is by design but [it's not the expected behavior in .NET](https://blog.usejournal.com/enumeration-in-net-d5674921512e). It breaks many of the LINQ operators.

For example, `Any()` is commonly used to check if an enumeration is empty. It calls `IEnumerator.MoveNext()` once (more if used with a filter operation). With the current `RowSet` behavior, it means that the first element will be lost for any consecutive enumerations. This may generate an hard to spot bug if the result contains a lot of elements.

The solution I've found most people use is to add a `ToList()`. This simply defeats all memory management strategies in `RowSet` because all resulting rows will end up in memory, putting a lot of pressure on the GC and risking an `OutOfMemoryException`. It also breaks any possible lazy evaluation strategy.

This PR refactors `RowSet` so that it supports multiple enumerations. I'm not completely happy with the page handling but I hope it hints a possible solution for this issue.

### DIFF
--- a/src/Cassandra.Tests/RowSetUnitTests.cs
+++ b/src/Cassandra.Tests/RowSetUnitTests.cs
@@ -203,7 +203,7 @@ namespace Cassandra.Tests
             //Invoke it in parallel 15 times
             Parallel.Invoke(iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration, iteration);
 
-            //Assert that the fetch was called just 15 time5
+            //Assert that the fetch was called just 15 times
             Assert.AreEqual(15, fetchCounter);
 
             //Sum all rows dequeued from the different threads 
--- a/src/Cassandra.Tests/RowSetUnitTests.cs
+++ b/src/Cassandra.Tests/RowSetUnitTests.cs
@@ -211,7 +211,7 @@ namespace Cassandra.Tests
             //Check that the total amount of rows dequeued are the same as pageSize * number of pages. 
             Assert.AreEqual(pageSize * 2 * 15, totalRows);
         }
-/* 
+
         /// <summary>
         /// Test RowSet fetch next with concurrent calls
         /// </summary>
@@ -245,9 +245,9 @@ namespace Cassandra.Tests
             Assert.AreEqual((pages - 1), fetchCounter);
 
             //Check that the total amount of rows dequeued are the same as pageSize * number of pages. 
-            Assert.AreEqual(pageSize * pages, Volatile.Read(ref counter));
+            Assert.AreEqual(Environment.ProcessorCount * pageSize * pages, Volatile.Read(ref counter));
         }
-*/
+
         [Test]
         public void RowSetFetchNext3Pages()
         {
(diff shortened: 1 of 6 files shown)


jorgebay:
Thanks for your contribution @aalmada. Just as a heads up, I will not be able to look at it before the end of this week.

jorgebay:
I understand that it can be unexpected that `RowSet` dequeues as part of being enumerated.

Some history: in version 2.0 of the driver (following Apache Cassandra 2.0 release) back in 2013 result paging was introduced. As part of the effort the idea was that results would always be paged in an attempt to prevent users shooting themselves in the foot by retrieving huge datasets inadvertently causing OOM.

Apache Cassandra is designed to deal with tons of data and at the same time be user friendly. From the server perspective, when a query like `SELECT * FROM large_table` comes you have 3 options:
- Serve all the results in a response.
- Reject the request (e.g., "Your query matches too many results")
- Return the first page of results and wait for the client to ask more.

The latter is what the C* team when for.

On the client side, it was designed that usability of `ResultSet` (java, python et al) / `RowSet` (C#) should not be affected by this change: the user should have to write the same code to deal with a small result set or with a large one. That way the user will see the same behaviour in development (few results) and in production (more results).

Currently, the only downside is that the user can only iterate the result once (which should not be a showstopper), which is usually a best practice (e.g., Resharper warns about multiple iterations of a single `IEnumerable` instance)

In your example, about using `Any()` you can use:

```c#
if (rs.GetAvailableWithoutFetch() > 0)
{
  // Do something
}
```

If somebody intentionally want to perform multiple iterations of the result they can always use:

```c#
// ToList() or ToArray() will fetch the following pages, so you will have all the results in memory
IList<Row> rows = rs.ToList();
// Iterate through the immutable list
```
In any case, anyway we can find ways to make `RowSet` more user friendly, like exposing our own `Any()` method to take precedence over Linq's extension method.

I'm personally not very happy with the Enumerator causing side effects (retrieving following pages), but making a change like this will have other consequences:
- All the rows in the result will be in the app memory (there can be millions of them)
- The user will have to make follow up queries to retrieve the following pages.

aalmada:
With PR I want to demonstrate that it's possible to have an implementation of RowSet with a internal memory management (i.e. ConcurrentQueue) that allows multiple enumerations. It is just necessary to move all the immutable data into the ```IEnumerable``` definition and keep the mutable (enumeration-related) data in the ```IEnumerator``` definition.

In this PR there are still two issues that are not well resolved. Better solutions for these require changes on other parts of the code that I don't really master.

- ```OutputRows``` adds ```Row```s into ```RowSet``` and sets its initial ```PagingState```. To workaround this, the ```IEnumerable``` keeps this data and passes it on to new ```IEnumerator``` instances. This is an internal behavior but it's still strange. I don't really know what's the role of ```OutputRows``` here.

- ```PagingState``` changes while ```RowSet``` is enumerated so it should be a property of ```RowSetEnumerator```. The rest of the code expect it to be a property of ```RowSet```...

aalmada:
So you avoid the user shooting one foot by letting him shoot the other...

PR #422, opened 2018-11-09
Fix for concurrency bug CSHARP-709

si31:
This is a fix for ticket https://datastax-oss.atlassian.net/browse/CSHARP-709

### DIFF
--- a/src/Cassandra/Tasks/HashedWheelTimer.cs
+++ b/src/Cassandra/Tasks/HashedWheelTimer.cs
@@ -201,7 +201,14 @@ namespace Cassandra.Tasks
             TimeoutItem timeout;
             while (_cancelledTimeouts.TryDequeue(out timeout))
             {
-                timeout.Bucket.Remove(timeout);
+                try
+                {
+                    timeout.Bucket.Remove(timeout);
+                }
+                catch (NullReferenceException)
+                {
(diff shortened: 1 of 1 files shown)


jorgebay:
Nice troubleshooting to come to the conclusion on the ticket.

On the first pass, it looks good. I'll look into it with more detail during this week.

jorgebay:
Sorry @si31, I'll look at it today/tomorrow.

si31:
Ahh yes of course, you're right. It would require an atomic operation.

I would suggest, within the cancelled check in TimerTick, to try to remove it from cancelled timeouts using the TryDequeue operation. Then you shouldn't get the error, but the try catch is probably a good idea anyway.

Thanks a lot for looking at this.

jorgebay:
Remember that `TimerTick()` is executed on an IO thread and we need to get out of `TimerTick()` as as soon as possible to try to have predictable performance.

I think using a try/catch on `RemoveCancelled()` is safer choice than looping over the cancelled timeouts again.

si31:
Cool, it would be great if that you could add that change asap, I've updated this pr with it if that's easier.

Thanks again

jorgebay:
I'm waiting for CI to finish and I'll merge it.

jorgebay:
Thanks @si31 for your contribution and the effort in finding the underlying reason of the bug! 🎉

PR #423, opened 2018-11-19
Fix bug while lowering capital letters in different cultures.

asyilmaz:
ToLower converts the letter "I" to "ı" (i without dot) in Turkish culture settings. 
For example, if a class name is "UserInfo", ToLower converts it to "userınfo" instead of "userinfo".
CultureInfo("en-US") prevents it.

### DIFF
--- a/src/Cassandra/UdtMappingDefinitions.cs
+++ b/src/Cassandra/UdtMappingDefinitions.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Globalization;
 using System.Threading.Tasks;
 using Cassandra.Serialization;
 using Cassandra.Tasks;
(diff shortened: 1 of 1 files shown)


src/Cassandra/UdtMappingDefinitions.cs
```diff
@@ -89,7 +90,8 @@ private async Task<UdtColumnInfo> GetDefinitionAsync(string keyspace, UdtMap map
             if (map.IgnoreCase)
             {
                 //identifiers are lower cased in Cassandra
-                caseSensitiveUdtName = caseSensitiveUdtName.ToLower();
+                CultureInfo textInfo = new CultureInfo("en-US");
+                caseSensitiveUdtName = textInfo.TextInfo.ToLower(caseSensitiveUdtName);
```
jorgebay: Can we use `ToLowerInvariant()`, that uses [`InvariantCulture`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture) instead?

asyilmaz: I tried `ToLowerInvariant()` but it didn't work.

jorgebay: [`ToLowerInvariant()` calls `ToLower(CultureInfo.InvariantCulture)` internally](https://referencesource.microsoft.com/#mscorlib/system/string.cs,2662). It should work, could you try again?

We should use either `ToLowerInvariant()` or `ToLower(CultureInfo.InvariantCulture)` and avoid using a new `CultureInfo("en-US")` instance.

asyilmaz: Sorry, it was my mistake. I tried both of them again and `ToLowerInvariant()` works. `ToLower(CultureInfo.InvariantCulture)` also works.

PR #424, opened 2018-12-06
Provide Serializer without open connection to Cassandra

vladyslav-vasyliev:
It would be nice to be able to create Table instance without open connection to Cassandra for the purpose of development and testing. At the moment it is not possble since Table.Create method calls initialization of Metadata and opens ControlConnection. And there is no way to mock ControlConnection since it's internal.
The reason why ControlConnection is used in Table is that it provides Serializer for CqlGenerator.

https://datastax-oss.atlassian.net/browse/CSHARP-712


joao-r-reis:
Hi @vladyslav-vasyliev can you give an example of a test that you can not create at this time which would be possible after this change?

It makes sense for `Serializer` to stay in the `ControlConnection` because it's this class that manages it. During the initialization of the `ControlConnection` the appropriate `ProtocolVersion` will be chosen taking into account what versions are supported by the server and the client. 

The `Configuration` class is supposed to be a simple configuration object that contains everything that is configured during the `Cluster` building process. The properties of this class should remain as static as possible.

Also, I would appreciate it if the formatting changes are kept on a different commit so that I can more easily review the changes.

joao-r-reis:
Closing due to lack of activity.

PR #425, opened 2018-12-21
CSHARP-715 Fixed some issues with tests and one bug in localtime class

joao-r-reis:
Marked tests that don't run on CI with `Ignore` (these tests need to be refactored/deleted when possible). Also marked some tests with `short` so they run on CI.

Set `Thread.CurrentCulture` to `InvariantCulture` on some tests that depend on it.

Fixed a bug in `LocalTime` class which causes a failure when parsing a localtime string with a culture that doesn't have `.` as the decimal symbol (needs review).

Fixed a test that was impacted by a Simulacron bug that has since been fixed (datastax/simulacron#31). Also removed `Ignore` from this test.


src/Cassandra.IntegrationTests/Core/HeartbeatTests.cs
```diff
@@ -45,8 +45,8 @@ public void OneTimeTearDown()
             _testCluster.Remove().Wait();
         }
 
-        [TestCase(true)]
-        [TestCase(false)]
+        [TestCase(true), Category("short")]
```
jorgebay: We can add the category to the fixture instead of a single test.

src/Cassandra.IntegrationTests/Core/SessionStateTests.cs
```diff
@@ -55,7 +55,7 @@ public void OneTimeTearDown()
             _testCluster.Remove().Wait();
         }
 
-        [Test]
+        [Test, Category("short")]
```
jorgebay: The same here.

src/Cassandra/LocalTime.cs
```diff
@@ -115,7 +116,7 @@ public static LocalTime Parse(string value)
             var nanos = 0;
             if (parts.Length == 3)
             {
-                var decimalSeconds = Convert.ToDecimal(parts[2]);
+                var decimalSeconds = Convert.ToDecimal(parts[2], CultureInfo.InvariantCulture);
```
jorgebay: I think we should leave it as culture dependent, I see no reason for change. We can document it in the method.

src/Cassandra.IntegrationTests/Core/SessionStateTests.cs
joao-r-reis: done

src/Cassandra.IntegrationTests/Core/HeartbeatTests.cs
joao-r-reis: done

src/Cassandra/LocalTime.cs
jorgebay: hm... I see, good catch!

PR #426, opened 2018-12-28
Added Metadata.CheckSchemaAgreement() and ExecutionInfo.IsSchemaInAgreement() methods

joao-r-reis:
Added `Metadata.CheckSchemaAgreement()` and `ExecutionInfo.IsSchemaInAgreement()` methods.

Also added a method to the `Builder` class to configure the max number of seconds that the driver will wait for the schema agreement after a DDL statement is executed.


src/Cassandra/Metadata.cs
```diff
@@ -422,16 +422,84 @@ internal void ClearAggregate(string keyspaceName, string aggregateName, string[]
...
         }
 
+        /// <summary>
+        /// Initiates a schema agreement check. This method does not perform retries so
+        /// <see cref="ProtocolOptions.MaxSchemaAgreementWaitSeconds"/> does not apply. 
+        /// </summary>
+        /// <returns>True if schema agreement was successful and false if it was not successful.</returns>
+        /// <exception cref="OperationCanceledException">If operation was canceled due to
+        /// the provided <see cref="CancellationToken"/></exception>
+        public async Task<bool> CheckSchemaAgreementAsync(CancellationToken ct = default(CancellationToken))
```
jorgebay: I think providing cancellation options here seems like overkill, as the method only executes 2 queries in parallel that should take less than 1ms on a healthy cluster and returns the result. Additionally, we will be introducing the first driver method with a cancellation token that from a user standpoint it might seem arbitrary.

src/Cassandra/Metadata.cs
```diff
@@ -422,16 +422,84 @@ internal void ClearAggregate(string keyspaceName, string aggregateName, string[]
             }
         }
 
+        /// <summary>
+        /// Initiates a schema agreement check. This method does not perform retries so
+        /// <see cref="ProtocolOptions.MaxSchemaAgreementWaitSeconds"/> does not apply.
```
jorgebay: We could include a small paragraph to detail what we mean by "schema agreement", something like:
> Schema changes need to be propagated to all nodes in the cluster. Once they have settled on a common version, we say that they are in agreement.

src/Cassandra/Metadata.cs
```diff
@@ -422,16 +422,84 @@ internal void ClearAggregate(string keyspaceName, string aggregateName, string[]
...
+        /// </param>
+        /// <param name="peerVersionsQuery">
+        /// Results obtained from a query to <code>system.peers</code> table.
+        /// Must contain the <code>schema_version</code> column.
+        /// </param>
+        /// <returns><code>True</code> if there is a schema agreement (only 1 schema version). <code>False</code> otherwise.</returns>
+        private bool CheckSchemaVersionResults(
+            IEnumerable<Row> localVersionQuery, IEnumerable<Row> peerVersionsQuery)
+        {
+            var versions = new HashSet<Guid>
```
jorgebay: NIT: It could be simplified to use linq all the way: 

```C#
return new HashSet<Guid>(
    peerVersionsQuery
    .Concat(localVersionQuery)
    .Select(r => r.GetValue<Guid>("schema_version"))).Count == 1
```

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -158,8 +158,18 @@ private void HandleRowSetResult(Response response)
...
+
+            // This is a schema change so initialize schema in agreement as false
+            result.Info.SetSchemaInAgreement(false);
+
+            // Wait for the schema change before returning the result
+            _parent.SetCompleted(
+                result, 
+                () =>
+                {
+                    var schemaAgreed = _session.Cluster.Metadata.WaitForSchemaAgreement(_connection);
```
jorgebay: We could improve this by using non-blocking code, but this is out of scope of this task. We can create a ticket to improve this by using non-blocking code, wdyt?

src/Cassandra/RowPopulators/ExecutionInfo.cs
```diff
@@ -85,6 +86,20 @@ public IPEndPoint QueriedHost
         /// </summary>
         public ConsistencyLevel AchievedConsistency { get; private set; }
 
+        /// <summary>
+        /// After a successful schema-altering query (ex: creating a table), the driver will check if
+        /// the cluster's nodes agree on the new schema version. If not, it will keep retrying for a given
+        /// delay (configurable via <see cref="Builder.WithMaxSchemaAgreementWaitSeconds"/>).<para/>
+        ///
```
jorgebay: NIT: We can use `<para>` tags to separate paragraphs.

src/Cassandra/ProtocolOptions.cs
```diff
@@ -31,11 +31,16 @@ public class ProtocolOptions
         /// </summary>
         internal const int MaximumFrameLength = 256*1024*1024;
 
+        /// <summary>
+        /// /// The default value for <see cref="get_MaxSchemaAgreementWaitSeconds"/>: 10.
```
jorgebay: Unintended double `///`

src/Cassandra.IntegrationTests/Core/SchemaAgreementTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        private Session _session;
+
+        private const int MaxSchemaAgreementWaitSeconds = 30;
+
+        public override void OneTimeSetUp()
+        {
+            base.OneTimeSetUp();
+            _cluster = Cluster.Builder().AddContactPoint(TestCluster.InitialContactPoint)
+                              .WithSocketOptions(new SocketOptions().SetReadTimeoutMillis(30000))
+                              .WithSocketOptions(new SocketOptions().SetConnectTimeoutMillis(60000))
```
jorgebay: Probably unintended: multiple `SocketOptions` instances.

src/Cassandra.IntegrationTests/Core/SchemaAgreementSimulacronTests.cs
```diff
@@ -0,0 +1,130 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+
+    using Cassandra.IntegrationTests.TestBase;
+    using Cassandra.IntegrationTests.TestClusterManagement.Simulacron;
+
+    using NUnit.Framework;
+
+    [TestFixture, Category("short")]
+    public class SchemaAgreementSimulacronTests
```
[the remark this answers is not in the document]
joao-r-reis: Yeah, I tried to use simulacron to also test the schema agreement waiting that is done by the driver when a `SCHEMA_CHANGED` response is received but I didn't find any way to do it so I had to use CCM for that.

src/Cassandra.Tests/TaskTests.cs
```diff
@@ -144,6 +144,8 @@ public void TaskHelper_TaskCompletionSourceWithTimeout_Does_Not_Invoke_Delegate_
             Assert.False(called);
         }
 
+        [Ignore("this test is not reliable at all, ReSharper shows there are " +
```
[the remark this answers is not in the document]
joao-r-reis: I didn't realize the test was just looking for the production code while the analysis was going through all the code including the tests. Despite that, the original reason for this `Ignore` was that the test was failing in two parts:

```
using (cancellationToken.Register(() => cancellationCompletionSource.TrySetResult(true)))
            {
                if (task != await Task.WhenAny(task, cancellationCompletionSource.Task).ConfigureAwait(false))
                {
                    throw new OperationCanceledException(cancellationToken);
                }
            }
```

```
 /// <summary>
        /// Designed to await tasks with a cancellation token when the method that returns the task doesn't
        /// accept a token.
        /// </summary>
        /// <param name="task"></param>
```

I originally tried to find a regex that would not result in false positives for these two cases and I didn't find one. After doing the analysis I thought there were also false negatives so I decided to `Ignore` it. 

Now that I see that this is not the case I've changed the regex to fix the first false positive (the regex always expected a `;`) and changed the text of the comment from `await` to `Await` so that the regex doesn't pick it up.

What do you think about the new regex?

src/Cassandra/Metadata.cs
joao-r-reis: Hmm I agree, the introduction of `CancellationToken` objects should be part of a larger initiative or release where it is introduced in multiple places of the API since it is not particularly relevant in this specific method. 

I'll remove the `CancellationToken` parameter but I'll leave the cancellation helper methods in there for future use, if that's ok.

src/Cassandra/Metadata.cs
joao-r-reis: Nice! Changed it

src/Cassandra/Requests/RequestExecution.cs
joao-r-reis: I definitely agree but I prefer to create a ticket for that just because that is something that can affect (or even break) some other things and this task is just a new feature. 

If this wasn't one of my first tasks I probably would have changed it right away 😄 but I'd rather take it slowly.

src/Cassandra/RowPopulators/ExecutionInfo.cs
joao-r-reis: Good point

src/Cassandra/ProtocolOptions.cs
joao-r-reis: Fixed

src/Cassandra.IntegrationTests/Core/SchemaAgreementTests.cs
joao-r-reis: Fixed

src/Cassandra.IntegrationTests/Core/SchemaAgreementTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                var tableName = TestUtils.GetUniqueTableName().ToLower();
+                var cql = new SimpleStatement(
+                    $"CREATE TABLE {tableName} (id int PRIMARY KEY, description text)");
+                await _session.ExecuteAsync(cql).ConfigureAwait(false);
+                Assert.IsFalse(await _cluster.Metadata.CheckSchemaAgreementAsync().ConfigureAwait(false));
+            }
+            finally
+            {
+                TestCluster.ResumeNode(2);
```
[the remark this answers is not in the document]
joao-r-reis: This improves readability quite a bit on these tests, nice one.

src/Cassandra.Tests/TaskTests.cs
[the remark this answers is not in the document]
jorgebay: Nice! using negative lookahead is more solid that lookbehind on the last portion of the statement 👍

src/Cassandra/Metadata.cs
jorgebay: Sounds good.

jorgebay:
👍

PR #428, opened 2018-12-28
CSHARP-716 - Hopefully improved flaky test

joao-r-reis:
Changed `Parallel.For` to a `for` loop with one `Task.Factory.StartNew` call with `LongRunning` flag in each iteration, i.e., each iteration will have its own thread. `Parallel.For` is not suitable for I/O blocking code because most of the time only N threads will be used where N is the number of CPU Cores.

Hopefully this fixes the flakiness of the following integration test: `ClusterSharedSingleNodeTests.Cluster_Init_Keyspace_Race_Test`

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ClusterSharedSingleNodeTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ClusterSharedSingleNodeTests.cs
@@ -1,13 +1,14 @@
-﻿using System.Linq;
-using System.Net;
-using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


joao-r-reis:
Ok I've changed the test so it doesn't use the synchronous `Execute`. 

Since we only use `ExecuteAsync` now in this test, I increased from 200 to 2000 the number of tasks. It's kind of arbitrary but the effects on execution time are not noticeable and it increases the chance of triggering the race condition if it exists.

PR #429, opened 2019-01-04
CSHARP-708 - Make retries on the current host obtain a new connection instead of reusing the same one

joao-r-reis:
Here is a bit of an explanation about the changes that might seem a bit out of scope for this bug fix:

#### `Session` dependency
In order to fix this bug I would have to do a explicit conversion from `ISession` to `Session` because the relevant code is placed within `Session`'s `internal` methods. Either this or I could just change `ISession` references to `Session` ones. 

However, in order to avoid creating more dependencies to the concrete implementation (`Session`) I created a new interface `IInternalSession` that declares every `internal` method in `Session` and changed almost all `Session` references to `IInternalSession`.

#### `IRequestHandler`, `IRequestExecution` and `IConnection` interfaces
I extracted these internal interfaces to make it easier to mock dependencies and therefore allow for more isolated unit tests. With these interfaces I was able to create some unit tests that only target `RequestExecution` and have no dependencies to other classes.

#### `RequestHandler` and `RequestExecution` refactor
Also in this PR I tried to simplify the main methods of `RequestHandler` and `RequestExecution` classes that deal with connections while also adding some XML documentation to them.

### DIFF
--- a/src/Cassandra.Tests/RequestExecutionTests.cs
+++ b/src/Cassandra.Tests/RequestExecutionTests.cs
@@ -19,7 +19,9 @@ using System.Collections.Generic;
 using System.Net;
 
 using Cassandra.Requests;
+using Cassandra.Responses;
 using Cassandra.Serialization;
+
 using Moq;
 
 using NUnit.Framework;
@@ -29,8 +31,8 @@ namespace Cassandra.Tests
     [TestFixture]
     public class RequestExecutionTests
     {
-        [Test]
-        public void Should_ThrowException_WhenNoValidHosts()
+        [Test, TestCase(true), TestCase(false)]
+        public void Should_ThrowException_When_NoValidHosts(bool currentHostRetry)
         {
             var mockSession = Mock.Of<IInternalSession>();
             var mockRequest = Mock.Of<IRequest>();
@@ -40,11 +42,11 @@ namespace Cassandra.Tests
                 .Throws(new NoHostAvailableException(new Dictionary<IPEndPoint, Exception>()));
             var sut = new ProxyRequestExecution(mockRequestExecution, mockSession, mockRequest);
 
-            Assert.Throws<NoHostAvailableException>(() => sut.Start(false));
+            Assert.Throws<NoHostAvailableException>(() => sut.Start(currentHostRetry));
         }
-        
-        [Test]
-        public void Should_NotThrowException_WhenAValidHostIsObtained()
+
+        [Test, TestCase(true), TestCase(false)]
+        public void Should_NotThrowException_When_AValidHostIsObtained(bool currentHostRetry)
         {
             var mockSession = Mock.Of<IInternalSession>();
             var mockRequest = Mock.Of<IRequest>();
@@ -52,7 +54,6 @@ namespace Cassandra.Tests
             var host = new Host(
                 new IPEndPoint(IPAddress.Parse("127.0.0.1"), 9047),
                 new ConstantReconnectionPolicy(1));
-            host.BringUpIfDown();
             var validHost = ValidHost.New(
                 host,
                 HostDistance.Local);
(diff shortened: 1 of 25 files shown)


src/Cassandra.Tests/RequestExecutionTests.cs
```diff
@@ -0,0 +1,184 @@
...
+
+using NUnit.Framework;
+
+namespace Cassandra.Tests
+{
+    [TestFixture]
+    public class RequestExecutionTests
+    {
+        [Test, TestCase(true), TestCase(false)]
+        public void Should_ThrowException_When_NoValidHosts(bool currentHostRetry)
```
jorgebay: `RequestExecution` can be unit tested now, nice!

src/Cassandra/Metadata.cs
```diff
@@ -494,19 +494,19 @@ internal bool WaitForSchemaAgreement(Connection connection)
             }
             var start = DateTime.Now;
             var waitSeconds = Configuration.ProtocolOptions.MaxSchemaAgreementWaitSeconds;
-            Logger.Info("Waiting for schema agreement");
+            Metadata.Logger.Info("Waiting for schema agreement");
```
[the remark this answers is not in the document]
joao-r-reis: It improves readability imo, it makes it clear when we are performing a static access. This way we can clearly distinguish between instance members/properties or even local variables/parameters and static ones. Some people go even further and use `this.` to clearly distinguish between instance members/properties and local variables/parameters but that might be too much.

In their c# conventions for samples and documentation, Microsoft recommends the qualifier for static access but doesn't say anything about `this.` here: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -1,84 +1,150 @@
...
-                //The request has not been sent yet
-                return;
+                SendToCurrentHostAsync().Forget();
+                return _host;
+            }
+
+            // fail fast: try to choose a host before leaving this thread
+            var validHost = _parent.GetNextValidHost(_triedHosts);
+
+            SendToNextHostAsync(validHost).Forget();
```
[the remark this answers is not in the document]
joao-r-reis: The purpose is to make it really clear that this is a fire and forget call at the same time that we use `async` methods. Glad you like it 😄

src/Cassandra.Tests/RequestExecutionTests.cs
joao-r-reis: 🎉

src/Cassandra/Session.cs
```diff
@@ -40,10 +37,12 @@ public class Session : ISession
         private int _disposed;
         private volatile string _keyspace;
 
-        public int BinaryProtocolVersion { get { return (int)_serializer.ProtocolVersion; } }
+        internal IInternalSession InternalRef => this;
```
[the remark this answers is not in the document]
jorgebay: oh, makes sense!

src/Cassandra/Metadata.cs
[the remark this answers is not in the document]
jorgebay: sounds good to me!

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -1,84 +1,150 @@
...
         {
-            if (!useCurrentHost)
+            try
             {
-                //Get a new connection from the next host
-                _parent.GetNextConnection(_triedHosts).ContinueWith(t =>
+                do
                 {
-                    if (t.Exception != null)
+                    _host = validHost.Host;
```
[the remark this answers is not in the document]
joao-r-reis: Nice, changed it!

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -1,84 +1,150 @@
...
-                throw new DriverInternalError("No current connection set");
+                HandleResponse(ex, null);
             }
-            Send(_request, HandleResponse);
         }
 
-        private void TryStartNew(bool useCurrentHost)
+        /// <summary>
+        /// Wrapper around <see cref="Start"/> that catches exceptions and invokes the exception handler.
+        /// Useful method to restart the execution from a different thread other than the main thread
```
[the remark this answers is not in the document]
joao-r-reis: You're right, after the changes the summary seemed a bit weird. I've changed it a bit.

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -1,84 +1,152 @@
...
+                
+                _connection = connection;
+                _host = validHost.Host;
+                Send(_request, HandleResponse);
             }
-            if (_connection == null)
+            catch (Exception ex)
             {
-                throw new DriverInternalError("No current connection set");
+                _host = validHost.Host;
```
jorgebay: Is this needed? It looks like it's setting the host to the last valid host (not necessary the one with connections).

src/Cassandra.Tests/RequestExecutionTests.cs
```diff
@@ -0,0 +1,184 @@
...
+            var host = new Host(
+                new IPEndPoint(IPAddress.Parse("127.0.0.1"), 9047),
+                new ConstantReconnectionPolicy(1));
+            var validHost = ValidHost.New(
+                host,
+                HostDistance.Local);
+            Mock.Get(mockRequestExecution)
+                .SetupSequence(m => m.GetNextValidHost(It.IsAny<Dictionary<IPEndPoint, Exception>>()))
+                .Returns(validHost)
+                .Returns(null);
```
jorgebay: The caller of this method doesn't expect a `null` value, causing an infinite recursion.
The second time it should throw a `NoHostAvailableException` instance.

joao-r-reis: Hmm that actually brings up an interesting question, what if a user provides a custom load balancing policy to the driver that returns `null` instead of throwing an exception? I should probably add something to check that scenario and treat `null` the same way as thrown exceptions, right?

jorgebay: Its a good question... I think `GetNextValidHost()` already validates for `null`, so callers should not expect `null` values. Its only a matter of the mock that's behaving in a forbidden way.

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -1,84 +1,152 @@
...
+                HandleResponse(ex, null);
```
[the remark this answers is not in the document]
joao-r-reis: I may be misunderstanding these two methods (`HandleException` and `HandleResponse`) but from what I can tell `HandleResponse` does a couple more things than the the former. It handles exceptions that may be thrown by `HandleException` and sets the parent request as complete in this case.

src/Cassandra.Tests/RequestExecutionTests.cs
joao-r-reis: Ah silly me you're right. Changed it!

src/Cassandra/Requests/RequestExecution.cs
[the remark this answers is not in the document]
joao-r-reis: Also forgot to mention, what I said might be the reason why `HandleResponse` was being used here before.

src/Cassandra/Requests/RequestExecution.cs
jorgebay: Ah I see... its to retry on the same host 🤦‍♂️  
thanks for the out of band explanation.

src/Cassandra/Requests/RequestExecution.cs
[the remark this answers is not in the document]
jorgebay: nvm, calling `HandleResponse()` makes sense here as the exception can be thrown when obtaining the connection or during the first phase of the sending.

src/Cassandra/Requests/RequestHandler.cs
```diff
@@ -247,54 +240,95 @@ private Host GetNextHost()
...
         /// </summary>
-        internal static async Task<Connection> GetConnectionFromHost(Host host, HostDistance distance, Session session,
-                                                                     IDictionary<IPEndPoint, Exception> triedHosts)
+        /// <param name="host">Host to which a connection will be obtained.</param>
+        /// <param name="distance">Output parameter that will contain the <see cref="HostDistance"/> associated with
+        /// <paramref name="host"/>. It is retrieved from the current <see cref="ILoadBalancingPolicy"/>.</param>
+        /// <param name="session">Session from where a connection will be obtained (or created).</param>
+        /// <param name="triedHosts">Hosts for which there were attempts to connect and send the request.</param>
+        /// <exception cref="InvalidQueryException">When the keyspace is not valid</exception>
+        /// <exception cref="NoHostAvailableException"></exception>
```
jorgebay: NIT: this exception is not thrown by this method, it stood out to me its weird to get `NoHostAvailableException` when the host is the parameter.

joao-r-reis: nice catch, fixed

src/Cassandra/Requests/RequestExecution.cs
jorgebay: This statement was throwing me off because, in the case of the sending, the `_host` was set before calling `Sending()`, so leaving `HandleResponse(ex, null)` for the case when you can't obtain a connection was weird to me.

jorgebay: But I see that `GetConnectionToValidHostAsync()` already handles all the exception (except keyspace switches, which is fine), so this catch will only catch unexpected exceptions and that's OK to me.

PR #430, opened 2019-01-17
CSHARP-731 - Add netstandard 2.0 target

joao-r-reis:


src/Cassandra.IntegrationTests/Cassandra.IntegrationTests.csproj
```diff
@@ -9,9 +9,29 @@
...
+    <ProjectReference Include="..\Cassandra\Cassandra.csproj">
+      <SetTargetFramework>TargetFramework=net45</SetTargetFramework>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">
+    <ProjectReference Include="..\Cassandra.Tests\Cassandra.Tests.csproj">
+      <SetTargetFramework>TargetFramework=netcoreapp2.1</SetTargetFramework>
+    </ProjectReference>
+    <ProjectReference Include="..\Cassandra\Cassandra.csproj">
+      <SetTargetFramework>TargetFramework=netstandard2.0</SetTargetFramework>
```
jorgebay: Nice trick to test multiple targets

jorgebay: It would be nice to have a way to test that `netcoreapp2.0` on tests map to `netstandard1.5` in case we inadvertently change something in the future that causes both test targets to test `netstandard2.0`... but its not something we have to do now.

.travis.yml
```diff
@@ -1,16 +1,22 @@
...
+    
+matrix:
+  include:
+    - env: TARGETF=netcore21
+    - env: TARGETF=netcore20
 
 script:
   - dotnet restore src
-  - dotnet test src/Cassandra.Tests/Cassandra.Tests.csproj -c Release -f netcoreapp2.0
+  - if [ "$TARGETF" = "netcore20" ]; then dotnet test src/Cassandra.Tests/Cassandra.Tests.csproj -c Release -f netcoreapp2.0; fi
```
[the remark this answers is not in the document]
joao-r-reis: Welp that's so much better 😄 changed it

build.yaml
```diff
@@ -56,8 +57,13 @@ build:
           dotnet restore src
 
           # Run the tests
-          dotnet test src/Cassandra.IntegrationTests/Cassandra.IntegrationTests.csproj -f netcoreapp2.0 -c Release --filter TestCategory=short --logger "trx;LogFileName=../../../TestResult_trx.xml"
-          java -jar saxon/saxon9he.jar -o:TestResult.xml TestResult_trx.xml tools/trx-to-junit.xslt
+          if [ $CSHARP_VERSION = 'netcore20' ]; then
```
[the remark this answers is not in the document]
joao-r-reis: Done, we don't even need that condition here when we use the actual target as the env variable. We only need the existing `if else` for `mono`.

src/Cassandra.IntegrationTests/Cassandra.IntegrationTests.csproj
joao-r-reis: I've added a `TargetTests` class that does this if I understood correctly. Can you check if that's what you were thinking of?

jorgebay: Yes! 👍

PR #431, opened 2019-01-24
[CSHARP-726] - improved token map building process

joao-r-reis:
~~Changed the way `TokenMap.Build` caches token replica information. With this change, the token replica information is computed once per unique replication configuration instead of being computed once per keyspace.~~

Included `SimpleStrategy` in the cache used during building of token maps. 
Also took the chance to refactor this part of the code and hopefully improve its readability.


src/Cassandra/MetadataHelpers/SimpleStrategy.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public bool Equals(SimpleStrategy other)
+        {
+            return other != null && _replicationFactor == other._replicationFactor;
+        }
+
+        public override int GetHashCode()
+        {
+            return -1194751085 + _replicationFactor.GetHashCode();
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/MetadataHelpers/NetworkTopologyStrategy.cs
```diff
@@ -0,0 +1,237 @@
...
+                var token = ring[i];
+                var context = ComputeReplicasForToken(ring, primaryReplicas, datacenters, i);
+
+                replicas[token] = context.TokenReplicas;
+            }
+
+            return replicas;
+        }
+
+        private NetworkTopologyTokenMapContext ComputeReplicasForToken(
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/MetadataHelpers/NetworkTopologyStrategy.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            return other != null && _replicationFactorsSet.SetEquals(other._replicationFactorsSet);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hash = 0;
+                foreach (var item in _replicationFactorsSet)
```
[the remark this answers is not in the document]
joao-r-reis: Oh wow I would never have thought of that, nice catch. Changed to use `SortedSet`

src/Cassandra/TokenMap.cs
```diff
@@ -77,187 +74,32 @@ public static TokenMap Build(string partitioner, ICollection<Host> hosts, IColle
...
                     }
                 }
             }
             var ring = new List<IToken>(allSorted);
             var tokenToHosts = new Dictionary<string, Dictionary<IToken, ISet<Host>>>(keyspaces.Count);
-            var ksTokensCache = new Dictionary<string, Dictionary<IToken, ISet<Host>>>();
+            var ksTokensCache = new Dictionary<IReplicationStrategy, Dictionary<IToken, ISet<Host>>>();
             //Only consider nodes that have tokens
-            var hostCount = hosts.Count(h => h.Tokens.Any());
+            var hostsWithTokens = hosts.Where(h => h.Tokens.Any()).ToList();
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/MetadataHelpers/DatacenterReplicationFactor.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            return other != null &&
+                   Datacenter == other.Datacenter &&
+                   ReplicationFactor == other.ReplicationFactor;
+        }
+
+        public override int GetHashCode()
+        {
+            var hashCode = -1601459050;
+            hashCode = hashCode * -1521134295 + Datacenter?.GetHashCode() ?? string.Empty.GetHashCode();
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/MetadataHelpers/DatacenterReplicationFactor.cs
```diff
@@ -0,0 +1,52 @@
...
+            hashCode = hashCode * -1521134295 + ReplicationFactor.GetHashCode();
+            return hashCode;
```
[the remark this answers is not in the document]
joao-r-reis: Oh I wasn't aware of this method, thanks for the tip. Changed code to use the method here and in `NetworkTopologyStrategy.ComputeHashCode()`

src/Cassandra/MetadataHelpers/OrderedHashSet.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Collections.Generic;
+
+namespace Cassandra.MetadataHelpers
+{
+    /// <summary>
+    /// Ordered hash set designed for building token to replica maps (specifically for keyspaces configured with <see cref="NetworkTopologyStrategy"/>)
+    /// <para></para>
+    /// See <see cref="NetworkTopologyStrategy.ComputeTokenToReplicaMap"/>.
+    /// </summary>
+    internal class OrderedHashSet<T> : ISet<T>
```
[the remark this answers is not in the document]
joao-r-reis: Makes sense, done

src/Cassandra/TokenMap.cs
```diff
@@ -77,187 +77,38 @@ public static TokenMap Build(string partitioner, ICollection<Host> hosts, IColle
...
-                    var nextReplica = primaryReplicas[ring[nextReplicaIndex]];
-                    replicas.Add(nextReplica);
-                    j++;
-                }
-                tokenToReplicas.Add(token, replicas);
-            }
-            return tokenToReplicas;
+            sw.Stop();
+            TokenMap.Logger.Info(
+                "Finished building TokenMap for {0} keyspaces and {1} hosts. It took {2} milliseconds.",
```
jorgebay: We should round it:

```suggestion
                "Finished building TokenMap for {0} keyspaces and {1} hosts. It took {2:0} milliseconds.", 
```

joao-r-reis:
Implemented a way to avoid full `TokenMap.Build` when a keyspace changes. Also only one build process will occur at a time.

src/Cassandra/Cassandra.csproj
```diff
@@ -50,5 +50,6 @@
     <PackageReference Include="System.Net.Security" Version="4.0.0" />
     <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Include="System.Threading.Thread" Version="4.0.0" />
+    <PackageReference Include="System.Threading.ThreadPool" Version="4.3.0" />
```
joao-r-reis: not needed since the move to `ConcurrentExclusiveSchedulerPair`

src/Cassandra/Metadata.cs
```diff
@@ -133,14 +148,120 @@ public IEnumerable<IPEndPoint> AllReplicas()
...
+        }
+
+        internal async Task<KeyspaceMetadata> UpdateTokenMapForKeyspace(string name)
         {
-            Logger.Info("Rebuilding token map");
-            if (Partitioner == null)
+            var currentTicks = DateTime.UtcNow.Ticks;
+            
+            // running this statement synchronously inside the exclusive scheduler deadlocks
+            var keyspaceMetadata = await _schemaParser.GetKeyspace(name).ConfigureAwait(false);
```
jorgebay: This can't be put inside a `StartNew()` block as is, as the continuation from the await will be made in a different thread.

I think that if we want to run it in single threaded fashion, it should be something like:

```c#
await factory.StartNew(async () => await _schemaParser.GetKeyspace(name).ConfigureAwait(false)).ConfigureAwait();
// Do checks about data expiration/validity
// ...
// Schedule a new one
await factory.StartNew(() => 
{
  // TODO: call token map
}).ConfigureAwait();

```

src/Cassandra/Metadata.cs
```diff
@@ -133,14 +148,120 @@ public IEnumerable<IPEndPoint> AllReplicas()
...
+            var task = _tokenMapTaskFactory.StartNew(() =>
+            {
+                var lastRebuildTicks = Interlocked.Read(ref _lastRebuildTicks);
+                if (lastRebuildTicks > currentTicks)
+                {
+                    return true;
+                }
+                Interlocked.Exchange(ref _lastRebuildTicks, currentTicks);
+                
+                Metadata.Logger.Info("Retrieving keyspaces metadata");
```
jorgebay: We could change this one to `"Updating keyspaces metadata"` and include one above schema parser call that logs the retrieval.

src/Cassandra/Metadata.cs
```diff
@@ -133,14 +148,120 @@ public IEnumerable<IPEndPoint> AllReplicas()
...
+            var task = _tokenMapTaskFactory.StartNew(
+                () =>
+                {
+                    var lastRebuildTicks = Interlocked.Read(ref _lastRebuildTicks);
+                    if (lastRebuildTicks > currentTicks)
+                    {
+                        return true;
+                    }
+                    
+                    if (_keyspacesLastUpdateTicks.TryGetValue(name, out var lastUpdateTicks) && lastUpdateTicks > currentTicks)
```
jorgebay: I think this check can be made here, before fetching the metadata.

src/Cassandra/Metadata.cs
joao-r-reis: Yes, I tried putting inside the factory with `.GetAwaiter().GetResult()` so that there's no continuations involved and everything runs on the same thread but unfortunately it deadlocks in some cases. 
Your code example would not guarantee that those two tasks execute sequentially though if I'm understanding it correctly. Something else might get scheduled between those two tasks, right?

src/Cassandra/Metadata.cs
```diff
@@ -38,6 +41,11 @@ public class Metadata : IDisposable
         private volatile TokenMap _tokenMap;
         private volatile ConcurrentDictionary<string, KeyspaceMetadata> _keyspaces = new ConcurrentDictionary<string,KeyspaceMetadata>();
         private volatile SchemaParser _schemaParser;
+        private volatile ConcurrentDictionary<string, long> _keyspacesLastUpdateTicks = new ConcurrentDictionary<string, long>();
+        private long _lastRebuildTicks = 0;
+        private readonly TaskFactory _tokenMapTaskFactory = new TaskFactory(
+            CancellationToken.None, TaskCreationOptions.DenyChildAttach | TaskCreationOptions.LongRunning,
```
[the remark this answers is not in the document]
joao-r-reis: My intention was exactly that, to avoid using tasks from the shared `ThreadPool` since this is not an asynchronous task but maybe my reasoning doesn't make much sense...

src/Cassandra/Cassandra.csproj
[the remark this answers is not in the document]
joao-r-reis: Done

src/Cassandra/Metadata.cs
jorgebay: yes, the sample I've made is no magic bullet.

It would be interesting to try to find ways to implement single threaded async code in the future with nice constructs.

joao-r-reis: My personal choice is [`ActionBlock` of the TPL Dataflow library](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.dataflow.actionblock-1?view=netcore-2.2) but it won't fix the root issue that is the "recursion" in these methods. 

I do think that I'm going to give it a go and use `ActionBlock` in the near future for some improvements (maybe in the next iteration).

src/Cassandra/Metadata.cs
```diff
@@ -133,14 +148,118 @@ public IEnumerable<IPEndPoint> AllReplicas()
             return Hosts.AllEndPointsToCollection();
         }
 
-        internal void RebuildTokenMap()
+        internal async Task<bool> RebuildTokenMapAsync(bool retry)
+        {
+            var currentTicks = DateTime.UtcNow.Ticks;
```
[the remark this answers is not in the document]
joao-r-reis: `Stopwatch` is not threadsafe though, I would have to wrap the call with a `lock`, which doesn't seem like a great idea.

src/Cassandra/Metadata.cs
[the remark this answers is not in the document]
joao-r-reis: Removed `TaskCreationOptions.LongRunning`

src/Cassandra/Metadata.cs
joao-r-reis: sounds good, done

src/Cassandra/Metadata.cs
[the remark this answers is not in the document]
jorgebay: I've edited the message to make it more clear, I've meant to avoid using system clock to measure time and use high resolution counters instead (when available). Using `DateTime.UtcNow` might get you a resolution of ~16 milliseconds (1000 / 60), which is not very good.

src/Cassandra/Metadata.cs
jorgebay: It looks like I've sent the message without checking if this made sense, sorry... I'm the worst at multi tasking...

I was thinking about moving this check outside of the new task, but looking at it more closely I see that given that the operation is not atomic (read, compare value and then right), it must be done within the ordered task scheduler.

eplowe:
If we are trying to be as accurate as possible, DateTime.Now/UtcNow/.Ticks
(System clock) is updated every 15 miloseconds (or 10 ms on some systems).
A StopWatch might suffer the same fate on other platforms. I think we might
need to check for the “IsHighResolution” property of StopWatch != true, and
handle as needed, if we are caring for high fidelity tick counts in this
particular piece of code.

eplowe:
Sorry, *milliseconds

src/Cassandra.IntegrationTests/MetadataTests/TokenMapSchemaChangeTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            var alterKeyspaceCql = $"ALTER KEYSPACE {KeyspaceName} WITH replication = {{'class': 'SimpleStrategy', 'replication_factor' : 3}}";
+            Session.Execute(alterKeyspaceCql);
+            TestUtils.WaitForSchemaAgreement(Cluster);
+            TestHelper.RetryAssert(() =>
+            {
+                var replicas = Cluster.Metadata.TokenToReplicasMap.GetByKeyspace(KeyspaceName);
+                Assert.AreEqual(Cluster.Metadata.Hosts.Sum(h => h.Tokens.Count()), replicas.Count);
+                Assert.AreEqual(3, Cluster.Metadata.GetReplicas(KeyspaceName, Encoding.UTF8.GetBytes("123")).Count);
+            });
+            Assert.IsTrue(object.ReferenceEquals(Cluster.Metadata.TokenToReplicasMap, oldTokenMap));
```
jorgebay: I think the CI failures are due to the fact that this involves multiple threads updating a value (`Cluster.Metadata.TokenToReplicasMap`) which is volatile but the `oldTokenMap` write and reads have no memory barriers.

We should use `Volatile.Write()` and `Volatile.Read()` when accessing `oldTokenMap` to make sure that read/write instructions are not reordered. The same applies to the rest of the tests.

joao-r-reis: There's only one write to `oldTokenMap` before the actual test begins, so I don't see the need to use `Volatile.Write()` and `Volatile.Read()`. There was only one test that was failing in CI which is `TokenMapUpdates_Should_FunctionCorrectly_When_MultipleThreadsCallingRebuildAndUpdateKeyspace` (unit test). It failed only once since the introduction of the `ticks` checks but it might have been because of the resolution like it was mentioned previously.

joao-r-reis:
I should have given it a bit more thought when I decided to use `ticks` in this case. I'm thinking about changing it to a simple counter that is updated with the `Interlocked` API.

jorgebay:
> I'm thinking about changing it to a simple counter that is updated with the Interlocked API.

The resolution of `Stopwatch.GetTimestamp()` will be high on all, properly configured, modern systems. On Windows systems, it will use a native call to [QueryPerformanceCounter](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) so it's what we are looking for in this case (you can expect other things to go wrong when this is not true).

src/Cassandra.IntegrationTests/MetadataTests/TokenMapSchemaChangeTests.cs
joao-r-reis: That failure might be because those tests are sharing a `Cluster` instance. I've changed them to create their own instance.

joao-r-reis:
> 
> 
> > I'm thinking about changing it to a simple counter that is updated with the Interlocked API.
> 
> The resolution of `Stopwatch.GetTimestamp()` will be high on all, properly configured, modern systems. On Windows systems, it will use a native call to [QueryPerformanceCounter](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) so it's what we are looking for in this case (you can expect other things to go wrong when this is not true).

Yeah but I would have to synchronize access to a `StopWatch` instance and I do believe that a simple counter will be much much better and simpler, don't you think?

src/Cassandra.IntegrationTests/MetadataTests/TokenMapSchemaChangeTests.cs
```diff
@@ -36,58 +37,85 @@ public TokenMapSchemaChangeTests() : base(3, true, true, new TestClusterOptions
...
+            newSession.Execute(removeKeyspaceCql);
+            TestUtils.WaitForSchemaAgreement(newCluster);
+            var oldTokenMap = newCluster.Metadata.TokenToReplicasMap;
             TestHelper.RetryAssert(() =>
             {
-                Assert.Throws<KeyNotFoundException>(() => Cluster.Metadata.TokenToReplicasMap.GetByKeyspace(KeyspaceName));
+                Assert.Throws<KeyNotFoundException>(() => newCluster.Metadata.TokenToReplicasMap.GetByKeyspace(keyspaceName));
             });
-            Assert.IsTrue(object.ReferenceEquals(Cluster.Metadata.TokenToReplicasMap, oldTokenMap));
+            var asd = Stopwatch.GetTimestamp();
```
jorgebay: Probably unintended :)

src/Cassandra/Metadata.cs
```diff
@@ -41,12 +40,15 @@ public class Metadata : IDisposable
         private volatile TokenMap _tokenMap;
         private volatile ConcurrentDictionary<string, KeyspaceMetadata> _keyspaces = new ConcurrentDictionary<string, KeyspaceMetadata>();
         private volatile SchemaParser _schemaParser;
-        private volatile ConcurrentDictionary<string, long> _keyspacesLastUpdateTicks = new ConcurrentDictionary<string, long>();
-        private long _lastRebuildTicks = 0;
+        private volatile ConcurrentDictionary<string, long> _keyspacesUpdateCounters = new ConcurrentDictionary<string, long>();
```
jorgebay: No need for `volatile` access, use `readonly` instead.

src/Cassandra/Metadata.cs
```diff
@@ -142,29 +144,49 @@ public ICollection<Host> AllHosts()
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private bool RebuildNecessary(long currentCounter)
+        {
+            if (AnotherRebuildCompleted(currentCounter))
```
jorgebay: Suggestion: this can be simplified by using `Interlocked.CompareExchange()`, with the current value as _comparand_ and use `AnotherRebuildCompleted()` only for single keyspace update.

joao-r-reis: I should have done that to begin with. Changed!

src/Cassandra/Metadata.cs
joao-r-reis: True! changed

src/Cassandra.IntegrationTests/MetadataTests/TokenMapSchemaChangeTests.cs
joao-r-reis: Oops, changed

PR #432, opened 2019-02-11
Add 3.8 to docs.yaml

jorgebay:
Add `3.8` version to `docs.yaml`.


jorgebay:
Once this is merged into `3.8`, we should merge it into `master` for the docs to be generated.

joao-r-reis:
Thanks!

PR #433, opened 2019-02-18
[CSHARP-618] - added DRIVER_NAME and DRIVER_VERSION startup options

joao-r-reis:
These options are going to be added in C* 4.0 but it is safe to send them in all C* versions.

### DIFF
--- a/src/Cassandra.Tests/Requests/StartupOptionsFactoryTests.cs
+++ b/src/Cassandra.Tests/Requests/StartupOptionsFactoryTests.cs
@@ -37,19 +37,16 @@ namespace Cassandra.Tests.Requests
             Assert.AreEqual(5, options.Count);
             Assert.AreEqual("snappy", options["COMPRESSION"]);
             Assert.AreEqual("true", options["NO_COMPACT"]);
-            Assert.AreEqual("DataStax C# Driver", options["DRIVER_NAME"]);
+            var driverName = options["DRIVER_NAME"];
+            Assert.True(driverName.Contains("DataStax") && driverName.Contains("C# Driver"), driverName);
             Assert.AreEqual("3.0.0", options["CQL_VERSION"]);
-            var assemblyVersion = MultiTargetHelpers.GetAssembly(typeof(Cluster)).GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+
+            var assemblyVersion = AssemblyHelpers.GetAssembly(typeof(Cluster)).GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
             Assert.AreEqual(assemblyVersion, options["DRIVER_VERSION"]);
             var indexOfVersionSuffix = assemblyVersion.IndexOf('-');
             var versionPrefix = indexOfVersionSuffix == -1 ? assemblyVersion : assemblyVersion.Substring(0, indexOfVersionSuffix);
             var version = Version.Parse(versionPrefix);
-            Assert.LessOrEqual(version.Major, 4);
-            Assert.GreaterOrEqual(version.Major, 3);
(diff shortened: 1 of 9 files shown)


src/Cassandra/Requests/StartupOptionsFactory.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    internal class StartupOptionsFactory : IStartupOptionsFactory
+    {
+        public const string CqlVersionOption = "CQL_VERSION";
+        public const string CompressionOption = "COMPRESSION";
+        public const string NoCompactOption = "NO_COMPACT";
+        public const string DriverNameOption = "DRIVER_NAME";
+        public const string DriverVersionOption = "DRIVER_VERSION";
+
+        public const string DriverName = "DataStax C# Driver";
```
jorgebay: Would it be possible to obtain this information from the assembly product name (`Title` in the `csproj` file), that way it won't require any adaptation on the DSE driver.

src/Cassandra/Helpers/MultiTargetHelpers.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Cassandra.Helpers
+{
+    internal static class MultiTargetHelpers
+    {
+        public static Assembly GetAssembly(Type type)
+        {
+#if NETSTANDARD1_5
+            return type.GetTypeInfo().Assembly;
+#else
+            return type.Assembly;
```
jorgebay: I think `type.GetTypeInfo().Assembly` works on both targets.

src/Cassandra.Tests/Requests/StartupOptionsFactoryTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            Assert.AreEqual("snappy", options["COMPRESSION"]);
+            Assert.AreEqual("true", options["NO_COMPACT"]);
+            Assert.AreEqual("DataStax C# Driver", options["DRIVER_NAME"]);
+            Assert.AreEqual("3.0.0", options["CQL_VERSION"]);
+            var assemblyVersion = MultiTargetHelpers.GetAssembly(typeof(Cluster)).GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+            Assert.AreEqual(assemblyVersion, options["DRIVER_VERSION"]);
+            var indexOfVersionSuffix = assemblyVersion.IndexOf('-');
+            var versionPrefix = indexOfVersionSuffix == -1 ? assemblyVersion : assemblyVersion.Substring(0, indexOfVersionSuffix);
+            var version = Version.Parse(versionPrefix);
+            Assert.LessOrEqual(version.Major, 4);
```
jorgebay: NIT: I think comparing against assembly version (and validating that assembly version contains x.y.z >= 1.0 is enough. I personally prefer having a looser test than having to adapt it on both drivers by version.

In any case, comparisons between `Version` instances can be used: `Assert.Greater(version1, new Version(1, 0))`.

src/Cassandra/Requests/StartupOptionsFactory.cs
joao-r-reis: good suggestion, changed it

src/Cassandra/Helpers/MultiTargetHelpers.cs
joao-r-reis: hmm you're right, not sure why I missed that, changed it!

src/Cassandra.Tests/Requests/StartupOptionsFactoryTests.cs
joao-r-reis: Makes sense, I've changed it so that it doesn't have to change in dse driver and uses `Version` instances for comparison.

PR #434, opened 2019-02-19
Moved startup options factory initialization to Configuration object

joao-r-reis:
This change removes the direct dependency between `Connection` and `StartupOptionsFactory`


jorgebay:
It looks like we need to add validation to `StartupRequest` constructor parameter.

joao-r-reis:
> 
> 
> It looks like we need to add validation to `StartupRequest` constructor parameter.

Yep, added the validation and removed the Mocked instances from integration tests that were causing the error.

PR #437, opened 2019-02-25
Improvements ControlConnection

joao-r-reis:
Added method to controlconnection that offers a way to send a query request without expecting rows as a result


jorgebay:
Makes sense, +1

PR #439, opened 2019-02-28
[CSHARP-725] Deprecate usedHostsPerRemoteDc in DCAwareRoundRobinPolicy

joao-r-reis:
Deprecating the parameter to give a heads up to the users that this functionality will be removed in the next major release.
The reason behind this change is the fact that the driver does not have enough context to handle DC Failover. https://datastax-oss.atlassian.net/browse/NODEJS-485 has a good explanation behind this change.

### DIFF
--- a/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
+++ b/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
@@ -23,19 +23,26 @@ using System.Net;
 namespace Cassandra
 {
     /// <summary>
-    ///  A data-center aware Round-robin load balancing policy. <p> This policy
-    ///  provides round-robin queries over the node of the local datacenter. It also
-    ///  includes in the query plans returned a configurable number of hosts in the
-    ///  remote datacenters, but those are always tried after the local nodes. In
-    ///  other words, this policy guarantees that no host in a remote datacenter will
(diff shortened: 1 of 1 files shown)


src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
```diff
@@ -23,19 +23,26 @@
...
-    ///  policy could be preferred to this policy in that case.</p>
+    /// A data-center aware Round-robin load balancing policy.
+    /// <para>
+    /// This policy provides round-robin queries over the node of the local datacenter. Currently, it also includes in the query plans
+    /// returned a configurable number of hosts in the remote datacenters (which are always tried after the local nodes)
+    /// but this functionality will be removed in the next major version of the driver.
+    /// See the comments on <see cref="DCAwareRoundRobinPolicy(string, int)"/> for more information.
+    /// </para>
+    /// <para>
+    /// If used with a single datacenter, this policy is equivalent to the <see cref="RoundRobinPolicy"/> policy,
```
jorgebay: I think its a good opportunity to remove this last paragraph, as it doesn't make much sense.

joao-r-reis: 👍 done

PR #440, opened 2019-03-19
CSHARP-744 Support legacy empty column names

joao-r-reis:
From CSHARP-744 description:

> Usually {{""}} is not a valid column name, but it can occasionally appear when migrating Thrift tables with COMPACT STORAGE to modern storage tables. 

Metadata API (i.e. `GetTable`) already supports this but LINQ and Mapper don't.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
@@ -1,0 +1,275 @@
+﻿//
+//       Copyright (C) 2019 DataStax Inc.
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+
+using System.Collections.Generic;
+using System.Linq;
+
+using Cassandra.Data.Linq;
+using Cassandra.IntegrationTests.TestClusterManagement.Simulacron;
+using Cassandra.Mapping;
+
+using NUnit.Framework;
+
+namespace Cassandra.IntegrationTests.Core
+{
+    [TestFixture]
+    public class EmptyColumnTests
+    {
+        [Test]
+        public void Should_ReturnCorrectValue_When_EmptyColumnName()
+        {
+            using (var simulacronCluster = SimulacronCluster.CreateNew(3))
+            {
+                simulacronCluster.Prime(new
+                {
+                    when = new
+                    {
+                        query = string.Format(
+                            "SELECT * FROM system_schema.tables WHERE table_name='{0}' AND keyspace_name='{1}'",
+                            "testtable",
+                            "testks")
+                    },
+                    then = new
+                    {
+                        result = "success",
+                        delay_in_ms = 0,
+                        rows = new[]
+                        {
+                                new
+                                {
+                                    compression = new { },
+                                    compaction = new { },
+                                    bloom_filter_fp_chance = 0.1,
+                                    caching = new { keys = "ALL", rows_per_partition = "NONE" },
+                                    comment = "comment",
+                                    gc_grace_seconds = 60000,
+                                    dclocal_read_repair_chance = 0.1,
+                                    read_repair_chance = 0.1,
+                                    keyspace_name = "testks"
+                                }
+                            },
+                        column_types = new
+                        {
+                            compression = "map<ascii, ascii>",
+                            compaction = "map<ascii, ascii>",
+                            bloom_filter_fp_chance = "double",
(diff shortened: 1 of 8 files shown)


src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
```diff
@@ -0,0 +1,275 @@
...
+        public void Should_ReturnCorrectValue_When_EmptyColumnName()
+        {
+            using (var simulacronCluster = SimulacronCluster.CreateNew(3))
+            {
+                simulacronCluster.Prime(new
+                {
+                    when = new
+                    {
+                        query = string.Format(
+                            "SELECT * FROM system_schema.tables WHERE table_name='{0}' AND keyspace_name='{1}'",
```
jorgebay: Nit: we could use string interpolation that is less error prone.

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
```diff
@@ -0,0 +1,275 @@
...
+
+                        var rs = session.Execute("SELECT \"\", \" \" FROM testks.testtable");
+                        var row = rs.SingleOrDefault();
+                        Assert.IsTrue(rs.Columns.Length == 2 && rs.Columns.Any(c => c.Name == string.Empty) && rs.Columns.Any(c => c.Name == " "));
+                        Assert.AreEqual("testval", row?.GetValue<string>(string.Empty));
+                        Assert.AreEqual("testval2", row?.GetValue<string>(" "));
+                    }
+                    finally
+                    {
+                        var a = "";
```
jorgebay: Probably leftover?

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
```diff
@@ -0,0 +1,275 @@
...
+                        var mapper = new Mapper(session, mapConfig);
+                        test = mapper.Fetch<TestTable>().FirstOrDefault();
+                        Assert.AreEqual("testval", test.TestColumn);
+                        Assert.AreEqual("testval2", test.TestColumn2);
+
+                        var tableMetadata = session.Cluster.Metadata.GetTable("testks", "testtable");
+                        Assert.IsNotNull(tableMetadata);
+
+                        var rs = session.Execute("SELECT \"\", \" \" FROM testks.testtable");
+                        var row = rs.SingleOrDefault();
```
jorgebay: NIT: We could use `Single()` that throws when there are no values, and avoid null conditional operators below.

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
```diff
@@ -0,0 +1,275 @@
...
+
+                        var mapper = new Mapper(session, mapConfig);
+                        test = mapper.Fetch<TestTable>().FirstOrDefault();
+                        Assert.AreEqual("testval", test.TestColumn);
+                        Assert.AreEqual("testval2", test.TestColumn2);
+
+                        var tableMetadata = session.Cluster.Metadata.GetTable("testks", "testtable");
+                        Assert.IsNotNull(tableMetadata);
+
+                        var rs = session.Execute("SELECT \"\", \" \" FROM testks.testtable");
```
jorgebay: Can we include a prepare statement execution?

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
joao-r-reis: oops

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
joao-r-reis: done

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
joao-r-reis: done

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
joao-r-reis: done

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
joao-r-reis: good idea, added

joao-r-reis:
Forgot to prime the queries for C* < 3.0.0, added them in a separate test. Also changed the test filter in `build.yaml` to match the one being used in appveyor.

src/Cassandra.IntegrationTests/Core/EmptyColumnTests.cs
```diff
@@ -0,0 +1,544 @@
...
+using Cassandra.Data.Linq;
+using Cassandra.IntegrationTests.TestBase;
+using Cassandra.IntegrationTests.TestClusterManagement.Simulacron;
+using Cassandra.Mapping;
+
+using NUnit.Framework;
+
+namespace Cassandra.IntegrationTests.Core
+{
+    [TestFixture]
```
jorgebay: In any case, I think we should add it to the `short` category.

joao-r-reis: Fair enough, I'll do it

PR #442, opened 2019-04-12
CSHARP-678 Execution profiles at session level

joao-r-reis:
This is rebased on top of #444 so whoever reviews this PR can see this smaller diff instead https://github.com/datastax/csharp-driver/compare/refactor-unittests..CSHARP-678. I will rebase the target branch after #444 is merged to master.


joao-r-reis:
Rebased after #444 was merged

src/Cassandra/Session.cs
```diff
@@ -252,49 +253,68 @@ private async Task Warmup()
...
         /// <inheritdoc />
         public PreparedStatement EndPrepare(IAsyncResult ar)
         {
             var task = (Task<PreparedStatement>)ar;
-            TaskHelper.WaitToComplete(task, Configuration.ClientOptions.QueryAbortTimeout);
+            TaskHelper.WaitToComplete(task, Configuration.DefaultRequestOptions.QueryAbortTimeout);
+            return task.Result;
+        }
+
+        public RowSet Execute(IStatement statement, string executionProfileName)
```
jorgebay: Missing `inheritdoc`.

src/Cassandra/Session.cs
```diff
@@ -252,49 +253,68 @@ private async Task Warmup()
...
         /// <inheritdoc />
         public Task<RowSet> ExecuteAsync(IStatement statement)
         {
             return Configuration.RequestHandlerFactory
-                                .Create(this, _serializer, statement)
+                                .Create(this, _serializer, statement, Configuration.DefaultRequestOptions)
+                                .SendAsync();
+        }
+
+        public Task<RowSet> ExecuteAsync(IStatement statement, string executionProfileName)
```
jorgebay: Same :)

src/Cassandra/ExecutionProfile.cs
```diff
@@ -0,0 +1,77 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// 
+
+using System;
+using Cassandra.ExecutionProfiles;
+
+namespace Cassandra
+{
+    public class ExecutionProfile
```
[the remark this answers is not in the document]
joao-r-reis: I agree with extracting an interface. Where should the `ExecutionProfile.Builder` method go in this case? We either remove it and have users just do `new ExecutionProfileBuilder()` or we can move it to the cluster builder class? `Builder.ExecutionProfileBuilder()`?

src/Cassandra/Builder.cs
```diff
@@ -674,6 +677,36 @@ public Builder WithMaxSchemaAgreementWaitSeconds(int maxSchemaAgreementWaitSecon
             return this;
         }
 
+        /// <summary>
+        /// Adds Execution Profiles to the Cluster instance.
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/ExecutionProfiles/IRequestOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        ITimestampGenerator TimestampGenerator { get; }
+
+        bool DefaultIdempotence { get; }
+
+        int QueryAbortTimeout { get; }
+
+        bool PrepareOnAllHosts { get; }
+
+        bool ReprepareOnUp { get; }
```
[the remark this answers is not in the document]
joao-r-reis: yeah that makes sense, done

src/Cassandra/Cluster.cs
```diff
@@ -474,7 +488,7 @@ public async Task ShutdownAsync(int timeoutMs = Timeout.Infinite)
             _metadata.ShutDown(timeoutMs);
             _controlConnection.Dispose();
             Configuration.Timer.Dispose();
-            Configuration.Policies.SpeculativeExecutionPolicy.Dispose();
+            Configuration.DefaultRequestOptions.SpeculativeExecutionPolicy.Dispose();
```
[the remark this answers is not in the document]
joao-r-reis: oops, nice catch 😄 added unit tests for it

src/Cassandra/Configuration.cs
```diff
@@ -114,6 +115,10 @@ public class Configuration
         internal IConnectionFactory ConnectionFactory { get; }
         
         internal IControlConnectionFactory ControlConnectionFactory { get; }
+        
+        internal IReadOnlyDictionary<string, IRequestOptions> RequestOptions { get; }
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/ExecutionProfile.cs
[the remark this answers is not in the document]
joao-r-reis: Maybe `Cluster.ExecutionProfileBuilder` is also a possibility since the method to create the cluster builder is also on the `Cluster` class. For now I've added it to the `Builder` class: `Builder.ExecutionProfileBuilder()`, let me know your thoughts

src/Cassandra/IExecutionProfileBuilder.cs
```diff
@@ -0,0 +1,35 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// 
+
+namespace Cassandra
+{
+    public interface IExecutionProfileBuilder
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/Policies/Policies.cs
```diff
@@ -59,6 +58,17 @@ public static IRetryPolicy DefaultRetryPolicy
                 return new DefaultRetryPolicy();
             }
         }
+        
+        /// <summary>
+        ///  The default extended retry policy.The default extended retry policy is <see cref="Cassandra.DefaultRetryPolicy"/>
+        /// </summary>
+        public static IExtendedRetryPolicy DefaultExtendedRetryPolicy
```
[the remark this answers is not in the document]
joao-r-reis: hmm yeah I noticed that, I guess it could cause some confusion to expose this... done!

src/Cassandra/Session.cs
```diff
@@ -252,49 +253,68 @@ private async Task Warmup()
...
         }
 
         /// <inheritdoc />
         public RowSet Execute(string cqlQuery)
         {
-            return Execute(new SimpleStatement(cqlQuery).SetConsistencyLevel(Configuration.QueryOptions.GetConsistencyLevel()).SetPageSize(Configuration.QueryOptions.GetPageSize()));
+            return Execute(GetDefaultStatement(cqlQuery));
+        }
+
+        public RowSet Execute(string cqlQuery, string executionProfileName)
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/Session.cs
joao-r-reis: done

src/Cassandra/Session.cs
joao-r-reis: done

src/Cassandra/Data/Linq/CqlQuerySingleElement.cs
```diff
@@ -77,7 +77,7 @@ public new TEntity EndExecute(IAsyncResult ar)
         /// </summary>
         public new TEntity Execute()
         {
-            var queryAbortTimeout = GetTable().GetSession().GetConfiguration()?.ClientOptions.QueryAbortTimeout ?? ClientOptions.DefaultQueryAbortTimeout;
+            var queryAbortTimeout = GetTable().GetSession().Cluster.Configuration.DefaultRequestOptions.QueryAbortTimeout;
```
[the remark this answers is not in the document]
joao-r-reis: hmm that makes sense, I'll create this comment on #445 since all of the work related to LINQ and Mapper for execution profiles is over there, this was just a quick fix.

src/Cassandra/ISession.cs
```diff
@@ -146,7 +153,14 @@ public interface ISession: IDisposable
         /// Executes the provided query.
         /// </summary>
         RowSet Execute(string cqlQuery);
-
+        
+        /// <summary>
+        /// Executes the provided query with the provided execution profile.
+        /// </summary>
+        /// <param name="cqlQuery">Query to execute.</param>
+        /// <param name="executionProfileName">ExecutionProfile name to be used while executing the statement.</param>
```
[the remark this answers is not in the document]
joao-r-reis: done

src/Cassandra/IExecutionProfileBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Sets the speculative execution policy. 
+        /// See <see cref="Builder.WithSpeculativeExecutionPolicy"/> and <see cref="ISpeculativeExecutionPolicy"/>
+        /// for additional context on this setting.
+        /// </summary>
+        /// <returns>This builder.</returns>
+        IExecutionProfileBuilder WithSpeculativeExecutionPolicy(ISpeculativeExecutionPolicy speculativeExecutionPolicy);
+        
+        /// <summary>
+        /// Sets the <code>SerialConsistencyLevel</code> setting.
+        /// See <see cref="QueryOptions.SetConsistencyLevel"/>, <see cref="QueryOptions.GetConsistencyLevel"/> and
```
jorgebay: I would avoid linking to other options to set the same value in the docs.

jorgebay:
What do you think about something like:

```
Cluster.Builder()
    .AddContactPoint("127.0.0.1")
    .WithExecutionProfiles(x => 
    {
        x.WithProfile("default", y => y.WithConsistencyLevel(ConsistencyLevel.LocalQuorum));
        x.WithProfile("timeseries", y => y
            .WithConsistencyLevel(ConsistencyLevel.LocalOne)
            .WithRetryPolicy(new LoggingRetryPolicy()));
	})
	.Build();
```

src/Cassandra/Configuration.cs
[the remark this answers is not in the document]
jorgebay: There is no `"default"` execution profile, like in node.js or python, is there a reason for it?

src/Cassandra.IntegrationTests/ExecutionProfiles/ExecutionProfileTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                    alive = 2,
+                    delay_in_ms = 0,
+                    message = "unavailable",
+                    ignore_on_prepare = false
+                }
+            };
+            _simulacron.Prime(primeQuery);
+
+            var exception = async
+                ? Assert.ThrowsAsync<UnavailableException>(() => session.ExecuteAsync(new SimpleStatement("SELECT * from test.test"), "write"))
```
jorgebay: I'm not sure I understand, shouldn't `"write"` exec profile use consistency level one and not fail?

src/Cassandra.IntegrationTests/ExecutionProfiles/ExecutionProfileTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            var rows = rs.ToList();
+            Assert.AreEqual(2, rows.Count);
+            Assert.AreEqual("test6", rows[0].GetValue<string>("text"));
+            Assert.AreEqual("test5", rows[1].GetValue<string>("text"));
+        }
+
+        [Test]
+        [TestCase(true)]
+        [TestCase(false)]
+        public async Task Should_UseClusterConsistency_When_ProfileIsNotProvided(bool async)
```
jorgebay: Nice test using simulacron.

joao-r-reis:
> 
> 
> What do you think about something like:
> 
> ```
> Cluster.Builder()
>     .AddContactPoint("127.0.0.1")
>     .WithExecutionProfiles(x => 
>     {
>         x.WithProfile("default", y => y.WithConsistencyLevel(ConsistencyLevel.LocalQuorum));
>         x.WithProfile("timeseries", y => y
>             .WithConsistencyLevel(ConsistencyLevel.LocalOne)
>             .WithRetryPolicy(new LoggingRetryPolicy()));
>     })
>     .Build();
> ```

Yeah the `IExecutionProfileOptions` supports this way already but it also has a method that takes the already built `IExecutionProfile` instance. Do you think that we should just remove the method take receives an already built profile instance? Here is the xml docs that I've added recently to show both ways that the user can go for when adding execution profiles: https://github.com/datastax/csharp-driver/blob/CSHARP-678/src/Cassandra/Builder.cs#L693-L722

jorgebay:
> Do you think that we should just remove the method take receives an already built profile instance?
It's funny that we came to the same methods, sorry I didn't check it existed already.
I'm +1 on removing the other, I think having just 1 way of doing is better.

src/Cassandra/ExecutionProfiles/ExecutionProfileOptions.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                throw new ArgumentNullException(nameof(baseProfile));
+            }
+
+            if (!_profiles.TryGetValue(baseProfile, out var baseProfileInstance))
+            {
+                throw new ArgumentException("Base Execution Profile must be added before the derived profile.");
+            }
+            
+            _profiles[name] = new ExecutionProfile(baseProfileInstance, profile);
```
jorgebay: I think we should actively throw an exception for duplicate names.

joao-r-reis:
> There is no "default" execution profile, like in node.js or python, is there a reason for it?
Discussed this offline, I will change the implementation to have a `"default"` profile which can be configured in the same way as the remaining profiles.

jorgebay:
I think we should include a doc entry for this, something like: https://github.com/datastax/nodejs-driver/tree/master/doc/features/execution-profiles

src/Cassandra.IntegrationTests/ExecutionProfiles/ExecutionProfileTests.cs
joao-r-reis: Ah I primed the request for the `"write"` profile and I'm using the `"write"` profile in the test but the derived profile is `"read"`. Changed the rtest to prime for consistency `"TWO"` and use `"read"` in the test.

joao-r-reis:
Should we also add a way to retrieve the execution profiles from the `Cluster`?

joao-r-reis:
> 
> 
> I think we should include a doc entry for this, something like: https://github.com/datastax/nodejs-driver/tree/master/doc/features/execution-profiles

Yeah I'm planning to do that after all of this is merged to the feature branch (related to https://datastax-oss.atlassian.net/browse/CSHARP-762)

jorgebay:
> Should we also add a way to retrieve the execution profiles from the `Cluster`?

It would be nice to have a read only property to get the execution profiles along with their names.

src/Cassandra/IExecutionProfileBuilder.cs
joao-r-reis: done, also added xml docs to `IExecutionProfile`

src/Cassandra/ExecutionProfiles/ExecutionProfileOptions.cs
joao-r-reis: done

joao-r-reis:
> 
> 
> > Should we also add a way to retrieve the execution profiles from the `Cluster`?
> 
> It would be nice to have a read only property to get the execution profiles along with their names.

Added a read only dictionary property getter to `Cluster` that allows a user to retrieve the execution profiles.

> 
> 
> > There is no "default" execution profile, like in node.js or python, is there a reason for it?
> 
> Discussed this offline, I will change the implementation to have a `"default"` profile which can be configured in the same way as the remaining profiles.

Done

Also noticed that I forgot to add `Prepare` methods that receive an execeution profile parameter so I've added them now to `Session`.

src/Cassandra/Cluster.cs
```diff
@@ -474,7 +502,23 @@ public async Task ShutdownAsync(int timeoutMs = Timeout.Infinite)
             _metadata.ShutDown(timeoutMs);
             _controlConnection.Dispose();
             Configuration.Timer.Dispose();
-            Configuration.Policies.SpeculativeExecutionPolicy.Dispose();
+            
+            var disposedPolicies = new List<object>();
+            foreach (var options in Configuration.RequestOptions.Values)
```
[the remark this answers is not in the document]
joao-r-reis: > 
> 
> When I saw that execution profiles for `Prepare()` method was being introduced, I thought it was going to be the "default profile" for the bound statements created from the prepared statement, but I see that the profile at `Prepare()` level affects how the prepare statement is executed.
> 
> I think we shouldn't expose many options on how an prepare request is executed, the user should only care about whether it's prepared or it failed and the rest should be defined by the driver (there are some settings like prepare on all nodes, and those are set cluster/session-wide).
> 
> Query preparation should take a few cycles on the server and can be blindly retried on the following node.

What made me go for this was seeing the load balancing policy being called. I'm concerned mostly about `Mapper` and `LINQ` where users don't actually do any calls to `Prepare` and they probably don't expect the default `LoadBalancingPolicy` being used in a `Mapper` / `LINQ` call if they specify a different one with an execution profile. Maybe we should make these methods internal and be used only in `LINQ` and `Mapper`?

joao-r-reis:
> 
> 
> When I saw that execution profiles for `Prepare()` method was being introduced, I thought it was going to be the "default profile" for the bound statements created from the prepared statement, but I see that the profile at `Prepare()` level affects how the prepare statement is executed.
> 
> I think we shouldn't expose many options on how an prepare request is executed, the user should only care about whether it's prepared or it failed and the rest should be defined by the driver (there are some settings like prepare on all nodes, and those are set cluster/session-wide).
> 
> Query preparation should take a few cycles on the server and can be blindly retried on the following node.

What made me go for this was seeing the load balancing policy being called. I'm concerned mostly about `Mapper` and `LINQ` where users don't actually do any calls to `Prepare` and they probably don't expect the default `LoadBalancingPolicy` being used in a `Mapper` / `LINQ` call if they specify a different one with an execution profile. Maybe we should make these methods internal and be used only in `LINQ` and `Mapper`?

src/Cassandra/Cluster.cs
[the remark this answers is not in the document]
joao-r-reis: That suggestion also goes for `Init` right? (although it has to be `HashSet<object>` there)

jorgebay:
> Maybe we should make these methods internal and be used only in LINQ and Mapper?

I think we can live without it:

- When "prepare on all nodes" is enabled (default), the driver should use all nodes that we actually have an open connection to, instead of a LBP, and prepare it (so all nodes that matter should already have prepared the query).
- When disabled, the flow -> `EXECUTE` -> `UNPREPARED` -> (re) `PREPARE` -> `EXECUTE` will kick in (once per node that hasn't prepared the query)

joao-r-reis:
> 
> 
> > Maybe we should make these methods internal and be used only in LINQ and Mapper?
> 
> I think we can live without it:
> 
>     * When "prepare on all nodes" is enabled (default), the driver should use all nodes that we actually have an open connection to, instead of a LBP, and prepare it (so all nodes that matter should already have prepared the query).
> 
>     * When disabled, the flow -> `EXECUTE` -> `UNPREPARED` -> (re) `PREPARE` -> `EXECUTE` will kick in (once per node that hasn't prepared the query)

I've pushed a draft implementation of this but haven't tested it yet. Lmk if this is what you were thinking of: https://github.com/datastax/csharp-driver/pull/442/commits/16dfe484dd5f069eb517fe47465048038e83f749

jorgebay:
yes! that approach lgtm.

src/Cassandra/Cluster.cs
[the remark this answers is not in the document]
joao-r-reis: Changed this in both `Init` and `Shutdown`. I took your suggestion of using `HashSet` but also implemented a `ReferenceEqualityComparer` to make sure it is doing comparisons based on reference only.

joao-r-reis:
Pushed final changes to `Session.Prepare` API and `PrepareHandler` class. Also added some unit tests.

src/Cassandra/Cluster.cs
```diff
@@ -488,5 +522,77 @@ internal static HostDistance RetrieveDistance(Host host, ILoadBalancingPolicy lb
...
+                               "affect performance. Consider preparing the statement only once. Query='{0}'", ps.Cql);
+                ps = psAdded;
+            }
+
+            return ps;
+        }
+        
+        private async Task PrepareAllQueries(Host host)
+        {
+            ICollection<PreparedStatement> preparedQueries = InternalRef.PreparedQueries.Values;
```
jorgebay: Now that I think about it, `ConcurrentDictionary<T>.Values` get property is an expensive operation (acquires all locks).

Outside the scope of this ticket, we should use a collection for prepared queries where reading is cheap.

src/Cassandra/ICluster.cs
```diff
@@ -98,5 +98,10 @@ public interface ICluster : IDisposable
         ///  used by it. <p> This method has no effect if the cluster was already shutdown.</p>
         /// </summary>
         void Shutdown(int timeoutMs = System.Threading.Timeout.Infinite);
+
+        /// <summary>
+        /// Gets a collection of execution profiles that were configured for this Cluster instance.
+        /// </summary>
+        IReadOnlyDictionary<string, IExecutionProfile> ExecutionProfiles { get; }
```
jorgebay: Shouldn't we expose it in the `Configuration` and avoid exposing a new prop on `Cluster` / `Session`, wdyt?

src/Cassandra/Connections/IConnection.cs
```diff
@@ -71,7 +70,7 @@ internal interface IConnection : IDisposable
 
         ProtocolOptions Options { get; }
 
-        Configuration Configuration { get; set; }
+        //Configuration Configuration { get; set; }
```
jorgebay: Probably a leftover.

joao-r-reis: nice catch! done

src/Cassandra/ICluster.cs
joao-r-reis: I honestly don't remember why I didn't just do that considering that the `Cluster` property is just a wrapper around the `Configuration` one. Done!

src/Cassandra/Cluster.cs
joao-r-reis: created https://datastax-oss.atlassian.net/browse/CSHARP-768

PR #443, opened 2019-04-16
MappingConfiguration add a method get TypeDefinition

SarcoZ:
/// <summary>
        /// If defined, returns the mapping for POCO type T, otherwise returns null.
        /// </summary>
        public ITypeDefinition Get<T>()
        {
            ITypeDefinition existingMapping;
            _typeDefinitions.TryGetItem(typeof(T), out existingMapping);

            return existingMapping;
        }

above is ok,hope add a method to check a type has ITypeDefinition

public ITypeDefinition Get(Type type)
        {
            ITypeDefinition existingMapping;
            _typeDefinitions.TryGetItem(type, out existingMapping);

            return existingMapping;
        }

then 

/// <summary>
        /// If defined, returns the mapping for POCO type T, otherwise returns null.
        /// </summary>
        public ITypeDefinition Get<T>()
        {
            return Get(typeof(T));
        }

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -68,7 +68,7 @@ build:
           msbuild /p:Configuration=Release /v:m /p:DynamicConstants=LINUX src/Cassandra.sln
 
           # Run the tests
-          mono ./testrunner/NUnit.ConsoleRunner.3.6.1/tools/nunit3-console.exe src/Cassandra.IntegrationTests/bin/Release/net452/Cassandra.IntegrationTests.dll --where:cat=short --labels=All --result:"TestResult_nunit3.xml" || error=true
+          mono ./testrunner/NUnit.ConsoleRunner.3.6.1/tools/nunit3-console.exe src/Cassandra.IntegrationTests/bin/Release/net452/Cassandra.IntegrationTests.dll --where "cat != long && cat != memory" --labels=All --result:"TestResult_nunit3.xml" || error=true
           
           java -jar saxon/saxon9he.jar -o:TestResult.xml TestResult_nunit3.xml tools/nunit3-xunit.xslt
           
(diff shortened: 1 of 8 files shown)


joao-r-reis:
I'm very confused, you opened a PR to merge one of my commits from `master` to `3.9` that relates to a bug fix and in the description you talk about adding a method to `MappingConfiguration`. Can you double check if you opened this correctly? 

Also can you provide of a valid use case behind the method you want to add?

SarcoZ:
@joao-r-reis   I‘m sorry for that,this project open issue puzzled me, so it leads to misoperation.sorry again,

public class CassandraModelBuilder : ICassandraModelBuilder
    {
        private readonly Dictionary<Type, object> _entityModelBuilders;

        private static readonly object _syncObj = new object();

        public CassandraModelBuilder()
        {
            _entityModelBuilders = new Dictionary<Type, object>();
        }

        public CassandraContextModel Build()
        {
            var entityModels = _entityModelBuilders
                .Select(x => x.Value)
                .Cast<ICassandraEntityModel>()
                .ToImmutableDictionary(x => x.EntityType, x => x);

            foreach (var entityModel in entityModels.Values)
            {
                var map = entityModel.As<IHasTypeDefinition>().GetTypeDefinition();
                lock (_syncObj)
                {                   
                    MappingConfiguration.Global.Define(map);                  
                }
            }

            return new CassandraContextModel(entityModels);
        }

in my case ,I want which given a type that want to check if has already defined to avoid repeat defined,
thanks

joao-r-reis:
Oh so you intended to open an issue not a PR, right? We don't use github issues in this project, we use JIRA: https://datastax-oss.atlassian.net/projects/CSHARP

Are you familiar with JIRA? If you aren't I can create the issue for you.

joao-r-reis:
Ok,  closing this then.

PR #448, opened 2019-04-29
CSHARP-766 LocalDate bug fix

joao-r-reis:
For certain dates, the `LocalDate` `ctor` computes the wrong date from the given timestamp (days since epoch centered in `2^31`). Added these scenarios to the existing unit tests and replaced the implementation with a port from http://howardhinnant.github.io/date_algorithms.html#civil_from_days

### DIFF
--- a/src/Cassandra/LocalDate.cs
+++ b/src/Cassandra/LocalDate.cs
@@ -18,6 +18,8 @@ namespace Cassandra
         // unix epoch is represented by the number  2 ^ 31
         private const long DateCenter = 2147483648L;
         private const long DaysFromYear0ToUnixEpoch = 719528L;
+        // ReSharper disable once InconsistentNaming
+        private const long DaysFromYear0March1stToUnixEpoch = LocalDate.DaysFromYear0ToUnixEpoch - 60L;
         
         private static readonly Regex RegexInteger = new Regex("^-?\\d+$", RegexOptions.Compiled);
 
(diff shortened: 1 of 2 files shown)


src/Cassandra/LocalDate.cs
```diff
@@ -45,47 +42,29 @@ public class LocalDate: IComparable<LocalDate>, IEquatable<LocalDate>
...
+
+        /// <summary>
+        /// Port from http://howardhinnant.github.io/date_algorithms.html#civil_from_days
+        /// There's a good explanation of the algorithm over there.
+        /// </summary>
+        /// <param name="daysSinceEpoch">Days since Epoch (January 1st, 1970), can be negative.</param>
+        private void InitializeFromDaysSinceEpoch(long daysSinceEpoch)
+        {
+            var z = daysSinceEpoch;
+            z += 719468L;
```
jorgebay: We can name these constants, like `daysSinceYearZeroToUnixEpoch`

joao-r-reis:
Your comment disappeared somehow @jorgebay so writing here: 

I didn't name that one exactly like that but added a comment explaining that particular computation.
Also added names and some comments to this method's code.

src/Cassandra/LocalDate.cs
```diff
@@ -45,47 +42,31 @@ public class LocalDate: IComparable<LocalDate>, IEquatable<LocalDate>
...
+        }
+
+        /// <summary>
+        /// Port from https://github.com/HowardHinnant/date which is based on http://howardhinnant.github.io/date_algorithms.html
+        /// There's a good explanation of the algorithm over there.
+        /// </summary>
+        /// <param name="daysSinceEpoch">Days since Epoch (January 1st, 1970), can be negative.</param>
+        private void InitializeFromDaysSinceEpoch(long daysSinceEpoch)
+        {
+            daysSinceEpoch += 719468L;                                                                  // shift epoch from 1970-01-01 to 0000-03-01
```
jorgebay: This constant is defined above: `DaysFromYear0ToUnixEpoch`.

src/Cassandra/LocalDate.cs
```diff
@@ -45,47 +42,31 @@ public class LocalDate: IComparable<LocalDate>, IEquatable<LocalDate>
...
+            var era = (daysSinceEpoch >= 0 ? daysSinceEpoch : daysSinceEpoch - 146096) / 146097;        // compute era
```
jorgebay: We could explain in code comment that in this algorithm, an era is a 400 year period.

src/Cassandra/LocalDate.cs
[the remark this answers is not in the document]
joao-r-reis: This is the same comment as https://github.com/datastax/csharp-driver/pull/448#discussion_r281965202 right?

src/Cassandra/LocalDate.cs
joao-r-reis: The constant is actually different, not sure why...

joao-r-reis: Ah it's because the year starts on March 1st here, nevermind. I'll change it to use the constand and remove 60 days

src/Cassandra/LocalDate.cs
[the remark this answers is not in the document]
jorgebay: yes, sorry, it was on my review buffer.

src/Cassandra/LocalDate.cs
joao-r-reis: done

src/Cassandra/LocalDate.cs
joao-r-reis: Ok created new constant `DaysFromYear0March1stToUnixEpoch` which is `DaysFromYear0ToUnixEpoch - 60L` and used that one in the method.

PR #452, opened 2019-05-19
CSHARP-762 Document next release's features

joao-r-reis:
- Execution profiles
- Metadata synchronization toggle
- UDT definitions with keyspace


doc/features/execution-profiles/README.md
```diff
@@ -0,0 +1,160 @@
...
+```csharp
+var cluster = 
+   Cluster.Builder()
+          .AddContactPoint("127.0.0.1")
+          .WithExecutionProfiles(opts => opts
+            .WithProfile("default", profile => profile
+                .WithLoadBalancingPolicy(new TokenAwarePolicy(new DCAwareRoundRobinPolicy(localDc: "dc1")))
+            .WithProfile("login", profile => profile
+                .WithConsistencyLevel(ConsistencyLevel.LocalOne)
+                .WithReadTimeoutMillis(100)
```
jorgebay: I would avoid using read timeout in the example as it's too low level and it shouldn't be used to enforce SLAs (maybe when we introduce "per execution timeout" in the next major, it could make sense).

I think we should point users to spec execs instead.

doc/features/execution-profiles/README.md
```diff
@@ -0,0 +1,160 @@
...
+var ps = await session.PrepareAsync("SELECT * FROM users WHERE username = ?").ConfigureAwait(false);
+
+// on request
+var statement = ps.Bind(username);
+var fetchResult = await session.ExecuteAsync(statement, "login").ConfigureAwait(false);
+```
+
+## Mapping Legacy Parameters to Profiles
+
+The name `default` is reserved for the default execution profile. This profile will be the one that is going to be used whenever no profile is specified in a request.
```
jorgebay: Good level of detail :)

doc/features/execution-profiles/README.md
joao-r-reis: Ok changed the example so that both profiles differ in the `ISpeculativeExecutionPolicy` parameter instead of `ReadTimeoutMillis`.

jorgebay: There are still references to read timeout below :)

joao-r-reis: Ah, I really went ham with `ReadTimeoutMillis` on these examples, 😄  fixed

abhiapuri:
Hi Jorge ,
Does version 3.0.9 support CQL3?
Thank You
Abhilash

jorgebay:
I see you posted the question on the mailing list, we can continue discussing from there.

PR #455, opened 2019-05-31
CSHARP-772 removed unnecessary timestamp converting from unix microseconds

levdimov:
Converting prevents from using .NET ticks as described in [CSHARP-772](https://datastax-oss.atlassian.net/browse/CSHARP-772).

I analyzed every external using of the `QueryProtocolOptions.Timestamp` property and made sure it was used just for nullability checks. I also analyzed `QueryProtocolOptions` and found no usings of class in public API which tells us that such change would not be a "braking" one.

### DIFF
--- a/src/Cassandra/QueryProtocolOptions.cs
+++ b/src/Cassandra/QueryProtocolOptions.cs
@@ -43,21 +43,11 @@ namespace Cassandra
         public readonly int PageSize;
         public readonly ConsistencyLevel SerialConsistency;
 
-        private readonly long? _timestamp;
-
         public byte[] PagingState { get; set; }
         public object[] Values { get; private set; }
         public ConsistencyLevel Consistency { get; set; }
 
-        public DateTimeOffset? Timestamp
-        {
-            get
-            {
-                return _timestamp == null ? (DateTimeOffset?) null :
-                    TypeSerializer.UnixStart.AddTicks(_timestamp.Value * 10);
-            }
-        }
-
+        public long? Timestamp { get; }
 
         /// <summary>
         /// Names of the query parameters
@@ -89,7 +79,7 @@ namespace Cassandra
             }
             PagingState = pagingState;
             SerialConsistency = serialConsistency;
-            _timestamp = timestamp;
+            Timestamp = timestamp;
         }
 
         internal static QueryProtocolOptions CreateFromQuery(
@@ -157,7 +147,7 @@ namespace Cassandra
             {
                 flags |= QueryFlags.WithSerialConsistency;
             }
-            if (protocolVersion.SupportsTimestamp() && _timestamp != null)
+            if (protocolVersion.SupportsTimestamp() && Timestamp != null)
             {
                 flags |= QueryFlags.WithDefaultTimestamp;
             }
@@ -223,7 +213,7 @@ namespace Cassandra
             {
                 // ReSharper disable once PossibleInvalidOperationException
                 // Null check has been done when setting the flag
-                wb.WriteLong(_timestamp.Value);
+                wb.WriteLong(Timestamp.Value);
             }
         }
     }
--- a/src/Cassandra/QueryProtocolOptions.cs
+++ b/src/Cassandra/QueryProtocolOptions.cs
@@ -47,7 +47,16 @@ namespace Cassandra
         public object[] Values { get; private set; }
         public ConsistencyLevel Consistency { get; set; }
 
-        public long? Timestamp { get; }
+        public DateTimeOffset? Timestamp
+        {
+            get
+            {
+                return RawTimestamp == null ? (DateTimeOffset?) null :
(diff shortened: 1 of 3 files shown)


src/Cassandra/QueryProtocolOptions.cs
```diff
@@ -43,21 +43,11 @@ public enum QueryFlags
...
-        public DateTimeOffset? Timestamp
-        {
-            get
-            {
-                return _timestamp == null ? (DateTimeOffset?) null :
-                    TypeSerializer.UnixStart.AddTicks(_timestamp.Value * 10);
-            }
-        }
-
+        public long? Timestamp { get; }
```
joao-r-reis: This is a breaking change. `QueryProtocolOptions` is a `public class` and this property is `public` so users might be using this property in their application code. 

The best way to go forward for now (without a major release) is to just make sure that this property is not used internally by the driver. I suggest adding an `internal` property called `RawTimestamp` or something similar that just returns the `_timestamp` field or we could make the new property an auto property and remove the field.

src/Cassandra/QueryProtocolOptions.cs
```diff
@@ -89,7 +79,7 @@ public enum QueryFlags
             }
             PagingState = pagingState;
             SerialConsistency = serialConsistency;
-            _timestamp = timestamp;
+            Timestamp = timestamp;
```
joao-r-reis: Due to my comment above, this needs to be changed after creating the new property (either set the field here or the auto property).

src/Cassandra/QueryProtocolOptions.cs
```diff
@@ -157,7 +147,7 @@ private QueryFlags GetFlags(ProtocolVersion protocolVersion)
             {
                 flags |= QueryFlags.WithSerialConsistency;
             }
-            if (protocolVersion.SupportsTimestamp() && _timestamp != null)
+            if (protocolVersion.SupportsTimestamp() && Timestamp != null)
```
joao-r-reis: Same as above.

src/Cassandra/QueryProtocolOptions.cs
```diff
@@ -223,7 +213,7 @@ internal void Write(FrameWriter wb, bool isPrepared)
             {
                 // ReSharper disable once PossibleInvalidOperationException
                 // Null check has been done when setting the flag
-                wb.WriteLong(_timestamp.Value);
+                wb.WriteLong(Timestamp.Value);
```
joao-r-reis: Same as above.

levdimov:
> This is a breaking change. QueryProtocolOptions is a public class and this property is public so users might be using this property in their application code.

I never developed any public library so please bear with my silly questions :-)

I can't see any using of QueryProtocolOptions in the public API (neither public interface nor public class methods & only in internals) even with help of ReSharper and full-text search. If it is really so, then does it make any sense for a user to use such class in their code? 

Either way I can agree with your review comments and fix the PR, but I would really like to hear your comments about the previous paragraph. Thank you!

joao-r-reis:
> I never developed any public library so please bear with my silly questions :-)

We're here to help 😄 

> I can't see any using of QueryProtocolOptions in the public API (neither public interface nor public class methods & only in internals) even with help of ReSharper and full-text search. If it is really so, then does it make any sense for a user to use such class in their code?
> 
> Either way I can agree with your review comments and fix the PR, but I would really like to hear your comments about the previous paragraph. Thank you!

You're absolutely right that it doesn't make any sense for a user to use this class in their code. One very weird use case that comes to mind is a user implementing custom `IStatement` objects and using the `QueryProtocolOptions` class with them. We don't intend to support this and [we've clarified recently our stance on breaking changes](https://docs.datastax.com/en/developer/csharp-driver/3.10/upgrade-guide/#our-policy-regarding-api-changes-and-release-versioning). 

Having said that, we still try to avoid any kind of breaking change to `public` classes, interfaces and enums unless the alternative is much worse to maintain and use. If this was a very problematic bug with a lot of potential to break stuff then I would be ok with it for the reasons you mentioned. But since the property has been there for a very very long time without any issue, I don't think there's any urgency to change it before a major release.

levdimov:
> ...custom IStatement objects and using the QueryProtocolOptions class with them

This is weird but I got your point. Very grateful for your explanation.

I used your suggestions and added a new property instead.

joao-r-reis:
Looks good after a very quick review, I'll take a few minutes tomorrow to download the branch and do a final pass before merging it in. Again, thanks for the contribution :+1:

Just fyi, the next release date is still TBD as we've just released `3.10` less than two weeks ago.

joao-r-reis:
I got a bit uncomfortable with the fact that there wasn't any test to verify this so I just added a simple integration test to the `master` branch. If everything goes well in CI, I'll merge this in.

joao-r-reis:
Confirmed that the new test is failing in `master` and is passing on this PR! Thanks!

levdimov:
That is wonderful! Thank you, @joao-r-reis!

PR #456, opened 2019-06-04
Add new examples folder and concurrent execution example

joao-r-reis:
This PR adds an example that shows how someone can limit the number of concurrent requests with the driver.

Deleted the old `src/samples` folder in favor of the new `examples` folder in the root of the repo (which is more consistent with the other drivers). There was a sample project in the old folder but it was very outdated and didn't load properly so I didn't bother moving it to the new folder, it's just gone.


examples/ConcurrentExecutions/ExecuteInLoop/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            // build cluster
+            _cluster =
+                Cluster.Builder()
+                    .AddContactPoint("127.0.0.1")
+                    .WithLoadBalancingPolicy(new TokenAwarePolicy(new DCAwareRoundRobinPolicy("datacenter1")))
+                    .Build();
+
+            // create session
+            _session = _cluster.Connect();
```
jorgebay: NIT: we can use `ConnectAsync()`.

joao-r-reis: Done 😄

jorgebay:
In the future, we could compile the examples solution as part of the .NET Core 2.1+ build.

joao-r-reis:
Oh that's a very cool idea, I'm thinking we should add this to the appveyor and travis netcoreapp2.1 unit test runs, putting the examples build command before running `dotnet test`, right?

jorgebay:
yes, something like that, compiling them in one CI provider is good enough :)

joao-r-reis:
Added the build command for the new `examples` solution in `travis.yml`. Had to add a condition property to the driver's csproj files so that we can call `dotnet build examples.sln` for dotnet core only. I tried using `-f netcoreapp2.1` but then it complained that the `Cassandra.csproj` only had `netstandard` targets instead of `netcoreapp` so this was the only solution (without having to build every single one by one). 

Also noticed that travis runs were taking almost 1 hour to finish. While troubleshooting this, I updated the nuget version for test related stuff but in the end it turned out to be this: https://github.com/dotnet/core/issues/2775
It went back to normal after pinpointing the `dotnet` sdk version on `travis.yml`.

jorgebay:
Nice findings!

.travis.yml
```diff
@@ -13,9 +13,20 @@ branches:
...
+    - env:
+        - TARGETF="netcoreapp2.0"
+        - BUILD_EXAMPLES="0"
 
 script:
+  - |
+    if [ "${BUILD_EXAMPLES}" == "1" ]
+    then
+        dotnet restore examples
+        dotnet build --no-restore examples/examples.sln -c Release /p:BuildCoreOnly=True
```
jorgebay: NIT: We could use environment variable instead of passing `/p:ENV_VAR=val`

joao-r-reis: oh nice tip, didn't know that, done!

PR #457, opened 2019-06-07
CSHARP-777: if there is an error writing a CQL frame, back up the str…

collinsauve:
…eam to before we started writing

See https://datastax-oss.atlassian.net/browse/CSHARP-777

### DIFF
--- a/src/Cassandra/Connections/Connection.cs
+++ b/src/Cassandra/Connections/Connection.cs
@@ -797,6 +797,7 @@ namespace Cassandra.Connections
                 }
                 _pendingOperations.AddOrUpdate(streamId, state, (k, oldValue) => state);
                 int frameLength;
+                var startLength = stream?.Length ?? 0; 
                 try
                 {
                     //lazy initialize the stream
@@ -816,6 +817,9 @@ namespace Cassandra.Connections
                     RemoveFromPending(streamId);
                     //Callback with the Exception
                     state.InvokeCallback(ex);
+
+                    //Reset the stream to before we started writing this frame
+                    stream?.SetLength(startLength);
                     break;
                 }
                 //We will not use the request any more, stop reference it.


src/Cassandra/Connections/Connection.cs
```diff
@@ -797,6 +797,7 @@ private void RunWriteQueueAction()
                 }
                 _pendingOperations.AddOrUpdate(streamId, state, (k, oldValue) => state);
                 int frameLength;
+                var startPosition = stream?.Position ?? 0;
```
joao-r-reis: Reading through `RecyclableMemoryStream.cs` it looks like it makes more sense to replace this with

`var startLength = stream?.Length ?? 0;`

@jorgebay do you agree?

src/Cassandra/Connections/Connection.cs
```diff
@@ -816,6 +817,8 @@ private void RunWriteQueueAction()
                     RemoveFromPending(streamId);
                     //Callback with the Exception
                     state.InvokeCallback(ex);
+
+                    stream?.SetLength(startPosition);
                     break;
```
joao-r-reis: @jorgebay can you think of any reason for this `break` being here? Shouldn't the driver keep writing "valid" requests to the stream and skip the "invalid" ones?

src/Cassandra/Connections/Connection.cs
jorgebay: In this case is the same, I'm ok with `Position` (equivalent to [writer index](https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#writerIndex--) in Netty).

src/Cassandra/Connections/Connection.cs
jorgebay: hm... good question...
It avoids further coalescing of the message and the valid messages will continue to be written (`socket.Write()` and `RunWriteQueue()` calls below), but I understand that it's not needed.

It doesn't have any real impact (outside having smaller messages when there is a error building an individual request) so we can evaluate in future revisions whether we want to take it out.

src/Cassandra/Connections/Connection.cs
jorgebay: hm... now that I think about it, as below we are setting the `Length` (that sets `Position` internally) it makes sense to use 1 property: `Length`.

joao-r-reis: :+1: @collinsauve can you change `stream?.Position` to `stream?.Length` and rename the `startPosition` variable accordingly? After that change and CI run is done I'll merge this.

src/Cassandra/Connections/Connection.cs
collinsauve: done

src/Cassandra/Connections/Connection.cs
collinsauve: done

src/Cassandra/Connections/Connection.cs
collinsauve: Just make sure if this `break` is removed that it doesn't fall through to:

```
state.Request = null;
totalLength += frameLength;
```

we don't want to incr `totalLength`

jorgebay: yes, I don't think its a good idea to change the existent logic in the context of this bug fix.

PR #463, opened 2019-08-29
Fixing CSHARP-802

infraweavers:
As suggested in https://datastax-oss.atlassian.net/browse/CSHARP-802 adding `OnConnectionClosing` into the `catch` makes the driver regard a node that is down as down earlier on and we find that timeout exceptions are no longer thrown when calling `Prepare`. We've tested this by unplugging a node and running a test application several times and it seems reliable with this change in.

For reference, our test cluster was built using:
```
return Cluster.Builder()
	.WithLoadBalancingPolicy(new TokenAwarePolicy(new DCAwareRoundRobinPolicy("TESTDC")))
	.AddContactPoints("cass1,cass2,cass3,cass4,".Split(','))
	.WithQueryTimeout(5000)
	.Build();
```

### DIFF
--- a/src/Cassandra/Connections/HostConnectionPool.cs
+++ b/src/Cassandra/Connections/HostConnectionPool.cs
@@ -829,6 +829,7 @@ namespace Cassandra.Connections
                         break;
                     }
 
+                    OnConnectionClosing();
                     throw;
                 }
             }


infraweavers:
@joao-r-reis Thanks, none of changes to the timeout values seemed to help. The workaround of using `PrepareAsync` and setting a timeout ourselves which is greater than the cluster defined `QueryTimeout` does work.

joao-r-reis:
Ok thanks for the information. Just a quick correction, `QueryTimeout` is not used in async methods like `PrepareAsync` so you shouldn't have to change `QueryTimeout` if you use `PrepareAsync` instead of `Prepare`

PR #464, opened 2019-09-02
CSHARP-799 Add category for integration tests with real clusters and optimize the test execution on appveyor

joao-r-reis:


src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
```diff
@@ -20,34 +20,46 @@ public ClientTimeoutTests()
...
             Diagnostics.CassandraStackTraceIncluded = true;    
         }
-
+        
         [Test]
         public void Should_Move_To_Next_Host_For_Simple_Queries()
         {
-            var testCluster = TestClusterManager.GetNonShareableTestCluster(2, 1, true, false);
-            var socketOptions = new SocketOptions().SetReadTimeoutMillis(3000);
+            var testCluster = SimulacronCluster.CreateNew(2);
```
jorgebay: Nice time saves in this fixture!

src/Cassandra.IntegrationTests/Ado/AdoBasicTests.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace Cassandra.IntegrationTests.Data
 {
-    [TestFixture, Category("short")]
+    [TestFixture, Category("short"), Category("realcluster")]
```
jorgebay: I really like the idea to have a category where tests can share a cluster.

We could use a category name that carries more meaning than "realcluster"? Some suggestions:
- minimal
- basic
- singlecluster

src/Cassandra.IntegrationTests/SharedSimulacronClusterTest.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace Cassandra.IntegrationTests
+{
+    /// <summary>
+    /// Represents a test fixture that on setup, it creates a test cluster available for all tests.
+    /// <para>
+    /// With a shared session and cluster.
+    /// </para>
+    /// </summary>
+    [TestFixture]
+    public class SharedSimulacronClusterTest
```
jorgebay: I think its a leftover.

src/Cassandra.IntegrationTests/Ado/AdoBasicTests.cs
jorgebay: Its hard to name categories that serve every purpose and `realcluster` allows us to mark those that require C* clusters.
I was thinking about `minimal`, as a subgroup of `short` that would be for the 5 min runs in appveyor and also allows us to smoke test new versions of the server, but there isn't much overlap (simulacron doesn't test the server 😃 )

Nevermind, sorry for the misunderstanding.

src/Cassandra.IntegrationTests/SharedSimulacronClusterTest.cs
joao-r-reis: removed

src/Cassandra.IntegrationTests/Ado/AdoBasicTests.cs
joao-r-reis: Sounds good!

PR #467, opened 2019-10-21
CSHARP-595 Log warning when local dc is not specified

joao-r-reis:
Also added code example to `Tuning policies` manual page and added a new utility method
to help users create instances of the default LBP with a local dc.

### DIFF
--- a/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
+++ b/src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
@@ -37,6 +37,8 @@ namespace Cassandra
             "The usedHostsPerRemoteDc parameter will be removed in the next major release of the driver. " +
             "DC failover should not be done in the driver, which does not have the necessary context to know " +
             "what makes sense considering application semantics. See https://datastax-oss.atlassian.net/browse/CSHARP-722";
+        
+        private static readonly Logger Logger = new Logger(typeof(DCAwareRoundRobinPolicy));
 
         private string _localDc;
         private readonly int _usedHostsPerRemoteDc;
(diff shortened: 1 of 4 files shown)


src/Cassandra/Policies/Policies.cs
```diff
@@ -35,6 +40,18 @@ public static ILoadBalancingPolicy DefaultLoadBalancingPolicy
...
         }
 
+        /// <summary>
+        /// Creates a new instance of the default load balancing policy with the provided local datacenter.
+        /// This is equivalent to:
+        /// <code>
+        /// new TokenAwarePolicy(new DCAwareRoundRobinPolicy(localDc))
+        /// </code>
+        /// </summary>
+        public static ILoadBalancingPolicy NewDefaultLoadBalancingPolicy(string localDc)
```
jorgebay: Having a forward-compatible method like this is a GREAT idea!

src/Cassandra/Policies/DCAwareRoundRobinPolicy.cs
```diff
@@ -117,25 +119,36 @@ public DCAwareRoundRobinPolicy(string localDc, int usedHostsPerRemoteDc)
         public void Initialize(ICluster cluster)
         {
             _cluster = cluster;
+
             //When the pool changes, it should clear the local cache
             _cluster.HostAdded += _ => ClearHosts();
             _cluster.HostRemoved += _ => ClearHosts();
+
+            var availableDcs = _cluster.AllHosts().Select(h => h.Datacenter).Where(dc => dc != null).Distinct().ToList();
```
jorgebay: NIT: iteration for `ToList()` can be avoided, as `string.Join()` can take an `IEnumerable<string>` instance.

joao-r-reis: Hmm I added `ToList` to avoid the multiple enumarion of the `IEnumerable` but I guess it doesn't make much sense, all it does is add one more iteration to the already existing two.

joao-r-reis: Actually I'd rather have the `ToList` just to make sure that the enumeration occurs only once.

PR #469, opened 2019-10-24
CSHARP-754 DataStax Apollo support

joao-r-reis:


build.yaml
```diff
@@ -45,7 +45,19 @@ build:
 
       echo "========== Copying ssl files to $HOME/ssl =========="
       cp -r /home/jenkins/ccm/ssl $HOME/ssl
-
+      
+      export SNI_ENABLED="true"
```
jorgebay: NIT: given that SNI project uses a single DSE version, we could build it on a single dimension of the matrix, e.g., build it only on `'3.11'` jobs

joao-r-reis: done, already checked in jenkins that it is correctly skipping the cloud tests for targets other than `3.11` do you agree with the way I implemented this in `build.yaml`? It's not very pretty but I didn't know any other way to do this

.travis.yml
```diff
@@ -2,23 +2,23 @@ language: csharp
...
     - dse_3.1
-    
-env:
-  global:
-    - BuildCoreOnly=True
-  matrix:
-    - TARGETF="netcoreapp2.1" BUILD_EXAMPLES="1"
-    - TARGETF="netcoreapp2.0" BUILD_EXAMPLES="0"
+
+matrix:
```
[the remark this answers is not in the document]
joao-r-reis: Nice! I didn't know about that, I'll look into it after this release

build.yaml
jorgebay: yeah, looks good. We're already overriding `CASSANDRA_VERSION` so makes a code a little more complicated but its OK :)

PR #470, opened 2019-10-28
Connected nodes metric bug fix and other small improvements

joao-r-reis:
Already verified in a small duration test that the gauge value updates when a node is taken down.

### DIFF
--- a/doc/features/metrics/app-metrics/README.md
+++ b/doc/features/metrics/app-metrics/README.md
@@ -65,14 +65,14 @@ var cluster = Cluster
     .WithSocketOptions(new SocketOptions().SetReadTimeoutMillis(10000))
     .WithMetrics(metricsRoot.CreateDriverMetricsProvider(
         new DriverAppMetricsOptions()
-            .SetHighestLatencyMilliseconds(12000) // should be set to slightly higher than the configured timeout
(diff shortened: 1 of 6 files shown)


joao-r-reis:
Added a commit to change the default threshold latency in `Timer` metrics and improve the RE-PREPARE error log message.

doc/features/metrics/app-metrics/README.md
```diff
@@ -65,14 +65,14 @@ var cluster = Cluster
     .WithSocketOptions(new SocketOptions().SetReadTimeoutMillis(10000))
     .WithMetrics(metricsRoot.CreateDriverMetricsProvider(
         new DriverAppMetricsOptions()
-            .SetHighestLatencyMilliseconds(12000) // should be set to slightly higher than the configured timeout
+            .SetHighestLatencyMilliseconds(15000) // should be set to a value that is higher than the configured timeout
```
jorgebay: Probably meant `30000` :)

joao-r-reis: Hmm this is just an example on the manual, you think we should use the default here? I mention a bit later that the default is `30000`

joao-r-reis: Yeah I think it's better actually since we were having log messages when the value wasn't high enough, I'll change it

PR #475, opened 2019-11-25
fix BatchStatement error message

panuoksala:
Code is executing check against **ushort** max value, but reporting max value from **short** datatype.

This is current error message:
_'There can be only 32767 child statement in a batch statement accordung to the cassandra native protocol')_


joao-r-reis:
Hi, nice catch! I opened https://datastax-oss.atlassian.net/browse/CSHARP-835 to track this.

Thanks for the contribution.

PR #476, opened 2019-12-02
CSHARP-828 Add test with a host that changes IP

joao-r-reis:
Also refactored a test to use simulacron instead of ccm to compensate the added execution time introduced by the new test.

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ReconnectionTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ReconnectionTests.cs
@@ -35,12 +35,20 @@ namespace Cassandra.IntegrationTests.Core
     public class ReconnectionTests : TestGlobals
     {
         private SimulacronCluster _testCluster;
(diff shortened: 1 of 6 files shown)


jorgebay:
There's a failure that occurs on the new test under mono on Jenkins:`ProtocolErrorException` coming from the server.

The stack trace is not helpful and I think it will be very hard to reproduce. I've triggered a few more builds, let's see if we can gather more info on the underlying cause.

joao-r-reis:
That failure was related to a bug that was fixed by #477 .

This failure was happening because we are targeting C* 3.11.5 and with [CASSANDRA-15193](https://issues.apache.org/jira/browse/CASSANDRA-15193) the node will reply that it doesn't support v4 until it receives gossip information that every node is running C* 3.0+. Our usage of `--skip-wait-other-notice` might be causing this race condition to occur but in any case it shouldn't break anything even if the server reports that it doesn't support v4 (the driver should just ignore it and carry on).

Rebased this branch after #477 merge, ran the test 100 times and it didn't fail.

PR #477, opened 2019-12-04
CSHARP-837 Fix protocol version bug when a node is added with a lower version

joao-r-reis:
When a node is bootstrapped and it doesn't support the current protocol version, the connection will change the version for all active connections. This PR makes it so that the serializer instances on each connection are tied to a specific version and it also makes it so that the control connection will not change the protocol version after init.


src/Cassandra/Connections/Connection.cs
```diff
@@ -536,18 +537,21 @@ internal bool ReadParse(byte[] buffer, int length)
 
             // Check if protocol version has already been determined (first message)
             ProtocolVersion protocolVersion;
-            var headerLength = Volatile.Read(ref _frameHeaderSize);
+            var headerLength = _frameHeaderSize;
```
jorgebay: hm... I think we still need the volatile read. The memory barrier below prevents reordering on that case, but `_frameHeaderSize` may still hold an old value for a given thread.

src/Cassandra/Serialization/SerializerManager.cs
```diff
@@ -0,0 +1,72 @@
...
+            _genericSerializer = new GenericSerializer(typeSerializers);
+            _serializer = new Serializer(protocolVersion, _genericSerializer);
+        }
+
+        public ProtocolVersion CurrentProtocolVersion => _serializer.ProtocolVersion;
+
+        public void ChangeProtocolVersion(ProtocolVersion version)
+        {
+            _serializer = new Serializer(version, _genericSerializer);
+            Interlocked.MemoryBarrier();
```
jorgebay: Mostly the same, I think nothing is preventing `_serializer` reference to be out of date for a given thread.

We would need volatile write and reads if we allow multiple threads modifying and reading a field.

src/Cassandra/Connections/Connection.cs
joao-r-reis: Yeah I was trying to avoid adding more volatile operations to this path but I don't think we have a choice. Same for the `serializer` instance

joao-r-reis:
Added `volatile` to `_serializer` in `SerializerManager` and to `_serializer` in `Connection`. Also noticed that the way we were retrieving the pooling options was prone to concurrency issues as well so I changed it

src/Cassandra.Tests/ConnectionTests.cs
```diff
@@ -310,17 +310,56 @@ public async Task ReadParse_While_Disposing_Faults_Tasks_But_Never_Throws()
             await TestHelper.WaitUntilAsync(() => totalFrames == responses.Count, 100, 30).ConfigureAwait(false);
             Assert.AreEqual(totalFrames, responses.Count);
         }
+        
+        [Test]
+        public void Should_HandleDifferentProtocolVersionsInDifferentConnections_When_OneConnectionResponseVersionIsDifferentThanSerializer()
```
jorgebay: Good coverage!

joao-r-reis: Yeah I copy pasted this test into `master` and it fails so I guess it accomplishes its goal but the reason of the failure is a bit weird. Basically it parses the body length as a ridiculously large number due to the header size being different than expected and the `Task` of the request is never set as `Complete` because of that. 

This makes me wonder if we should make sure that we respect the version that comes in the header even if we are past the `Startup` phase... Theoretically it should never happen but there's no reason not to use the version in the header to figure out the remaining length of the header we are parsing. This would allow us to properly recognize what's going on, log a warning and maybe set the result of the request as a `ProtocolError`? 

I'm going to merge this PR but we can continue discussing this here or offline. After the discussion I'll capture this into a ticket if we think it's worth it to do so.

PR #478, opened 2019-12-06
CSHARP-818 Add simulacron fluent api and move several ccm tests to simulacron

joao-r-reis:


src/Cassandra.IntegrationTests/Core/ClientTimeoutTests.cs
```diff
@@ -61,18 +63,11 @@ public void Should_Move_To_Next_Host_For_Simple_Queries()
...
-                        result = "success",
-                        delay_in_ms = 2000,
-                        rows = new[] { new { key = "123" } },
-                        column_types = new { key = "ascii" },
-                        ignore_on_prepare = false
-                    }
-                });
+                node.PrimeFluent()
+                    .WhenQuery("SELECT key FROM system.local")
+                    .ThenRowsSuccess(new [] { ("key", "ascii") }, rows => rows.WithRow("123"))
```
joao-r-reis: add fluent builder to setup column types

jorgebay:
37 mins on appveyor! what!? 💥 

NICE

joao-r-reis:
Almost all of the code changes are on the test suite code. The only changes to production code are the ones that affect 2 or 3 files and they are related to CQL generation:

- Create keyspace was genearting strings with weird line breaks
- `IF EXISTS` had an extra whitespace at the end
- Column names are now appended to CQL statement strings alphabetically (previously it depended on the order that `type.GetFields` would return them which [was by definition not deterministic](https://docs.microsoft.com/en-us/dotnet/api/system.type.getfields?view=netframework-4.8):
>The GetFields method does not return fields in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which fields are returned, because that order varies.

joao-r-reis:
I was trying to make the consoler test logger work but https://github.com/microsoft/vstest/issues/2120 still causes a mess and updating the test sdk nuget doesn't fix it, only changing the sdk version might work but then the tests stopped running for `netcoreapp2.0` so for now I'm keeping the `dotnet` sdk version the same that it was before on travis. On jenkins and appveyor we don't see any issue because we use other `loggers` like `xunit` and `Appveyor` (I think)

build/appveyor_install.ps1
```diff
@@ -133,7 +133,7 @@ If (!(Test-Path C:\Users\appveyor\.ccm\repository\$env:cassandra_version)) {
 #Download simulacron jar
 $simulacron_path = "$($dep_dir)\simulacron.jar"
 If (!(Test-Path $simulacron_path)) {
-  $simulacron_version = "0.8.2"
+  $simulacron_version = "0.8.11"
```
jorgebay: We can jump to 0.9.0, as we will need it soon.

src/Cassandra.IntegrationTests/Core/PreparedStatementsTests.cs
```diff
@@ -858,6 +860,65 @@ private void AssertValid(ISession session, PreparedStatement ps, object value)
             }
         }
 
+        [Test]
+        [TestCassandraVersion(2, 0)]
+        public void Batch_PreparedStatement_With_Unprepared_Flow()
```
jorgebay: Since most batch integration tests have been migrated to use simulacron, I was going to ask that we include some tests using a real ccm cluster, but you already thought of that :)

src/Cassandra/Data/Linq/SessionExtensions.cs
```diff
@@ -22,8 +22,8 @@ namespace Cassandra.Data.Linq
 {
     public static class SessionExtensions
     {
-        /// <summary>
-        /// <para>Extension method used for backward-compatibility, use <see cref="Cassandra.Data.Linq.Table{T}(Cassandra.ISession)"/> constructor instead.</para>
+        /// <summary> 
+        /// <para>Deprecated. Extension method used for backward-compatibility, use <see cref="Cassandra.Data.Linq.Table{T}(Cassandra.ISession)"/> constructor instead.</para>
```
jorgebay: +1 for the soft deprecation message.

src/Cassandra.IntegrationTests/SimulacronTest.cs
```diff
@@ -0,0 +1,206 @@
...
+using Cassandra.IntegrationTests.SimulacronAPI.Models.Logs;
+using Cassandra.IntegrationTests.SimulacronAPI.SystemTables;
+using Cassandra.IntegrationTests.TestClusterManagement;
+using Cassandra.IntegrationTests.TestClusterManagement.Simulacron;
+using NUnit.Framework;
+
+namespace Cassandra.IntegrationTests
+{
+    [TestFixture, Category("short")]
+    public class SimulacronTest
```
jorgebay: Good idea to have a clean base class for simulacron-based tests.

Shouldn't this be `abstract`?

src/Cassandra.IntegrationTests/Core/PreparedStatementsTests.cs
joao-r-reis: actually this was more due to a limitation with simulacron that prevented me from moving these tests but yeah I agree with you that we should have some real cluster tests  here

build/appveyor_install.ps1
joao-r-reis: done

src/Cassandra.IntegrationTests/SimulacronTest.cs
joao-r-reis: yep good catch, done

PR #479, opened 2020-01-06
CSHARP-843 Add google tag manager script

joao-r-reis:
In order to accomplish this I created a custom presentation style inside the repo and I copied the VS2013 presentation style files since we were using this style before.


jorgebay:
I think that this patch adds too much noise to the repo.

Can we wait until we migrate to docfx ([CSHARP-489](https://datastax-oss.atlassian.net/browse/CSHARP-489))?

joao-r-reis:
Oh I awasn't aware of that ticket. Should we migrate to docfx or doxygen? Doxygen is already supported by documentor

jorgebay:
I haven't tried doxygen with C# xmldocs, but if it works it can be an straightforward integration.
In any case, can the addition of this script be postponed until we revamp API docs generation?

joao-r-reis:
As discussed offline I'll close this PR and keep the branch so we use this branch to generate the api docs for the next release(s) before we move the api docs generation to docsfx or doxygen

PR #481, opened 2020-01-28
CSHARP-816 Set legacy settings from the default profile

joao-r-reis:


joao-r-reis:
Jenkins builds are failing due to saxon download. I will rerun the build once #480 is merged which fixes this.

build.yaml
```diff
@@ -92,7 +92,7 @@ build:
 
       # Download and uncompress saxon
       mkdir saxon
-      curl -L -o saxon/saxon9he.jar http://central.maven.org/maven2/net/sf/saxon/Saxon-HE/9.8.0-12/Saxon-HE-9.8.0-12.jar
+      curl -L -o saxon/saxon9he.jar https://repo1.maven.org/maven2/net/sf/saxon/Saxon-HE/9.8.0-12/Saxon-HE-9.8.0-12.jar
```
jorgebay: In the future we should add saxon to the C# driver Jenkins image.

src/Cassandra/Builder.cs
```diff
@@ -217,7 +214,98 @@ public Configuration GetConfiguration()
...
+                    lbp,
+                    _reconnectionPolicy,
+                    rep,
+                    sep,
+                    _timestampGenerator);
+            }
+
+            if (profile.LoadBalancingPolicy != null && _loadBalancingPolicy != null)
+            {
+                Builder.Logger.Warning(
```
jorgebay: Good idea to log a warning in these cases.

build.yaml
joao-r-reis: good idea, https://datastax-oss.atlassian.net/browse/CSHARP-848

PR #483, opened 2020-01-31
CSHARP-839 Fix concurrency bug with prepared statements cache (Linq2Cql/Mapper)

joao-r-reis:
Managed to reproduce the bug with the unit test changes.

### DIFF
--- a/src/Cassandra/Mapping/Statements/StatementFactory.cs
+++ b/src/Cassandra/Mapping/Statements/StatementFactory.cs
@@ -15,10 +15,11 @@
 //
 
 using System;
-using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+
+using Cassandra.Collections;
 using Cassandra.Tasks;
 
 namespace Cassandra.Mapping.Statements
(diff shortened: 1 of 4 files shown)


src/Cassandra/Mapping/Statements/StatementFactory.cs
```diff
@@ -66,7 +66,9 @@ public async Task<Statement> GetStatementAsync(ISession session, Cql cql, bool?
             var prepareTask = _statementCache.GetOrAdd(psCacheKey, _ =>
             {
                 wasPreviouslyCached = false;
-                return session.PrepareAsync(query);
+
+                // Use Task.Run to spend as little time as possible inside the collection lock
+                return Task.Run(() => session.PrepareAsync(query));
```
jorgebay: Maybe that's over-optimizing but 👍

src/Cassandra/Mapping/Statements/StatementFactory.cs
```diff
@@ -78,11 +80,12 @@ public async Task<Statement> GetStatementAsync(ISession session, Cql cql, bool?
             {
                 // The exception was caused from awaiting upon a Task that was previously cached
                 // It's possible that the schema or topology changed making this query preparation to succeed
-                // in a new attemp
-                prepareTask = session.PrepareAsync(query);
+                // in a new attempt
+                prepareTask = _statementCache.CompareAndUpdate(
```
jorgebay: I think we can make `CompareAndUpdate()` more generic with 2 params: `key` and `valueFactory`, then it can be used in the following way:

```csharp
// If there is a new prepare task from other thread, using the new one without making new requests
cache.CompareAndUpdate(key, (k, v) => v != prepareTask ? v : Task.Run(() => session.PrepareAsync(query)));
```
wdyt?

joao-r-reis: yeah that's a good idea, changed it!

PR #484, opened 2020-02-03
CSHARP-798 fail fast when there's a prepared statement mismatch

joao-r-reis:
This introduces a new exception that is thrown whenever a response to a re-prepare request contains a different `Id` than the one on the original `PreparedStatement`. Before this change, the driver would get into an infinite loop of re-prepare requests.

For `Mapper`/`Linq2Cql` users, the exception isn't very helpful since they have no control on the prepared statement cache. Making the mapper capable of handling this scenario would be quite the effort and the java driver Mapper has the same issue. I'm ok with just introducing this change to fail fast now since it's already much better than the current infinite loop.

### DIFF
--- a/src/Cassandra/Requests/RequestExecution.cs
+++ b/src/Cassandra/Requests/RequestExecution.cs
@@ -535,14 +535,9 @@ namespace Cassandra.Requests
 
                     if (!outputPrepared.QueryId.SequenceEqual(originalError.UnknownId))
                     {
-                        throw new PreparedStatementIdMismatchException(
-                            originalError.UnknownId,
-                            "ID mismatch while trying to reprepare (expected " 
-                            + $"{BitConverter.ToString(originalError.UnknownId).Replace("-", "")}, " 
-                            + $"got {BitConverter.ToString(outputPrepared.QueryId).Replace("-", "")}). " 
-                            + "This prepared statement won't work anymore. " 
-                            + "This usually happens when you run a 'USE...' query after " 
-                            + "the statement was prepared.");
+                        _parent.SetCompleted(new PreparedStatementIdMismatchException(
+                            originalError.UnknownId, outputPrepared.QueryId));
+                        return;
                     }
 
                     if (_parent.Statement is BoundStatement boundStatement)
--- a/src/Cassandra/Requests/RequestExecution.cs
+++ b/src/Cassandra/Requests/RequestExecution.cs
@@ -338,7 +338,7 @@ namespace Cassandra.Requests
             if (ex is PreparedQueryNotFoundException foundException &&
                 (_parent.Statement is BoundStatement || _parent.Statement is BatchStatement))
             {
-                PrepareAndRetry(foundException.UnknownId, host);
+                PrepareAndRetry(foundException, host);
                 return;
             }
             if (ex is NoHostAvailableException exception)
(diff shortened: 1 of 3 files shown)


src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -501,47 +501,66 @@ private void PrepareAndRetry(byte[] id, Host host)
...
+                    if (!outputPrepared.QueryId.SequenceEqual(originalError.UnknownId))
+                    {
+                        throw new PreparedStatementIdMismatchException(
+                            originalError.UnknownId,
+                            "ID mismatch while trying to reprepare (expected " 
+                            + $"{BitConverter.ToString(originalError.UnknownId).Replace("-", "")}, " 
+                            + $"got {BitConverter.ToString(outputPrepared.QueryId).Replace("-", "")}). " 
+                            + "This prepared statement won't work anymore. " 
+                            + "This usually happens when you run a 'USE...' query after " 
+                            + "the statement was prepared.");
```
jorgebay: NIT: The message building could be done within the `PreparedStatementIdMismatchException`, passing the `originalError.UnknownId` and `originalError.UnknownId` as parameters.

joao-r-reis: +1 on that, I'll push that change before merging

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -501,47 +501,66 @@ private void PrepareAndRetry(byte[] id, Host host)
...
+                    var output = ((ResultResponse) response).Output;
+                    if (!(output is OutputPrepared outputPrepared))
+                    {
+                        _parent.SetCompleted(new DriverInternalError("Expected prepared response, obtained " + output.GetType().FullName));
+                        return;
+                    }
+
+                    if (!outputPrepared.QueryId.SequenceEqual(originalError.UnknownId))
+                    {
+                        throw new PreparedStatementIdMismatchException(
```
joao-r-reis: @jorgebay now that I think about it, we should probably replace this with 

```
_parent.SetCompleted(new PreparedStatementIdMismatchException(...));
return;
```

This has performance benefits and is more consistent with the way we are returning errors in this same method.

joao-r-reis: The downside is that we lose stacktrace information (I think)

jorgebay: oh, good catch, yes I think it should be done with `SetCompleted()` (!)

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -501,47 +501,61 @@ private void PrepareAndRetry(byte[] id, Host host)
...
 
         /// <summary>
         /// Handles the response of a (re)prepare request and retries to execute on the same connection
         /// </summary>
-        private void ReprepareResponseHandler(IRequestError error, Response response, Host host)
+        private Action<IRequestError, Response, Host> NewReprepareResponseHandler(
+            PreparedQueryNotFoundException originalError)
         {
-            try
+            void ResponseHandler(IRequestError error, Response response, Host host)
```
jorgebay: Does creating a new handler each time impact performance? I suspect that this will be optimized by the compiler, into some kind of closure.

jorgebay: bah, nvm, this code is not within the hot path...

jorgebay:
Changes look good 👍

PR #485, opened 2020-02-03
CSHARP-845 CSHARP-697 Improve handling of synchronous Socket.ConnectAsync() call

joao-r-reis:


joao-r-reis:
> 
> 
> On `socket.ReceiveAsync()` and `socket.SendAsync()` we use a similar strategy (check the boolean) but we don't use `Task.Run()` execute the event asynchronously, should we unify it?

Hmm I propose addressing that when I work on [CSHARP-841](https://datastax-oss.atlassian.net/browse/CSHARP-841) which should be about next week. I propose this because that is a change to the hot path so I'll have to do some more testing (and benchmarks) which I will have to do anyway for CSHARP-841.

This way I would also merge only the connect changes contained in this PR to the DSE driver which is the bug fix itself.

jorgebay:
👍 Make sense.

PR #491, opened 2020-02-27
Introduce serverapi test category for server smoke tests

jorgebay:
https://datastax-oss.atlassian.net/browse/CSHARP-870

Add a new test category for server smoke tests.

I've included some DSE-specific tests into the category for in order to be useful for DSE smoke tests as well.


jorgebay:
Thanks! I'll merge it in order to update https://github.com/datastax/cassandra-drivers-smoke-test/pull/1

PR #492, opened 2020-02-27
Travis: Use .NET Core SDK 2.1.804

jorgebay:
Try to fix these [types of failures on Travis](https://travis-ci.com/datastax/csharp-driver/jobs/291690913) by updating the SDK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
   include:
     - dotnet: 2.2.204
       env: TARGETF="netcoreapp2.1" BUILD_EXAMPLES="1" BuildCoreOnly=True NETCORE_RUNTIME="2.1"
-    - dotnet: 2.1.202
+    - dotnet: 2.1.804
       env: TARGETF="netcoreapp2.0" BUILD_EXAMPLES="0" BuildCoreOnly=True NETCORE_RUNTIME="2.0"
       
 


joao-r-reis:
I've been down this rabbit hole in the past. What I found out at the time was that I was unable to get a version of the SDK that resolved this but didn't have the weird console bug where each test takes a couple of seconds to run (which causes the whole unit test run to last like 30 minutes or something ridiculous). https://github.com/dotnet/core/issues/2775

Looking at the issue it should be fixed with 804 but looking at travis output it's not: https://travis-ci.com/datastax/csharp-driver/jobs/291693589 it's extremely slow again and it will take 30 mins to complete

jorgebay:
😱 

Should we build against 1 of the .NET Core sdks alone on Travis to simplify it? like latest [lts 3.1.102](https://dotnet.microsoft.com/download/dotnet-core)

joao-r-reis:
Hmm I think that the only reason we have .NET Core 2.0 sdk there right now is because of this test: https://github.com/datastax/csharp-driver/blob/e50547f6f57eec71dd309988b3e2ec5607be61c7/src/Cassandra.Tests/DataStax/Cloud/CloudMetadataServiceTests.cs

We could try using 3.1 instead of 2.1 for the .NET Core 2.1 tests but we wouldn't really be running the tests against the LTS platform that we claim we support 😞

PR #494, opened 2020-03-03
CSHARP-405 improve handling of timeoutexception on cluster init

joao-r-reis:
With this change, the exception messages are more descriptive to hopefully make it obvious to the user that further attempts to initialize the cluster instance are useless. Also, I've added a way for the original error to be set as the `_initException` but we can also opt for logging this error instead of setting `_initException`.

### DIFF
--- a/src/Cassandra/Cluster.cs
+++ b/src/Cassandra/Cluster.cs
@@ -227,12 +227,12 @@ namespace Cassandra
                             " between the client driver instance and the cluster. You can increase this timeout via " +
                             "the SocketOptions.ConnectTimeoutMillis config setting. This can also be related to deadlocks " +
                             "caused by mixing synchronous and asynchronous code.", ex);
-                        _initException = new CachedInitErrorException(newEx);
+                        _initException = new InitFatalErrorException(newEx);
                         initTask.ContinueWith(t =>
                         {
                             if (t.IsFaulted && t.Exception != null)
                             {
-                                _initException = new CachedInitErrorException(t.Exception.InnerException);
+                                _initException = new InitFatalErrorException(t.Exception.InnerException);
                             }
                         }, TaskContinuationOptions.ExecuteSynchronously).Forget();
                         throw newEx;
@@ -267,7 +267,7 @@ namespace Cassandra
                 {
                     //There was an error that the driver is not able to recover from
                     //Store the exception for the following times
-                    _initException = new CachedInitErrorException(ex);
+                    _initException = new InitFatalErrorException(ex);
                     //Throw the actual exception for the first time
                     throw;
                 }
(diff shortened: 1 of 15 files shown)


src/Cassandra/Cluster.cs
```diff
@@ -213,8 +213,30 @@ private async Task Init()
...
+                    catch (TimeoutException ex)
+                    {
+                        var newEx = new TimeoutException(
+                            "Cluster initialization was aborted after timing out. This mechanism is put in place to" +
+                            " avoid blocking the calling thread forever. This usually caused by a networking issue" +
+                            " between the client driver instance and the cluster. You can increase this timeout via " +
+                            "the SocketOptions.ConnectTimeoutMillis config setting. This can also be related to deadlocks " +
+                            "caused by mixing synchronous and asynchronous code.", ex);
+                        _initException = new CachedInitErrorException(newEx);
+                        initTask.ContinueWith(t =>
```
jorgebay: I'm not sure I understand this continuation.

src/Cassandra/Exceptions/CachedInitErrorException.cs
```diff
@@ -0,0 +1,31 @@
...
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+using System;
+
+namespace Cassandra
+{
+    public class CachedInitErrorException : Exception
```
jorgebay: Nit: `Cached` is an implementation detail. We could use like `InitFatal...` or `InitUnrecoverableErrorException`.

src/Cassandra/Cluster.cs
joao-r-reis: When the init failure is caused by an `OperationTimedOutException`, currently we don't actually handle the result of the task so even if the init task returns an exception with more details about the failure, we just keep throwing the `OperationTimedOutException`. 

This continuation replaces the "cached" exception with the one that the init task returns. Another option would be to just make this continuation log the exception and keep throwing the timed out exception but I decided to go with the exception replacement approach since the single log message is very likely to be hidden in all the log spam that will inevitably happen if an application continuously calls `Connect` on a failed cluster.

On the next major we should change this so that there is no cached exceptions and no unrecoverable state https://datastax-oss.atlassian.net/browse/CSHARP-698

src/Cassandra/Exceptions/CachedInitErrorException.cs
joao-r-reis: `InitFatalErrorException` sgtm

src/Cassandra/Cluster.cs
jorgebay: ah, makes sense.

PR #495, opened 2020-03-04
CSHARP-846 Added missing replication strategies and reduced log verbosity

joao-r-reis:
I took this chance to cleanup some "noisy" log messages that were appearing during initialization. The "missing" strategies log message is no longer present because we have implementations for them now. 

Also the driver was logging messages for the default custom serializers (`Duration`, `DateRange`, etc.) which is just noise tbh. 

I moved these default serializers setup to the `GenericSerializer`. It made sense to add them in the builder before the unified driver (since these were only used by the DSE driver AFAIK) but now that is no longer required.


joao-r-reis:
I'll add a couple of unit tests later today

PR #497, opened 2020-03-05
CSHARP-851 CSHARP-877 Add refreshintervalmilliseconds option to appmetrics extension

joao-r-reis:


src/Extensions/Cassandra.AppMetrics/DriverAppMetricsOptions.cs
```diff
@@ -88,5 +93,34 @@ public DriverAppMetricsOptions SetTimersTimeUnit(TimeUnit timeUnit)
...
+        /// <para>
+        /// The interval at which percentile data is refreshed.
+        /// </para>
+        /// <para>
+        /// The driver records latency data in a "live" histogram, and serves results from a cached
+        /// snapshot. Each time the snapshot gets older than the interval, the two are switched.
+        /// Note that this switch happens upon fetching the metrics, so if you never fetch the
+        /// recording interval might grow higher (that shouldn't be an issue in a production
+        /// environment because you would typically have a metrics reporter that exports to a
+        /// monitoring tool at a regular interval).
```
jorgebay: Good docs!

joao-r-reis: I copied a lot of it from the java driver 😄 After this change the HdrHistogram metrics of both drivers have very similar implementations (if not equivalent)

PR #498, opened 2020-03-10
CSHARP-489 Move api docs to docfx

joao-r-reis:
DocFX has a nice template system where we can reference more than one template and it will merge them. I referenced one embedded template (`statictoc`) and one custom template. This custom template contains only the files that needed to be changed from the `statictoc` template so it's only a few files and I think it's fine if they are added to the repo. This allows anyone to build the docs (I added a `README` with instructions).

Preview:
imagem


joao-r-reis:
Also I added a custom empty index page that only redirects to the API docs section but maybe we could add a simple index page with a small README or something... Maybe it would be a better user experience. (it's basically the same workaround that is being used in the C# Graph API Docs)

jorgebay:
Would it be possible to move the template to documentor repo / doc team repo?

That way we can avoid having to maintain it along side our source code.

joao-r-reis:
Done

doc/api-docs/README.md
```diff
@@ -0,0 +1,12 @@
+# How to build the C# Driver API Docs
```
jorgebay: Good idea to include the doc here.

PR #502, opened 2020-03-23
Fix re-prepare error log spam

joao-r-reis:
This change removes the duplicate log message when a `PreparedQueryNotFound` error is returned by the server.

No need to review.


joao-r-reis:
Merged in 6bf05fc15af98bcc2b13a2de11faf2819fa1e393

PR #503, opened 2020-03-29
Cleaning unused code.

ralmsdeveloper:


src/Cassandra/Helpers/PlatformHelper.cs
```diff
@@ -52,17 +52,17 @@ public static CpuInfo GetCpuInfo()
...
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                 {
-                    return PlatformHelper.GetLinuxProcCpuInfo();
+                    return GetLinuxProcCpuInfo();
                 }
             }
             catch (Exception ex)
             {
-                PlatformHelper.Logger.Info("Could not get cpu name. Exception: {0}", ex.ToString());
+                Logger.Info("Could not get cpu name. Exception: {0}", ex.ToString());
```
joao-r-reis: To make static access clear, we should keep the qualifiers here.

ralmsdeveloper: Hello @joao-r-reis, although it doesn't make much sense, since the access is internal and not external, I will make the change.


Thanks for the comment.

joao-r-reis:
Thanks for the contribution!

PR #506, opened 2020-03-30
Standardization System.String to string

ralmsdeveloper:
* Simple code improvements, low impact.

Note:
There is no gain or loss of performance.
**string** is just an alias for **System.String**, it's a cleaner way to code, eliminates the possibility of using the namespace "System".

IHMO: I believe this is a way to start standardizing small things, the project is great!

### DIFF
--- a/src/Cassandra.IntegrationTests/Core/ClientWarningsTests.cs
+++ b/src/Cassandra.IntegrationTests/Core/ClientWarningsTests.cs
@@ -58,8 +58,8 @@ namespace Cassandra.IntegrationTests.Core
             });
             _testCluster.InitClient();
             Session = _testCluster.Session;
-            Session.Execute(String.Format(TestUtils.CreateKeyspaceSimpleFormat, Keyspace, 1));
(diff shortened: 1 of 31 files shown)


ralmsdeveloper:
Okay, I have more stuff to send :)

I'm starting to use ScyllaDB, so don't be mad at me sending some PR's 😂

After I send other things, I will try to take a deep dive into asynchronous processing!

joao-r-reis:
Sure, always happy to see contributions. Keep in mind that it might be worth opening a JIRA ticket and get some discussion going in case you want to contribute a new feature or improvement that hasn't been discussed previously. For small stuff like what you've been submitting so far, it's fine if there's no JIRA.

Here's a link to the JIRA project https://datastax-oss.atlassian.net/projects/CSHARP

joao-r-reis:
You can also use the mailing list to start a discussion that is not related to a specific JIRA ticket: https://groups.google.com/a/lists.datastax.com/forum/#!forum/csharp-driver-user

PR #507, opened 2020-03-31
Improvement: Simplify GetValues

ralmsdeveloper:
* Risk level: low

### DIFF
--- a/src/Cassandra/Connections/ControlConnection.cs
+++ b/src/Cassandra/Connections/ControlConnection.cs
@@ -680,8 +680,7 @@ namespace Cassandra.Connections
             //The address in the Cassandra event message needs to be translated
             var address = TranslateAddress(e.Address);
             ControlConnection._logger.Info("Received Node status change event: host {0} is {1}", address, e.What.ToString().ToUpper());
-            Host host;
-            if (!_metadata.Hosts.TryGet(address, out host))
+            if (!_metadata.Hosts.TryGet(address, out var host))
             {
                 ControlConnection._logger.Info("Received status change event for host {0} but it was not found", address);
                 return;
--- a/src/Cassandra/Mapping/PocoDataFactory.cs
+++ b/src/Cassandra/Mapping/PocoDataFactory.cs
@@ -66,7 +66,7 @@ namespace Cassandra.Mapping
         private PocoData CreatePocoData(Type pocoType)
         {
             // Try to get mapping from predefined collection, otherwise fallback to using attributes
-            if (_predefinedTypeDefinitions.TryGetItem(pocoType, out ITypeDefinition typeDefinition) == false)
+            if (!_predefinedTypeDefinitions.TryGetItem(pocoType, out ITypeDefinition typeDefinition))
             {
                 typeDefinition = new AttributeBasedTypeDefinition(pocoType);
             }
--- a/src/Cassandra/Mapping/PocoDataFactory.cs
+++ b/src/Cassandra/Mapping/PocoDataFactory.cs
@@ -40,8 +40,7 @@ namespace Cassandra.Mapping
         /// <param name="predefinedTypeDefinitions">Explicitly declared type definitions</param>
         public PocoDataFactory(LookupKeyedCollection<Type, ITypeDefinition> predefinedTypeDefinitions)
         {
-            if (predefinedTypeDefinitions == null) throw new ArgumentNullException("predefinedTypeDefinitions");
-            _predefinedTypeDefinitions = predefinedTypeDefinitions;
+            _predefinedTypeDefinitions = predefinedTypeDefinitions ?? throw new ArgumentNullException("predefinedTypeDefinitions");
             _cache = new ConcurrentDictionary<Type, PocoData>();
         }
 
@@ -67,8 +66,7 @@ namespace Cassandra.Mapping
         private PocoData CreatePocoData(Type pocoType)
         {
             // Try to get mapping from predefined collection, otherwise fallback to using attributes
-            ITypeDefinition typeDefinition;
-            if (_predefinedTypeDefinitions.TryGetItem(pocoType, out typeDefinition) == false)
+            if (_predefinedTypeDefinitions.TryGetItem(pocoType, out ITypeDefinition typeDefinition) == false)
             {
                 typeDefinition = new AttributeBasedTypeDefinition(pocoType);
             }
--- a/src/Cassandra/Outputs/OutputError.cs
+++ b/src/Cassandra/Outputs/OutputError.cs
@@ -59,7 +59,7 @@ namespace Cassandra
 
         internal static OutputError CreateOutputError(int code, string message, FrameReader cb)
         {
-            if (OutputErrorFactoryMethods.TryGetValue(code, out Func<OutputError> factoryMethod) == false)
+            if (!OutputErrorFactoryMethods.TryGetValue(code, out Func<OutputError> factoryMethod))
                 throw new DriverInternalError(string.Format("Received unknown error with code {0} and message {1}", code, message));
 
             var error = factoryMethod();
--- a/src/Cassandra/Outputs/OutputError.cs
+++ b/src/Cassandra/Outputs/OutputError.cs
@@ -59,8 +59,7 @@ namespace Cassandra
 
         internal static OutputError CreateOutputError(int code, string message, FrameReader cb)
         {
-            Func<OutputError> factoryMethod;
-            if (OutputErrorFactoryMethods.TryGetValue(code, out factoryMethod) == false)
+            if (OutputErrorFactoryMethods.TryGetValue(code, out Func<OutputError> factoryMethod) == false)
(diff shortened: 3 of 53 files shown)


src/Cassandra/RecyclableMemoryStreamManager.cs
```diff
@@ -290,9 +289,8 @@ internal byte[] GetBlock()
         /// will be at least the requiredSize and always be a multiple of largeBufferMultiple.
         /// </summary>
         /// <param name="requiredSize">The minimum length of the buffer</param>
-        /// <param name="tag">The tag of the stream returning this buffer, for logging if necessary.</param>
         /// <returns>A buffer of at least the required size.</returns>
-        internal byte[] GetLargeBuffer(int requiredSize, string tag)
+        internal byte[] GetLargeBuffer(int requiredSize)
```
ralmsdeveloper: FYI: Removed unused parameter tag

src/Cassandra/Mapping/PocoDataFactory.cs
```diff
@@ -67,8 +66,7 @@ public void AddDefinitionDefault(Type type, Func<ITypeDefinition> definitionHand
         private PocoData CreatePocoData(Type pocoType)
         {
             // Try to get mapping from predefined collection, otherwise fallback to using attributes
-            ITypeDefinition typeDefinition;
-            if (_predefinedTypeDefinitions.TryGetItem(pocoType, out typeDefinition) == false)
+            if (_predefinedTypeDefinitions.TryGetItem(pocoType, out ITypeDefinition typeDefinition) == false)
```
joao-r-reis:
```suggestion
            if (!_predefinedTypeDefinitions.TryGetItem(pocoType, out ITypeDefinition typeDefinition))
```

src/Cassandra/Outputs/OutputError.cs
```diff
@@ -59,8 +59,7 @@ public void Dispose()
 
         internal static OutputError CreateOutputError(int code, string message, FrameReader cb)
         {
-            Func<OutputError> factoryMethod;
-            if (OutputErrorFactoryMethods.TryGetValue(code, out factoryMethod) == false)
+            if (OutputErrorFactoryMethods.TryGetValue(code, out Func<OutputError> factoryMethod) == false)
```
joao-r-reis:
```suggestion
            if (!OutputErrorFactoryMethods.TryGetValue(code, out Func<OutputError> factoryMethod))
```

src/Cassandra/RecyclableMemoryStreamManager.cs
joao-r-reis: I know that this parameter is not used atm but it could be very useful for trace/verbose logging as the XML comment states so I'd rather keep them than removing them now and readding them later. There's a couple other places with this same parameter and it's the same scenario.

src/Cassandra/RowPopulators/Row.cs
```diff
@@ -133,12 +133,9 @@ public virtual bool IsNull(int index)
...
         {
-            int index;
-            if (!ColumnIndexes.TryGetValue(name, out index))
-            {
-                return null;
-            }
-            return Columns[index];
+            return !ColumnIndexes.TryGetValue(name, out var index) 
+                ? null 
+                : Columns[index];
```
joao-r-reis:
```suggestion
            return ColumnIndexes.TryGetValue(name, out var index) 
                ? Columns[index]
                : null;
```

easier to read imo

src/Cassandra/SimpleStatement.cs
```diff
@@ -160,14 +160,10 @@ public SimpleStatement(IDictionary<string, object> valuesDictionary, string quer
...
             }
-            if (query == null)
-            {
-                throw new ArgumentNullException("query");
-            }
             //The order of the keys and values is unspecified, but is guaranteed to be both in the same order.
             SetParameterNames(valuesDictionary.Keys);
             base.SetValues(valuesDictionary.Values.ToArray());
-            _query = query;
+            _query = query ?? throw new ArgumentNullException("query");
```
joao-r-reis: move this to before `SetParameterNames` to fail faster

src/Cassandra/Connections/ControlConnection.cs
```diff
@@ -680,8 +680,7 @@ private void HandleStatusChangeEvent(StatusChangeEventArgs e)
             //The address in the Cassandra event message needs to be translated
             var address = TranslateAddress(e.Address);
             ControlConnection._logger.Info("Received Node status change event: host {0} is {1}", address, e.What.ToString().ToUpper());
-            Host host;
-            if (!_metadata.Hosts.TryGet(address, out host))
+            if (!_metadata.Hosts.TryGet(address, out var host))
```
joao-r-reis: This the only place you've used `out var`, is there a particular reason?

src/Cassandra/RowPopulators/Row.cs
ralmsdeveloper: Yes of course, I'm still analyzing the whole code, but I thought about using the pareto principle here!

I will make the adjustment!

src/Cassandra/Mapping/PocoDataFactory.cs
ralmsdeveloper: Generally for methods I find it much more robust to use ` == (boolean)`

For sample:
```csharp
{
    bool? getTryValue() => false;

    if (!getTryValue()) { } //Error
    if (getTryValue() == false) { } //Success
}
```

src/Cassandra/Connections/ControlConnection.cs
ralmsdeveloper: Just simplification, I don't need to declare a local variable, since if it exists in my list, I will bring an instance of it.

joao-r-reis: I'm sorry I don't understand. I'm asking specifically about `out var host` vs `out Host host` since everywhere else you used explicit type declarations and this is the only place where you used an implicit one.

src/Cassandra/RowPopulators/Row.cs
```diff
@@ -133,9 +133,9 @@ public virtual bool IsNull(int index)
         /// </summary>
         public CqlColumn GetColumn(string name)
         {
-            return !ColumnIndexes.TryGetValue(name, out var index) 
-                ? null 
-                : Columns[index];
+            return !ColumnIndexes.TryGetValue(name, out var index)
+                ? Columns[index]
+                : null;
```
joao-r-reis:
```suggestion
            return ColumnIndexes.TryGetValue(name, out var index)
                ? Columns[index]
                : null;
```
this change is what is causing the test failures (at least some of them)

src/Cassandra/Mapping/PocoDataFactory.cs
joao-r-reis: I understand the argument for that approach for sure but we should use `!` here to keep it consistent.

src/Cassandra/Connections/ControlConnection.cs
ralmsdeveloper: Sorry I got your question wrong, this was a point I would like to talk to you about, if we could start using `var` instead of implicit declaration.

This was specifically due to my impulse to use `var`.

Would you be okay with using var?

src/Cassandra/RowPopulators/Row.cs
ralmsdeveloper: I'm looking at this, I didn't expect to run the tests.

src/Cassandra/Connections/ControlConnection.cs
joao-r-reis: Yes! I've been changing type declarations to `var` when I touch code that is not using `var` so +1 from me.

src/Cassandra/RowPopulators/Row.cs
joao-r-reis: You forgot to swap the lines when you removed the `!`, not a big deal. The build should return to green after this is fixed.

ralmsdeveloper: @joao-r-reis thanks for the interaction, let me ask you a question.

Is there any limitation in the provider for ScyllaDB, is anything not yet supported?

joao-r-reis: We support this driver for use with Apache Cassandra and any other backend that is fully compliant with Cassandra's native protocol and CQL specifications. 
We do not test this driver with ScyllaDB and do not monitor the ScyllaDB implementation for functionality that would impact this driver's capability with that database.

ralmsdeveloper: I'm going to create another project focused on ScyllaDB based on this driver and start doing deeper tests with ScyllaDB, my main focus is ScyllaDB.
I will keep all copyrights, initially it will be a project focused on testing the Driver and compatibility.
I'm enjoying everything I saw here.
Congratulations to you and @datastax.

PR #510, opened 2020-04-08
CSHARP-884 Fix controlconnection race condition in dispose

joao-r-reis:


src/Cassandra/Connections/Connection.cs
```diff
@@ -417,7 +420,9 @@ public async Task<Response> Open()
         {
             try
             {
-                return await DoOpen().ConfigureAwait(false);
+                var response = await DoOpen().ConfigureAwait(false);
+                Connection.Logger.Verbose("Opened Connection #{0} to {1}.", GetHashCode(), EndPoint.EndpointFriendlyName);
```
jorgebay, line 424: Good idea to log it at verbose level.

PR #512, opened 2020-04-12
CSHARP-895 Table<TEntity>: add ability to create tables asynchronously

halex2005:
Hi!
It is necessary to me to create tables in thread pool.
I suggest to create async-counterparts to `Table<TEntity>.Create()` and `Table<TEntity>.CreateIfNotExists()` methods.


joao-r-reis:
Created https://datastax-oss.atlassian.net/browse/CSHARP-895 to track this

halex2005:
When new version of driver with that feature could be available in nuget?

joao-r-reis:
This will be part of release 3.15.0 of the driver which is scheduled to be released somewhere in the middle of the next month (May).

PR #513, opened 2020-04-14
CSHARP-887 Add support for transient replicas

joao-r-reis:
There's also a test improvement that is not related to the ticket, I saw that random test failure in today's nightly build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -447,6 +447,7 @@ pipeline {
             values '2.2',     // latest 2.2.x Apache Cassandara�
                   '3.0',     // latest 3.0.x Apache Cassandara�
                   '3.11',    // latest 3.11.x Apache Cassandara�
+                  '4.0',    // Development Apache Cassandara�
                   'dse-5.1', // latest 5.1.x DataStax Enterprise
                   'dse-6.7', // latest 6.7.x DataStax Enterprise
                   'dse-6.8.0' // 6.8.0 current DataStax Enterprise
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -444,10 +444,10 @@ pipeline {
         axes {
           axis {
             name 'SERVER_VERSION'
-            values '2.2',     // latest 2.2.x Apache Cassandara�
-                  '3.0',     // latest 3.0.x Apache Cassandara�
-                  '3.11',    // latest 3.11.x Apache Cassandara�
-                  '4.0',    // Development Apache Cassandara�
+            values '2.2',     // latest 2.2.x Apache Cassandra�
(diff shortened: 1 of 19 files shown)


src/Cassandra.IntegrationTests/TestBase/TestDseVersion.cs
```diff
@@ -44,8 +46,8 @@ public class TestDseVersion : NUnitAttribute, IApplyToTest
         /// Determines if the DSE version required should be "greater or equals to" = 1,
         /// "equals to" = 0, "less than or equal to " = -1
         /// </param>
-        public TestDseVersion(int major, int minor, Comparison comparison = Comparison.GreaterThanOrEqualsTo)
-            : this(major, minor, 0, comparison)
+        public TestDseVersion(int major, int minor, Comparison comparison = Comparison.GreaterThanOrEqualsTo, bool isOssRequired = false)
+            : this(major, minor, 0, comparison, isOssRequired)
```
jorgebay, line 50: Good idea to filter by server type.

src/Cassandra.Tests/MetadataHelpers/SimpleStrategyTests.cs
```diff
@@ -24,9 +24,27 @@ namespace Cassandra.Tests.MetadataHelpers
...
+            foreach (var token in result)
+            {
+                Assert.AreEqual(2, token.Value.Count);
+            }
+        }
+        
+        [Test]
+        public void Should_ReturnTwoReplicasPerToken_When_TransientReplicationIsEnabled()
+        {
+            var target = new SimpleStrategy(ReplicationFactor.Parse("3/1"));
```
jorgebay, line 47: Replication factor is 3 but expected tokens are 2, is this because of the test data?

joao-r-reis: We basically only consider full replicas when computing token map so in this case there are 2 full replicas and 1 transient replica so the expected replicas count per token is 2

jorgebay: oh snap, I was reading it as 3+1, nvm

Jenkinsfile
```diff
@@ -447,6 +447,7 @@ pipeline {
             values '2.2',     // latest 2.2.x Apache Cassandara�
                   '3.0',     // latest 3.0.x Apache Cassandara�
                   '3.11',    // latest 3.11.x Apache Cassandara�
+                  '4.0',    // Development Apache Cassandara�
```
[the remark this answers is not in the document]
joao-r-reis: I don't know how I missed this 🤣  fixed!

PR #514, opened 2020-04-15
CSHARP-664 Implement SKIP_METADATA for protocols v5+ and fix CSHARP-896

joao-r-reis:
- [x] add tests
- [x] test both `paging_state` and `new_metadata_id` flags at the same time (I think they're being parsed in the wrong order) - CSHARP-896
- [x] performance tests (use DSE 6.7 to compare this branch vs master)

Note: this PR also includes some improvements to the `ISerializer` references but I tried to split them into a separate commit for an easier review (it's not a perfect split though, so the first commit will probably not compile)


joao-r-reis:
Benchmarks show an improvement in throughput and cpu usage for read workloads when using protocol v5+ (Apache Cassandra 4.0+) or DseV2 (DSE 6.0+)

src/Cassandra/BoundStatement.cs
```diff
@@ -115,31 +109,32 @@ public BoundStatement SetRoutingKey(params RoutingKey[] routingKeyComponents)
             return this;
         }
 
-        internal override void SetValues(object[] values)
+        internal override void SetValues(object[] values, ISerializer serializer)
```
jorgebay, line 112: 💯 get rid of the class level field for the serializer!

src/Cassandra/Connections/Connection.cs
```diff
@@ -712,7 +723,7 @@ void DeserializeResponseStream(MemoryStream stream, long timestamp)
                         plainTextStream = compressor.Decompress(new WrappedStream(stream, header.BodyLength));
                         plainTextStream.Position = 0;
                     }
-                    response = FrameParser.Parse(new Frame(header, plainTextStream, serializer));
+                    response = FrameParser.Parse(new Frame(header, plainTextStream, serializer), resultMetadata);
```
jorgebay, line 726: Result metadata could be stored at Frame level, that way we can avoid passing the result metadata as parameter for  `Create()` handlers, wdyt?

src/Cassandra/Requests/IRequest.cs
```diff
@@ -25,5 +25,11 @@ internal interface IRequest
         /// Writes the frame for this request on the provided stream
         /// </summary>
         int WriteFrame(short streamId, MemoryStream stream, ISerializer serializer);
+
+        /// <summary>
+        /// Result Metadata to parse the response rows. Only EXECUTE requests set this value so it will be null
+        /// for other types of requests.
+        /// </summary>
+        ResultMetadata ResultMetadata { get; }
```
jorgebay, line 33: We should move this declaration to `IQueryRequest`, that way we avoid to create it for requests that don't make sense to have it.

joao-r-reis: I thought about doing that but then I would have to add a `request as IQueryRequest` statement on the `OperationState` constructor. Do you think we should add that cast instead of defining this property on every request class?

src/Cassandra/Connections/Connection.cs
joao-r-reis: I like that idea, I'll change it

src/Cassandra/Requests/IRequest.cs
jorgebay: I was saying that because having the property in each implementation didn't make much sense for me (i.e., having a `ResultMetadata` prop within `OptionsRequest`), but it doesn't affect execution and its a matter of opinions so I understand if you prefer leaving it as is and avoid a cast.

joao-r-reis: Oh I agree that it doesn't make sense, it's just that my personal opinion is that the cast is worse here and other potentially better alternatives would require rewriting a lot of stuff. 

I'll go ahead and merge as is but I'll keep this in mind in the future (maybe when implementing protocol checksumming if it requires a lot of changes to this code?)

PR #515, opened 2020-04-17
CSHARP-871 add support for system.peers_v2

joao-r-reis:
- [x] Add basic unit tests
- [x] Add unit test to verify that peers v2 queries are never attempted again after one fails (with an `InvalidQuery` error)
- [x] Add integration test with hosts on the same address but different ports

### DIFF
--- a/src/Cassandra/Connections/Control/TopologyRefresher.cs
+++ b/src/Cassandra/Connections/Control/TopologyRefresher.cs
@@ -19,6 +19,7 @@ using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using Cassandra.Responses;
+using Cassandra.Tasks;
 
 namespace Cassandra.Connections.Control
 {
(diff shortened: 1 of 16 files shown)


src/Cassandra/Connections/Control/TopologyRefresher.cs
```diff
@@ -18,41 +18,54 @@
...
-                                          _config.MetadataRequestHandler.SendMetadataRequestAsync(
-                                              connection, version, TopologyRefresher.SelectPeers, QueryProtocolOptions.Default))
-                                      .ConfigureAwait(false);
+            // safe guard against concurrent changes of this field
+            var localIsPeersV2 = _isPeersV2;
+
+            var localTask = SendSystemLocalRequestAsync(connection, version);
+            var peersTask = SendSystemPeersRequestAsync(localIsPeersV2, connection, version);
+            
+            await Task.WhenAll(localTask, peersTask).ConfigureAwait(false);
```
jorgebay, line 61: NIT: There's no need to have `WhenAll()` as both tasks are awaited upon below.

jorgebay: oh, true... my bad, nvm.

joao-r-reis: Changed the following `await`s to `.Result`

PR #517, opened 2020-04-27
CSHARP-833 fix api docs of keyspacemetadata export method

joao-r-reis:


joao-r-reis:
+1 but I think we should wait for https://issues.apache.org/jira/browse/CASSANDRA-14825 and support this on C* 4.0+ only

PR #518, opened 2020-04-27
CSHARP-881 Add notes about server name and certificate name mismatch errors

joao-r-reis:


doc/features/tls/README.md
```diff
@@ -29,9 +31,18 @@ var cluster = Cluster.Builder()
...
 
-By default, `Builder.WithSSL()` adds a `RemoteCertificateValidationCallback` that logs any errors returned by .NET's SSL API.
+```csharp
+var cluster = Cluster.Builder()
+    .AddContactPoints(...)
+    .WithSSL(new SSLOptions().SetHostNameResolver(...))
+    .Build();
+```
+
+Alternatively, you can provide your own `RemoteCertificateValidationCallback` to handle the SSL errors and perform your own validation logic. Note that you can introduce security vulnerabilities with this so you should avoid it if you can.
```
adutra, line 43: Is it worth adding an example for this? Or do we consider this as an advanced feature and as such, people should know how to do it?

doc/features/tls/README.md
```diff
@@ -56,7 +67,9 @@ var cluster = Cluster.Builder()
     .Build();
 ```
 
-### Client authentication
+The examples on the [driver's repository][Github repository] have a basic custom certificate validator which can be used as a starting point.
+
+### Enabling client authentication
 
 To enable client authentication, you need to provide the driver the client certificate(s):
```
adutra, line 74: Unrelated but you might want to fix the grammar: `provide the driver with the client certificate(s)`.

doc/features/tls/README.md
joao-r-reis: We have an example of a validator for a custom root certificate scenario and I mention it below. Do you think that I should mention it here too?

doc/features/tls/README.md
joao-r-reis: nice catch, fixed 👍

doc/features/tls/README.md
adutra: Oh right, there is an inlined example at line 65, then a mention of an external example line 70. That's fine then :-) Maybe just add a `(see below for an example of this)` kind of thing?

joao-r-reis: done!

PR #520, opened 2020-04-30
CSHARP-719: use K4os.Compression.LZ4 instead of lz4net to support LZ4…

halex2005:
Use [K4os.Compression.LZ4](https://www.nuget.org/packages/K4os.Compression.LZ4/1.1.11) instead of [lz4net](https://www.nuget.org/packages/lz4net/) to support LZ4 compression for all netstandard targets.

Original issue: https://datastax-oss.atlassian.net/browse/CSHARP-719

### DIFF
--- a/src/Cassandra/Compression/LZ4Compressor.cs
+++ b/src/Cassandra/Compression/LZ4Compressor.cs
@@ -16,7 +16,14 @@
 
 using System;
 using System.IO;
+
+#if NETFRAMEWORK
+using LZ4;
+#endif
+
+#if NETSTANDARD2_0
 using K4os.Compression.LZ4;
+#endif
 
 namespace Cassandra.Compression
 {
--- a/src/Cassandra/Compression/LZ4Compressor.cs
+++ b/src/Cassandra/Compression/LZ4Compressor.cs
@@ -14,9 +14,9 @@
 //   limitations under the License.
 //
 
-#if NETFRAMEWORK
 using System;
 using System.IO;
+using K4os.Compression.LZ4;
 
 namespace Cassandra.Compression
 {
@@ -25,13 +25,39 @@ namespace Cassandra.Compression
         public Stream Decompress(Stream stream)
         {
             var buffer = Utils.ReadAllBytes(stream, 0);
+            if (buffer.Length < 4)
+            {
+                throw new CorruptionException("Corrupt literal length");
+            }
+
             var outputLengthBytes = new byte[4];
             Buffer.BlockCopy(buffer, 0, outputLengthBytes, 0, 4);
             Array.Reverse(outputLengthBytes);
             var outputLength = BitConverter.ToInt32(outputLengthBytes, 0);
-            var decompressStream = new MemoryStream(LZ4.LZ4Codec.Decode(buffer, 4, buffer.Length - 4, outputLength), 0, outputLength, false, true);
+            var outputBuffer = new byte[outputLength];
+            var uncompressedSize = LZ4Codec.Decode(buffer, 4, buffer.Length - 4, outputBuffer, 0, outputLength);
+            if (uncompressedSize < 0)
+            {
+                throw new CorruptionException("Can't decode LZ4 bytes");
+            }
+
+            if (outputLength != uncompressedSize)
+            {
+                throw new CorruptionException(string.Format("Recorded length is {0} bytes but actual length after decompression is {1} bytes ",
+                    outputLength,
+                    uncompressedSize));
+            }
+
+            var decompressStream = new MemoryStream(outputBuffer, 0, outputLength, false, true);
             return decompressStream;
+        }
+
+        private class CorruptionException : Exception
+        {
+            public CorruptionException(string message)
+                : base(message)
+            {
+            }
         }
     }
 }
-#endif
--- a/src/Cassandra/Compression/LZ4Compressor.cs
+++ b/src/Cassandra/Compression/LZ4Compressor.cs
@@ -27,7 +27,7 @@ namespace Cassandra.Compression
             var buffer = Utils.ReadAllBytes(stream, 0);
             if (buffer.Length < 4)
             {
-                throw new CorruptionException("Corrupt literal length");
+                throw new DriverInternalError("Corrupt literal length");
             }
 
             var outputLengthBytes = new byte[4];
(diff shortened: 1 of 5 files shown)


joao-r-reis:
As mentioned on the JIRA ticket, compatibility between these two libraries is not guaranteed so we don't want to replace the implementation on a minor release. 

How urgent is this for your use case? If you can't wait for the next major (which will be under development for several months) then we could add this new dependency and implementation just for `netstandard` until the next major where we drop `lz4net` completely.

src/Cassandra/Compression/LZ4Compressor.cs
```diff
@@ -25,13 +25,39 @@ internal class LZ4Compressor : IFrameCompressor
...
+            {
+                throw new CorruptionException("Can't decode LZ4 bytes");
+            }
+
+            if (outputLength != uncompressedSize)
+            {
+                throw new CorruptionException(string.Format("Recorded length is {0} bytes but actual length after decompression is {1} bytes ",
+                    outputLength,
+                    uncompressedSize));
+            }
```
joao-r-reis, line 49: This implementation doesn't take into account frames without body, which is the reason for the test failures. This should fix it:

```suggestion
            if (outputLength != 0)
            {
                var uncompressedSize = LZ4Codec.Decode(buffer, 4, buffer.Length - 4, outputBuffer, 0, outputLength);
                if (uncompressedSize < 0)
                {
                    throw new CorruptionException("Can't decode LZ4 bytes");
                }

                if (outputLength != uncompressedSize)
                {
                    throw new CorruptionException(string.Format("Recorded length is {0} bytes but actual length after decompression is {1} bytes ",
                        outputLength,
                        uncompressedSize));
                }
            }
```

src/Cassandra/Compression/LZ4Compressor.cs
```diff
@@ -25,13 +25,39 @@ internal class LZ4Compressor : IFrameCompressor
...
+
+            var decompressStream = new MemoryStream(outputBuffer, 0, outputLength, false, true);
             return decompressStream;
         }
+
+        private class CorruptionException : Exception
```
joao-r-reis, line 55: I would honestly just use `DriverInternalError` here instead of creating a new exception

src/Cassandra/Compression/LZ4Compressor.cs
```diff
@@ -25,13 +25,39 @@ internal class LZ4Compressor : IFrameCompressor
         public Stream Decompress(Stream stream)
         {
             var buffer = Utils.ReadAllBytes(stream, 0);
+            if (buffer.Length < 4)
+            {
+                throw new CorruptionException("Corrupt literal length");
```
joao-r-reis, line 30: I would change this exception message to be a bit more specific like `Unable to parse uncompressed size due to invalid buffer length`

src/Cassandra/Compression/LZ4Compressor.cs
```diff
@@ -25,13 +25,39 @@ internal class LZ4Compressor : IFrameCompressor
...
             var outputLengthBytes = new byte[4];
             Buffer.BlockCopy(buffer, 0, outputLengthBytes, 0, 4);
             Array.Reverse(outputLengthBytes);
             var outputLength = BitConverter.ToInt32(outputLengthBytes, 0);
-            var decompressStream = new MemoryStream(LZ4.LZ4Codec.Decode(buffer, 4, buffer.Length - 4, outputLength), 0, outputLength, false, true);
+            var outputBuffer = new byte[outputLength];
+            var uncompressedSize = LZ4Codec.Decode(buffer, 4, buffer.Length - 4, outputBuffer, 0, outputLength);
+            if (uncompressedSize < 0)
+            {
+                throw new CorruptionException("Can't decode LZ4 bytes");
```
joao-r-reis, line 41: I would change this message to something more specific like `LZ4 decoded buffer has a larger size than the expected uncompressed size`

src/Cassandra/Compression/LZ4Compressor.cs
halex2005: Agreed, changed.

src/Cassandra/Compression/LZ4Compressor.cs
halex2005: Thanks for the point!

src/Cassandra/Compression/LZ4Compressor.cs
halex2005: Agreed, done

src/Cassandra/Compression/LZ4Compressor.cs
halex2005: Yes, LZ4 tests are passing now

halex2005:
Hi, @joao-r-reis.

> As mentioned on the JIRA ticket, compatibility between these two libraries is not guaranteed so we don't want to replace the implementation on a minor release.

> How urgent is this for your use case? If you can't wait for the next major (which will be under development for several months) then we could add this new dependency and implementation just for netstandard until the next major where we drop lz4net completely.

Actually, I'd like to have LZ4 support for `netstandard` targets.

Now I understand that `lz4net` and `K4os.Compression.LZ4` may have incompatibilities. Therefore in this PR I've returned `lz4net` dependency for .NET Framework target and use `K4os.Compression.LZ4` only for `netstandard2.0` target.

halex2005:
> until the next major where we drop lz4net completely

@joao-r-reis, one more question: is `Snappy` preferred compression algorithm for Cassandra for the future?

joao-r-reis:
> @joao-r-reis, one more question: is `Snappy` preferred compression algorithm for Cassandra for the future?

Sorry I meant just the legacy `lz4net` dependency, not LZ4 compression altogether, i.e., using `K4os.Compression.LZ4`  instead of `lz4net`.

LZ4 is generally preferred over Snappy. This is mentioned on the [Apache Cassandra docs](https://cassandra.apache.org/doc/latest/operating/compression.html)
> Snappy is kept for backwards compatibility and LZ4 will typically be preferable.

halex2005:
> LZ4 is generally preferred over Snappy.

Yes, therefore I was wondered :)  Thanks!

All review suggestions are fixed, all checks have passed.
Please, resolve conversations if all is fine now?

joao-r-reis:
> 
> 
> Lgtm, I will post an update here when we figure out which release will have this.

I posted an update on the [JIRA ticket](https://datastax-oss.atlassian.net/browse/CSHARP-719)

PR #528, opened 2020-06-23
Change session/cluster initialization to happen in the background CSHARP-916

joao-r-reis:
Currently `cluster.Connect()` starts the initialization task of the control connection, then creates the session, then initializes the session.

This PR changes this behavior so that `cluster.Connect()`  creates the session object and returns it to the user immediately. The session initialization task is started in the background which will start the cluster initialization if it's not initialized yet. 

If a user attempts to do something with the `session`, `cluster` or `cluster.Metadata` that requires the initialization to be finished, the method will await/block until the initialization task is complete. 

I also implemented [CSHARP-698](https://datastax-oss.atlassian.net/browse/CSHARP-698) on this PR. It uses the reconnection policy to keep retrying the control connection initialization in the background. **After an initialization attempt fails, all method calls (`session.Execute`, etc.) will throw the original exception **until a new attempt is started**. When a new attempt is started, the methods will block again until the current attempt finishes.

This PR also adds `Session.Connect()` for users that want to wait for the initialization to be finished.

Several API changes are required in order for this to be a good experience for the user.

The `Cluster.Metadata` property no longer blocks. Also introduced a `IMetadata` interface and a `IMetadataSnapshotProvider` that declares some `Metadata` methods which do not block.

`ILoadBalancingPolicy.Initialize` is called by the initialization method but a `ICluster` instance is passed to that method which means that if the user attempts to access `Metadata` it will deadlock because it requires the initialization task to be finished. This PR changes this method so that it receives a `IMetadataSnapshotProvider` instance instead (which is inherited by `IMetadata`). It also changes the method to support async (returns `Task`).

A lot of methods from `ICluster` and `ISession` are moved to `Metadata`. This simplifies the driver's API because these methods and properties were just wrappers of those in `Metadata` so it makes more sense to be on the `IMetadata` interface only.

Also updated the upgrade guide so it's easier to get the overall picture of these API changes by reading the guide.


joao-r-reis:
This PR doesn't unify Session and Cluster yet but a lot of these changes were made with the unification in mind which will come in a following PR.

I'm opening this PR before fixing the test code in order to allow for an earlier review of the API changes/design @jorgebay

dvyskrebets:
Hi @joao-r-reis 
I'd like to ask you about this PR, are there any plans to finish this item?
There is an issue with connecting to cassandra in my team. Sometimes, for some reason, we see Cassandra.NoHostAvailableException and the code gets stuck in such state even if the cluster is actual available. We're trying to find the reason for the issue, thinking that it might be something related to how session is handled on the driver level. 
So if you have any advice on such problem or a timeline for this PR it'd be great to get help from you (I guess this PR relates to described issue though I'm not 100% sure)
Thanks

joao-r-reis:
This was being worked on when we had a plan to release a new major version but those plans have for the most part been abandoned for now so there is no timeline for this PR.

When the cluster initialization fails, it is recommended to close it and create the cluster object again (from the builder that you already have). If you are building the cluster object and then registering it as a singleton for example, the initialization will happen when the first session is created (likely through the DI container) and if that fails then there is a fail case where all subsequent session creations will fail from that same cluster object.

You should create the session during app startup and only then you register it as a singleton (if it fails you let the app crash and the app launch should be retried by whatever deployment tooling you're using as an example, or you can implement retries of this first session creation during startup), or change the DI configuration so that if the first session creation fails then a new cluster object is created.

dvyskrebets:
@joao-r-reis Thanks for quick response. Can I ask you quick question though a bit off topic. Sometimes our team experience issue with connection pool. No new connections opened. Any suggestions at what direction should we dig?

joao-r-reis:
That's too vague, I'd have to take a look at the driver logs (at least INFO level)

dvyskrebets:
this is cassandra in kubernetes cluster. load balancer used to expose private id to external service (web app in another cluster). at this point, I can't say for 100% how it should be configured correctly. 
I believe this setup is used for cassandra - https://github.com/k8ssandra/k8ssandra/

PR #531, opened 2020-07-24
Remove explicit reference to System.Net.Http

verdie-g:
The assembly is only used in CloudMetadataService for .NET Standard
builds but System.Net.Http is built-in in .NET Standard and doesn't
need an explicit reference.

Solution mixing nuget and built-in references to System.Net.Http leads
to runtime issues that we are trying to solve by removing the nuget
references.

https://github.com/dotnet/runtime/issues/26131#issuecomment-388199236
> In most cases, we don't advise people use the separate System.Net.Http
> NuGet package anymore

### DIFF
--- a/src/Cassandra/Cassandra.csproj
+++ b/src/Cassandra/Cassandra.csproj
@@ -35,7 +35,6 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
-    <PackageReference Include="System.Net.Http" Version="4.3.2" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="[4.6.0,5.0)" />
   </ItemGroup>


joao-r-reis:
Hmm I think that we were required to add it for `netstandard1.5` if I'm not mistaken but since we dropped support for that version we can probably remove it. 

I want to run some more tests to make sure nothing breaks and I'll try to write an implementation of `SimulacronManager` that doesn't depend on `System.Net.Http` so that we can remove that reference from the test project as well, otherwise we aren't really testing the scenario that we want.

I should have an update about this soon. Is this urgent or can you wait for the next release? (The next release should be out around end of August / early September).

verdie-g:
Thanks for the quick response. We will be using an internal fork until the next release as this issue is in our critical path.

The jenkins pipeline was not successful. Is it something I should worry about?

joao-r-reis:
No, the first run was green but this one ran the nightly suite and it had a couple of test failures which I will investigate later, they are not related to this change.

We can get out a patch release with just this change if this is blocking you, I'll run the tests that I need to run after the weekend and after that I will merge this PR and push a new patch release.

verdie-g:
I would be, indeed, very interested in a quick release with this change, thanks :)

joao-r-reis:
I opened another PR https://github.com/datastax/csharp-driver/pull/532 with all the necessary changes to remove the reference and I'm currently running tests using that branch. If everything goes smoothly I should be able to push a 3.15.1 release still today.

Sorry for not taking your commit here, I removed the reference as part of f48568d30a1cf08ee3b0251e990608e7e5fc2064

joao-r-reis:
Just fyi 3.15.1 has been published to nuget.org: https://www.nuget.org/packages/CassandraCSharpDriver/3.15.1

Mailing list thread: https://groups.google.com/a/lists.datastax.com/d/msg/csharp-driver-user/jUbrVTAKqQA/-rqofs1iCAAJ

verdie-g:
Thank you :)

PR #534, opened 2020-08-07
CSHARP-700 Add UDT and Tuple support with DSE 6.8 Graph

joao-r-reis:
The tuple test is failing with a weird error, I think it might be a server bug related to how `Timestamp` is handled within tuples. If I get confirmation that this is a server bug, I'll change the test or mark it as ignored and add another tuple test without `Timestamp`.

### DIFF
--- a/src/Cassandra/Serialization/GenericSerializer.cs
+++ b/src/Cassandra/Serialization/GenericSerializer.cs
@@ -335,6 +335,16 @@ namespace Cassandra.Serialization
             return true;
         }
 
+        public UdtMap GetUdtMapByName(string name)
+        {
+            return _udtSerializer.GetUdtMap(name);
+        }
+
+        public UdtMap GetUdtMapByType(Type type)
+        {
+            return _udtSerializer.GetUdtMap(type);
+        }
+
         public byte[] Serialize(ProtocolVersion version, object value)
         {
             if (value == Unset.Value)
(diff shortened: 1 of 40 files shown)


src/Cassandra/Serialization/Graph/GraphSON3/Dse/UdtGraphSONDeserializer.cs
```diff
@@ -0,0 +1,137 @@
...
+                    || type.GetGenericTypeDefinition() == typeof(Dictionary<,>)
+                    || type.GetGenericTypeDefinition() == typeof(IDictionary<,>)))
+            {
+                targetTypeIsDictionary = true;
+                var genericArgs = type.GetTypeInfo().GetGenericArguments();
+                if (genericArgs[0] != typeof(string))
+                {
+                    throw new InvalidOperationException(
+                        "Deserializing UDT to Dictionary is only supported when the dictionary key is of type \"string\".");
+                }
```
jorgebay, line 51: Good idea to support dictionaries and limit them to `Dictionary<string, T>`

src/Cassandra/Serialization/GenericSerializer.cs
```diff
@@ -81,6 +82,13 @@ internal GenericSerializer(IEnumerable<ITypeSerializer> typeSerializers = null)
             _udtSerializer.SetChildSerializer(this);
             InitDefaultTypes();
             InitTypeAdapters();
+            _defaultGraphTypes = new Dictionary<ColumnTypeCode, Func<IColumnInfo, Type>>(_defaultTypes)
```
jorgebay, line 85: Defaults keep growing 😄

PR #538, opened 2020-08-31
CSHARP-902 fix multiple protocol event handler invocation

joao-r-reis:


src/Cassandra/ProtocolEvents/ProtocolEventDebouncer.cs
```diff
@@ -240,28 +243,36 @@ private static async Task ProcessQueue(EventQueue queue)
...
                     catch (Exception ex)
                     {
                         foreach (var cb in keyspace.Value.Events.Select(e => e.Callback).Where(e => e != null))
                         {
-                            cb.TrySetException(ex);
+                            Task.Run(() => cb.TrySetException(ex)).Forget();
                         }
                     }
+
+                    continue;
```
jorgebay, line 257: Oh, I was looking for the fix and checked a few times!

jorgebay (approved):

src/Cassandra/ProtocolEvents/ProtocolEventDebouncer.cs
```diff
@@ -218,14 +215,20 @@ private static async Task ProcessQueue(EventQueue queue)
                     await queue.MainEvent.Handler().ConfigureAwait(false);
                     foreach (var cb in queue.Callbacks)
                     {
-                        cb?.TrySetResult(true);
+                        if (cb != null)
+                        {
+                            Task.Run(() => cb.TrySetResult(true)).Forget();
```
jorgebay, line 220: In the future, it could be interesting to create a `TaskCompletionSource` that sets results / exceptions in a new thread, somethings like `TaskCompletionAsyncSource` :)

joao-r-reis: That's a good idea, it's probably something that I should have done by now 😅 I'll definitely do that in the near future

PR #540, opened 2020-09-01
CSHARP-906 Support dse auth transitional mode by default

joao-r-reis:


src/Cassandra.IntegrationTests/Core/TransitionalAuthenticationTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                    300);
+            }
+        }
+
+        [OneTimeSetUp]
+        public void OneTimeSetUp()
+        {
+            Diagnostics.CassandraTraceSwitch.Level = TraceLevel.Info;
+            _testClusterForAuthTesting = GetTestCcmClusterForAuthTests();
+            RetryUntilClusterAuthHealthy(_testClusterForAuthTesting);
```
jorgebay, line 64: Is this related to a failure? I didn't get it on the Node.js driver with `-Dcassandra.superuser_setup_delay_ms=0`.

src/Cassandra.IntegrationTests/Core/TransitionalAuthenticationTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                                 .Build())
+            {
+                var session = cluster.Connect();
+                var rs = session.Execute("SELECT * FROM system.local");
+                Assert.Greater(rs.Count(), 0);
+            }
+        }
+
+        [Test]
+        public void WrongCredentials_DsePlainText_AuthSuccess()
```
jorgebay, line 127: TIL that with transitional mode any credentials would do!

jorgebay (approved):
lgtm!

src/Cassandra.IntegrationTests/Core/TransitionalAuthenticationTests.cs
joao-r-reis: Tbh I'm not sure why this is here, I copied this from the pre exiting authentication tests... I can give it a go and just remove this

jorgebay: bah, whatever you prefer, I think this was needed before "delay" setting to 0ms but its not anymore.

joao-r-reis: It seems that it wasn't needed indeed, tests are green without it :+1:

PR #544, opened 2020-10-20
CSHARP-943 CSHARP-942 Fix ArgumentException on TokenMap ctor

joao-r-reis:
Also add logging to control connection failures.


jorgebay (approved):
Nice findings and good coverage :shipit:

PR #549, opened 2021-02-16
allow AddPropertyMapping to be called directly

dylandhall:
Very small PR for allowing consumers to add properties with a reflected `PropertyInfo` - please see [issue](https://datastax-oss.atlassian.net/browse/CSHARP-949?atlOrigin=eyJpIjoiYjBjZTFkM2I0NGQ4NGM2Mjg3MzcwZTExMmExNmFiYTMiLCJwIjoiaiJ9) for details

PR #551, opened 2021-03-23
Removed constraint on System.Threading.Tasks.Dataflow version

Lirrik:
This will allow users of the driver to upgrade to System.Threading.Tasks.Dataflow v5 and beyond, fixes https://datastax-oss.atlassian.net/browse/CSHARP-946

### DIFF
--- a/src/Cassandra/Cassandra.csproj
+++ b/src/Cassandra/Cassandra.csproj
@@ -42,7 +42,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <PackageReference Include="System.Management" Version="4.7.0" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" />
-    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="[4.6.0,5.0)" />
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.6.0" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net452' ">
     <Reference Include="System.Data" />


joao-r-reis:
Hey, we don't have a release plan for the next release yet, is this blocking you? I remember that we talked about this a couple of months ago:

https://community.datastax.com/questions/9548/c-driver-nuget-package-prevents-upgrade-for-system.html

turowicz:
@joao-r-reis yes this is a bit of an issue since our company has a strict nuget dependencies policy.

It would be amazing if you could merge that in.

turowicz:
cc @Lirrik

turowicz:
cc @pabloordonez

joao-r-reis:
> 
> 
> @joao-r-reis yes this is a bit of an issue since our company has a strict nuget dependencies policy.
> 
> It would be amazing if you could merge that in.

Merging it is not an issue, the issue is the release date since we don't have one at the moment. If you can build the driver yourselves then you don't have to be blocked until this is released.

turowicz:
@joao-r-reis if you merge it in, then perhaps we will not have to build a fork?

joao-r-reis:
Yeah if you're able to build the driver then I'm happy to merge this in and keep it in `master` until the next release.

joao-r-reis:
@Lirrik please sign our CLA if you haven't already: https://cla.datastax.com/

Lirrik:
> Hey, we don't have a release plan for the next release yet, is this blocking you? I remember that we talked about this a couple of months ago:
> 
> https://community.datastax.com/questions/9548/c-driver-nuget-package-prevents-upgrade-for-system.html

@joao-r-reis yeah, it's been a while and this has been quite an annoying block for one of our teams who are using this driver. I saw a recent #549 where you also had mentioned there was no plan for a release yet, but then there was a release 11 days ago. I would just like to make sure this change makes next release whenever that might be and finally unblocks us.

halex2005:
Please, release this merge as PATCH release. Or release it at least as beta semantic version release.
It blocks me from moving to .NET 5 due to nuget errors.

Telavian:
I agree. This is also stopping me.

halex2005:
Actually, I've found temporary workaround - I just include dataflow package near CassandraCSharpDriver like this:

```
<PackageReference Include="CassandraCSharpDriver" Version="3.16.3" />
<PackageReference Include="System.Threading.Tasks.Dataflow" Version="5.0.0" />
```

nuget emits warning in this case, not an error (that's annoying, but it works).

Still waiting new version of CassandraCSharpDriver :)

Telavian:
How does this work when the package has a strict < 5.0.0 dependency?
System.Threading.Tasks.Dataflow (>= 4.6.0 && < 5.0.0)

joao-r-reis:
I've tried this out before and I can confirm that it does indeed work despite the warning: https://community.datastax.com/answers/9581/view.html

halex2005:
It works because nuget will select version that specified in csproj instead of transitive dependency version.

So, if csproj includes CassandraCSharpDriver (which has Dataflow v4 restiction) and Dataflow (of version 5)
then nuget will select Dataflow v5 because Dataflow v5 is more selective (it specified in concrete csproj that uses Dataflow v5).

PR #552, opened 2021-06-21
Make csharp-driver conform to common Queryable data implementation semantics

CassOnMars:
This could be theoretically split into two PRs if needed, but they were encountered at the same time when using the library as someone accustomed to other IQueryProvider implementations :

- Updates AttributeBasedTypeDefinition to permit System.ComponentModel.Annotations attributes when the driver-specific attributes are omitted, enabling consistent POCO mapping when swapping between IQueryable contexts, such as EntityFramework) 
- Removes InvalidOperationException when GetEnumerator() is called, along with relevant changes needed to support base Queryable methods in CqlExpressionVisitor, which again helps support consistent behaviors between injected IQueryable contexts.

### DIFF
--- a/src/Cassandra/Cassandra.csproj
+++ b/src/Cassandra/Cassandra.csproj
@@ -40,6 +40,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.7.0" />
     <PackageReference Include="System.Management" Version="4.7.0" />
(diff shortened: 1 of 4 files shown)


joao-r-reis:
Hi, thanks for your interest in contributing to this project. Unfortunately, the decision to throw `InvalidOperationException` at those places was made to prevent users from running into issues where their code is accidentally executing CQL queries. We really do not have an intention of supporting those `IQueryable` use cases.

The goal of our `LINQ` component is to enable users to write queries in a more user friendly way but it is not meant to be used with `EntitiyFramework` or other similar libraries that support `IQueryProvider` / `IQueryable` implementations.

The best way to support `EntityFramework` would be to create a separate nuget package dedicated to provide that support and potentially make the necessary changes on the API of the driver to expose the necessary extension points.

Since we don't have the goal of supporting `EntityFramework` at this time, the `AttributeBasedTypeDefinition` changes are also unnecessary.

PR #555, opened 2021-11-24
Turkish lower I char problem when using int datatype

kartegagithub:
GetTypeString return ınt instead of int.

### DIFF
--- a/src/Cassandra/Mapping/Statements/CqlGenerator.cs
+++ b/src/Cassandra/Mapping/Statements/CqlGenerator.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 //      Copyright (C) DataStax Inc.
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


joao-r-reis:
Hello, I need some context here, can you provide a code example that reproduces the issue?

kartegagithub:
Hi
We are trying to create an entity and then trying to create table in CassandraDB, "cassandra line 1:156 no viable alternative at character 'ı'" exception is thrown;
image

I checked the Cassandra source code and found out GetTypeString return ınt instead of int. We use Turkish chars i and ı and ı.

joao-r-reis:
I am able to reproduce the issue but the correct fix isn't to add a `Replace` because that doesn't fix other potential issues with other cultures and it's not very efficient. I will change all `ToLower` and `ToUpper` calls to `ToLowerInvariant` and `ToUpperInvariant` which shoud fix this. Follow [CSHARP-960](https://datastax-oss.atlassian.net/browse/CSHARP-960) for updates.

Note that there is no guarantee that you will not run into other issues after this is fixed since we don't test the driver with other cultures. It's always best to try and use `InvariantCulture` on your App to avoid these issues.

joao-r-reis:
Closing this in favor of #556

joao-r-reis:
Version 3.17.1 is now available with the fix.

PR #567, opened 2022-07-28
CSHARP-975 fix: add check timestamp that cassandra gives if less then…

okanlodos:
this is for https://datastax-oss.atlassian.net/browse/CSHARP-975.

i added the min value as long because DateTimeOffset.ToUnixTimeMilliseconds() is not available in .net=4.5.2

### DIFF
--- a/src/Cassandra/Serialization/Primitive/DateTimeOffsetSerializer.cs
+++ b/src/Cassandra/Serialization/Primitive/DateTimeOffsetSerializer.cs
@@ -28,6 +28,7 @@ namespace Cassandra.Serialization.Primitive
         internal static DateTimeOffset Deserialize(byte[] buffer, int offset)
         {
             var milliseconds = BeConverter.ToInt64(buffer, offset);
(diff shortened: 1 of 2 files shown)


joao-r-reis:
Hi, thanks for the issue report and the PR! It's still not clear to me what the best solution for this problem is so I'll keep investigating this. 

I don't think the solution implemented in this PR is good enough because it basically modifies the data that is being read so the user will not actually be able to read the correct data from the database.

Since the type itself does not support the values that are being read from the database it seems to me that we have 2 options:

1. Introduce a breaking change and change the default C# type that is associated with this C* type
2. Introduce a custom serializer that users can use if they run into this issue (and potentially set this as the default in the next major version of the driver, whenever that happens)

okanlodos:
i guess, the best way to fix it is fix the datetime lib in java :D
even if we fix it, it won't work old incorrect data.

we are using csharpdriver with json select, and then deserialize to entity object with external json lib(newtonsoft etc.).
therefore we prevented that DateTimeSerialization throws exception by not using.
maybe, new serializer that handle this issue would be good.

if i have a time, i am gonna look at what i can do better on this problem than this.

PR #568, opened 2022-08-16
CSHARP-977 Avoid memory allocations for ToSingle() and ToDouble() conversions

marcin-krystianc:
This change reduces GC pressure by avoiding temporary arrays allocation for `ToDouble` and `ToSingle` conversions.

We've discovered this problem because we use wide tables, i.e. ~90 columns of mostly doubles. Our benchmarks show that driver's performance is greatly improved by this change. For the "workstation GC", we see a stunning 2x-2.5x speed-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # ChangeLog - DataStax C# Driver
 
+## Unreleased
+
+TBD
+
+### Bug fixes
+
+*   [[CSHARP-974](https://datastax-oss.atlassian.net/browse/CSHARP-974)] Race condition in Hashed Wheel Timer causes unhandled exception
+
 ## 3.18.0
 
 2022-07-04
--- a/src/Cassandra.Tests/SerializerTests.cs
+++ b/src/Cassandra.Tests/SerializerTests.cs
@@ -40,7 +40,6 @@ namespace Cassandra.Tests
         [Test]
         public void EncodeDecodeSingleValuesTest()
         {
-
             var initialValues = new object[]
             {
                 "utf8 text mañana",
@@ -49,10 +48,13 @@ namespace Cassandra.Tests
                 1234F,
                 1.14D,
                 double.MinValue,
+                float.MinValue,
                 -1.14,
                 0d,
                 double.MaxValue,
+                float.MaxValue,
                 double.NaN,
+                float.NaN,
                 1.01M,
                 72.727272727272727272727272727M,
                 -72.727272727272727272727272727M,
@@ -68,7 +70,7 @@ namespace Cassandra.Tests
                 true,
                 new byte[] {16},
                 Guid.NewGuid(),
-                Guid.NewGuid()
+                Guid.NewGuid(),
             };
             foreach (var version in _protocolVersions)
             {
--- a/src/Cassandra/BeConverter.cs
+++ b/src/Cassandra/BeConverter.cs
@@ -157,6 +157,7 @@ namespace Cassandra
             Swap(value, offset + 3, offset + 4);
             var result =  BitConverter.ToDouble(value, offset);
             
+            // Restore the original order
             Swap(value, offset + 0, offset + 7);
             Swap(value, offset + 1, offset + 6);
             Swap(value, offset + 2, offset + 5);
@@ -179,8 +180,10 @@ namespace Cassandra
             Swap(value, offset + 1, offset + 2);
             var result = BitConverter.ToSingle(value, offset);
             
+            // Restore the original order
             Swap(value, offset + 0, offset + 3);
             Swap(value, offset + 1, offset + 2);
+
             return result;
         }
 
--- a/src/Cassandra/BeConverter.cs
+++ b/src/Cassandra/BeConverter.cs
@@ -15,7 +15,7 @@
 //
 
 using System;
-using System.Runtime.CompilerServices;
+using System.Buffers;
 
 namespace Cassandra
 {
@@ -149,8 +149,20 @@ namespace Cassandra
             {
                 return BitConverter.ToDouble(value, offset);
             }
-            
-            return BitConverter.Int64BitsToDouble(ToInt64(value, offset));
+
+            //Invert the first 8 bytes, starting from offset
+            var rentedArray = ArrayPool<byte>.Shared.Rent(8);
+            rentedArray[7] = value[offset + 0];
+            rentedArray[6] = value[offset + 1];
+            rentedArray[5] = value[offset + 2];
+            rentedArray[4] = value[offset + 3];
+            rentedArray[3] = value[offset + 4];
+            rentedArray[2] = value[offset + 5];
+            rentedArray[1] = value[offset + 6];
+            rentedArray[0] = value[offset + 7];
+            var result = BitConverter.ToDouble(rentedArray, 0);
+            ArrayPool<byte>.Shared.Return(rentedArray);
+            return result;
         }
 
         /// <summary>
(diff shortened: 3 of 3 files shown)


joao-r-reis:
Hi, thanks for the PR. Unfortunately we aren't planning on releasing a new version of the driver in the near future so just a heads up that we might take a while until we take a look at this PR and merge it.

marcin-krystianc:
@joao-r-reis Thank you for the heads up. Can you say when approximately we can expect the next release? Is there a chance for a release happening by the end of this year?

joao-r-reis:
Yes, we are planning at least one release before the end of this year.

marcin-krystianc:
Actually, there are a few more options to run the conversion which are even more efficient than the `ArrayPool` approach:
- swap bytes in the source array (option 1):
image
- swap bytes in the source array, convert, and swap it again so the buffer is untouched (if the above is not an option) (option 2):
image
- Let the driver use unsafe code (`AllowUnsafeBlocks`=`true`) so it is possible to convert to double in place without any byte swapping at all (option 3)

All of these are significantly faster than the `ArrayPool` approach. Please let me know which one you would be willing to accept.  I think option 2 is the least controversial but is still significantly faster than the current ArrayPool approach.

M0n7y5:
Any update about this?

joao-r-reis:
No updates, as I mentioned earlier we plan to take a look at open PRs and get a release out before the end of the year so there's unlikely to be any update on this until late October at the very least.

joao-r-reis:
Option 2 is definitely the one that is the safest so if you don't have any objections you can modify this PR with that approach instead and I'll review it.

Also, please sign our CLA if you haven't already: https://cla.datastax.com

marcin-krystianc:
> Option 2 is definitely the one that is the safest so if you don't have any objections you can modify this PR with that approach instead and I'll review it.

I've found another possibility (https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.int64bitstodouble?view=net-6.0) that doesn't require unsafe code but it is more performant than option 2.

joao-r-reis:
> I've found another possibility (https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.int64bitstodouble?view=net-6.0) that doesn't require unsafe code but it is more performant than option 2.

So your suggestion is to use the `ToInt64` method of the same class first and then calling `BitConverter.Int64BitsToDouble(Int64)` on that `long`? And for float/single you would use `ToInt32` and `BitConverter.Int32BitsToSingle`?

This is a very interesting idea, I'm also ok with this one but I'd ask you to implement some unit tests in the `Cassandra.Tests.SerializerTests` class that use `double` and `float` values to get some test coverage on these 2 methods that are going to be changed in this PR.

marcin-krystianc:
> So your suggestion is to use the `ToInt64` method of the same class first and then calling `BitConverter.Int64BitsToDouble(Int64)` on that `long`? 

I didn't thought about using `ToInt64`, I coded a manual conversion instead - I'll try compare both approaches and if there is no performance difference I'll use to `ToInt64` approach as it is simpler.

> And for float/single you would use `ToInt32` and `BitConverter.Int32BitsToSingle`?

Unfortunately `BitConverter.Int32BitsToSingle` is not available  for .NET Framework / .NET Standard so we need to use byte swapping approach.

> This is a very interesting idea, I'm also ok with this one but I'd ask you to implement some unit tests in the Cassandra.Tests.SerializerTests class that use double and float values to get some test coverage on these 2 methods that are going to be changed in this PR.

Good idea, I'll add tests.

Telavian:
Would BinaryPrimitives from System.Buffers.Binary work?
That is available in .Net Framework and .Net Core.

joao-r-reis:
> Would BinaryPrimitives from System.Buffers.Binary work? That is available in .Net Framework and .Net Core.

I believe that would require adding the `System.Memory` package dependency since this driver is still targeting .NET Standard 2.0 and .NET Framework 4.5.2 and we should try to avoid adding extra dependencies that can cause version conflicts in the applications that use this driver if possible.

marcin-krystianc:
> > So your suggestion is to use the `ToInt64` method of the same class first and then calling `BitConverter.Int64BitsToDouble(Int64)` on that `long`?
> 
> I didn't thought about using `ToInt64`, I coded a manual conversion instead - I'll try compare both approaches and if there is no performance difference I'll use to `ToInt64` approach as it is simpler.

I've updated my code to use `ToInt64` as it is simpler and the perf is the same.

Regarding the tests, the `EncodeDecodeSingleValuesTest` seems to be already covering methods which I'm updating.

joao-r-reis (approved):
👍 I'm dealing with some issues on our internal CI infrastructure as we prepare for the next C# release but I'll merge this PR as soon as

xsoheilalizadeh:
Do we have an ETA for the 3.19.0 release?

joao-r-reis:
> Do we have an ETA for the 3.19.0 release?

No specific date but we're trying to get the release out before the end of the year so mid December is the current target.

joao-r-reis:
I just realized that all our serializers are implemented in a thread safe way and this change effectively makes this particular serializer not thread safe as concurrent calls to `Deserialize` with the same buffer will cause issues. I'd prefer if we went with the first approach (`ArrayPool`) even though it is not as performant as just avoiding any new buffers completely. Sorry for not realizing this before.

We finally have things sorted out and we are ready to get a new release out so as soon as this is handled I will merge this PR.

joao-r-reis:
I cherry picked your initial commit on this PR to this branch https://github.com/datastax/csharp-driver/pull/575 so I can start testing it, let me know if you agree with this change and whether you want me to merge the new PR or if you want to revert your new commits on this PR and merge this one instead.

marcin-krystianc:
> I just realized that all our serializers are implemented in a thread safe way and this change effectively makes this particular serializer not thread safe as concurrent calls to Deserialize with the same buffer will cause issues. I'd prefer if we went with the first approach (ArrayPool) even though it is not as performant as just avoiding any new buffers completely. Sorry for not realizing this before.
Ah I see, fair enough. 

> I cherry picked your initial commit on this PR to this branch #575 so I can start testing it, let me know if you agree with this change and whether you want me to merge the new PR or if you want to revert your new commits on this PR and merge this one instead.

I'll revert changes on this branch.

marcin-krystianc:
@joao-r-reis I reverted changes on this branch so we are back to the `ArrayPool` approach. Another quick idea I've is to use `stackalloc byte[8]` instead of renting the array from `ArrayPool`. I'll measure whether it makes any difference.

marcin-krystianc:
> @joao-r-reis I reverted changes on this branch so we are back to the `ArrayPool` approach. Another quick idea I've is to use `stackalloc byte[8]` instead of renting the array from `ArrayPool`. I'll measure whether it makes any difference.

After some trying I thing that it is not actually possible to use stackalloc as it requires `Span<T>` based APIs and these are not available until `netstandard 2.1` / `netcore`

joao-r-reis:
We can keep the `BitConverter.Int64BitsToDouble(ToInt64(value, offset));` approach for `ToDouble` though, can you make that quick change?

marcin-krystianc:
> We can keep the `BitConverter.Int64BitsToDouble(ToInt64(value, offset));` approach for `ToDouble` though, can you make that quick change?

👍

joao-r-reis (approved):

joao-r-reis:
justfyi: [C# Driver 3.19.0 is now GA](https://groups.google.com/a/lists.datastax.com/g/csharp-driver-user/c/IuuCltBoQ8g).

marcin-krystianc:
@joao-r-reis FYI I've compared performance of versions 3.18.0 and 3.19.0 with a benchmark that mostly does double value reading. I've noticed about 50% improvement of the performance.

joao-r-reis:
Awesome! In our benchmark that doesn't focus on a particular type or request we also noticed an improvement of ~10% so I'm fairly confident that this release has performance benefits for everyone regardless of workload type. Great stuff!

PR #569, opened 2022-08-24
CSHARP-979 Improve data reading performance by inlining hot methods in RecyclableMemoryStream

marcin-krystianc:
As explained in the [Jira ticket](https://datastax-oss.atlassian.net/browse/CSHARP-979), these methods are executed a lot of times. After inlining them, I can see improvements of 3-5% in reading data. I know that it is a micro-optimisation that is rarely worth doing. However, in this case, it gives non-negligible performance improvements.

### DIFF
--- a/src/Cassandra/RecyclableMemoryStream.cs
+++ b/src/Cassandra/RecyclableMemoryStream.cs
@@ -20,6 +20,7 @@
 // THE SOFTWARE.
 // ---------------------------------------------------------------------
 
+using System.Runtime.CompilerServices;
 using Cassandra.Collections;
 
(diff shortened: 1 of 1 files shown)


joao-r-reis:
This one seems pretty straight forward, I don't have any objections to merging it in as is.
Have you signed our CLA? If you haven't yet, you can do it here: https://cla.datastax.com/

marcin-krystianc:
> Have you signed our CLA? If you haven't yet, you can do it here: https://cla.datastax.com/

Done

joao-r-reis (approved):

PR #581, opened 2022-12-23
CSHARP-989 Change return type of CqlFunction.WriteTime() to DateTimeOffset

bastianeicher:

PR #582, opened 2023-01-17
CSHARP-989 Fix writetime support in LINQ

joao-r-reis:
Integration test and fix on top of #581


joao-r-reis:
Just fyi @bastianeicher had to change the return type of the function to `long` because that's what the DB returns.

PR #584, opened 2023-03-30
Fixes for CqlFunction.WriteTime() issues

bastianeicher:
I ran into two issues using `CqlFunction.WriteTime()` in LINQ expressions:

When the value selected by `WRITETIME()` has never been set Cassandra returns `null`. This causes the current implementation to fail with `NullReferenceException` when deserializing the result.  
The first commit in this pRR should hopefully resolve this, by making the result of `CqlFunction.WriteTime()` nullable.

When using `CqlFunction.WriteTime()` in a `.Select()` clause _before_ a `.Where()` clause everything works fine. But when putting the `.Where()` clause first, the otherwise identical request fails with `InvalidOperationException`.  
The second commit in this PR adjusts an integration test to reproduce this error. Don't know how to go about fixing it yet, though.


bastianeicher:
> Can you change the test so that it covers both cases?

Done. 👍 Accidentally closed this PR, so it's in #585 now.

bastianeicher:
> This is expected, `.Where().Select()` will cause the driver to evaluate the `Select()` in memory after the result has been received while `.Select().Where()` will cause the driver to convert the select statement to CQL before sending the statement to the server.

I see. 💡  I actually find that quite surprising. I had assumed that the driver would try to convert everything before `.Execute()` into CQL and treat everything after as a regular `IEnumerable<>` with in-memory filtering and mapping.

Would it perhaps be possible to change the driver's behavior in this way?

Since `.Where()` clauses often just match against a single (primary key) column, I often create extension methods like this, to avoid having to repeat the same expression:

```csharp
public static CqlQuery<MyTable> WhereId(this CqlQuery<MyTable> query, string id) =>
  query.Where(x => x.Id == id);
```

But this wouldn't work with `myTable.Select(x => new {x.SomeColumn, x.OtherColumn}).WhereId(id)`, since the type is anonymous after `.Select()`.

joao-r-reis:
> Would it perhaps be possible to change the driver's behavior in this way?

Not without a breaking change, maybe we could think of changing this for a potential upcoming 4.0 release but there's a decent chance this would break several existing applications.

I'm not 100% sure if this was the intended behavior when this was designed several years ago but it's too late to change it now unless it's part of a bigger overhaul of this API in a new major release.

bastianeicher:
> Not without a breaking change, maybe we could think of changing this for a potential upcoming 4.0 release but there's a decent chance this would break several existing applications.
> 
> I'm not 100% sure if this was the intended behavior when this was designed several years ago but it's too late to change it now unless it's part of a bigger overhaul of this API in a new major release.

Perhaps an optional parameter for `.Execute()`, like `bool convertToCqlAggressively`, could be used to opt-in to the new behavior while keeping it unchanged for existing consumers.

I could imagine this would still be a pretty big change of course, so just something for the wishlist. 😊

PR #585, opened 2023-03-30
CSHARP-992 Support null writetime values in LINQ

bastianeicher:
When the value selected by WRITETIME() has never been set Cassandra returns null. This causes the current implementation to fail with NullReferenceException when deserializing the result.

This MR resolves this, by making the result of CqlFunction.WriteTime() nullable.

### DIFF
--- a/src/Cassandra.IntegrationTests/Linq/LinqRealClusterTests.cs
+++ b/src/Cassandra.IntegrationTests/Linq/LinqRealClusterTests.cs
@@ -368,6 +368,31 @@ namespace Cassandra.IntegrationTests.Linq
             Assert.Less(recordsArr[0].wt2, nowPlus5Minutes);
         }
 
+        [Test]
+        public void Linq_Writetime_Returns_Null()
+        {
+            var writetimeEntityTableName = "writetime_entities";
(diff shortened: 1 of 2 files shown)


src/Cassandra.IntegrationTests/Linq/LinqRealClusterTests.cs
```diff
@@ -368,6 +368,31 @@ public void Linq_Writetime_Returns_ValidResult()
...
+        {
+            var writetimeEntityTableName = "writetime_entities";
+            Session.Execute($"CREATE TABLE IF NOT EXISTS {writetimeEntityTableName} (id uuid primary key, propertyint int, propertystring text)");
+
+            var table = new Table<WritetimeEntity>(Session, new MappingConfiguration().Define(
+                new Map<WritetimeEntity>().TableName(writetimeEntityTableName)));
+            var id = Guid.NewGuid();
+            var writetimeEntity = new WritetimeEntity()
+            {
+                Id = id,
```
joao-r-reis, line 382: Change the type of `WritetimeEntity.PropertyInt` to `int?` otherwise the driver will send a `0` and `writetime()` will return a valid timestamp (which is why this test is failing atm).

bastianeicher: Changed, thanks 👍

PR #587, opened 2023-04-04
Add Guid operators for CqlFunction

xsoheilalizadeh:
I'm working on improving the infrastructure, meanwhile, I noticed builds are failing on .NET 7 because the operators are overrided for [Guid](https://github.com/dotnet/runtime/blame/v7.0.4/src/libraries/System.Private.CoreLib/src/System/Guid.cs).

The build fails with the following error:
```
Ambiguous invocation:   
bool operator <=(object, Cassandra.Data.Linq.CqlFunction) (in class CqlFunction)   
bool operator <=(System.Guid, System.Guid) (in struct Guid) match
```


joao-r-reis:
> I noticed builds are failing on .NET 7

Building the driver or building an app with the driver installed?

xsoheilalizadeh:
Both

joao-r-reis:
> I'm working on improving the infrastructure

Can you clarify what you are working on? Deep changes to the driver should be discussed beforehand.

> Both

I just installed .NET 7 and I can build the driver just fine, you probably need to install an earlier version of the SDK maybe? Can you tell me which sdk versions you have installed? I also created a new Console App that targets .NET 7, installed the driver and I can build this app just fine too.

xsoheilalizadeh:
> Can you clarify what you are working on? Deep changes to the driver should be discussed beforehand.

Sure, I'm trying to drop the old .NET Core version since they are deprecated and cleaning the csproj files.

> I just installed .NET 7 and I can build the driver just fine, you probably need to install an earlier version of the SDK maybe? Can you tell me which sdk versions you have installed? I also created a new Console App that targets .NET 7, installed the driver and I can build this app just fine too.

If you set the driver to build on .NET 7  these tests would have the error that I mentioned above. Same issue when using the driver as a package and trying to use operators on MaxTimeUuid/MinTimeUuid methods
https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.Tests/Mapping/Linq/LinqToCqlFunctionTests.cs#L39-L48

joao-r-reis:
Also, please sign our CLA: https://cla.datastax.com/

xsoheilalizadeh:
> Keep in mind that we will not change the targets of the driver project itself, we will keep targeting net452;netstandard2.0 for the driver package.

Yes, I'm aware of this. 

> Because of the backend work that this requires I think it would be best if you held off on making these changes for now.

Thanks, I will wait for the update.

joao-r-reis:
Sorry for the huge delay but we're finally upgrading the .NET versions on our test projects and build infrastructure and I've confirmed this fix works 👍

joao-r-reis (approved):
I'll just let CI run and will merge afterwards.

PR #592, opened 2023-08-23
CSHARP-995 Respect .SetIdempotence() on LINQ queries

bastianeicher:
Currently the value of .IsIdempotent is not copied from LINQ statements to generated BoundStatements. This makes .SetIdempotence() ineffective on LINQ statements, preventing things like speculative execution from working.

This PR copies the idempotence value along side other properties that are already copied, like consistency level and retry policy.

### DIFF
--- a/src/Cassandra/Data/Linq/CqlQueryExtensions.cs
+++ b/src/Cassandra/Data/Linq/CqlQueryExtensions.cs
@@ -29,7 +29,9 @@ namespace Cassandra.Data.Linq
                .SetPageSize(src.PageSize)
                .SetPagingState(src.PagingState)
                .SetRetryPolicy(src.RetryPolicy)
-               .SetAutoPage(src.AutoPage);
+               .SetAutoPage(src.AutoPage)
+               .SetReadTimeoutMillis(src.ReadTimeoutMillis)
+               .SetOutgoingPayload(src.OutgoingPayload);
             if (src.SerialConsistencyLevel != ConsistencyLevel.Any)
             {
                 dst.SetSerialConsistencyLevel(src.SerialConsistencyLevel);
@@ -37,6 +39,10 @@ namespace Cassandra.Data.Linq
             if (src.IsIdempotent.HasValue)
             {
                dst.SetIdempotence(src.IsIdempotent.Value);
+            }
+            if (src.Timestamp.HasValue)
+            {
+               dst.SetTimestamp(src.Timestamp.Value);
             }
         }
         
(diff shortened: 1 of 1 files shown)


joao-r-reis:
Hi, thanks for the PR! Nice catch, it looks like there's a couple other properties that are not being copied. I did a quick check and it seems that the following are missing as well, can you add them?

 `ReadTimeoutMillis`, `OutgoingPayload`, `Timestamp`

bastianeicher:
I just discovered `CqlQueryOptions.CopyOptionsToStatement()`, which also seems to copy values from LINQ statements.

The following properties are currently copied by both:

- `IsTracing`
- `ConsistencyLevel`
- `SerialConsistencyLevel`
- `PageSize`
- `PagingState`
- `RetryPolicy`

The following is only copied by `CqlQueryExtensions.CopyQueryPropertiesTo()`:

- `AutoPage`

The following is only copied by `CqlQueryOptions.CopyOptionsToStatement()`:

- `Timestamp`

The following are copied by neither:

- `IsIdempotent`
- `ReadTimeoutMillis`
- `OutgoingPayload`

I'm not really sure which should be added where. :/

bastianeicher:
Thanks for the explanation.

I've added copying the remaining properties to `CqlQueryExtensions.CopyQueryPropertiesTo()` as you suggested.

joao-r-reis:
I'm just going to write a couple of tests to verify this before merging the PR.

Is this urgent? We weren't planning on doing a release soon but we can do a small patch release just for this depending on the urgency.

bastianeicher:
A patch release would be great, since this is blocking us from using speculative execution in a latency-sensitive use case.

joao-r-reis:
How are you setting idempotence with LINQ? Can you provide a small code snippet?

joao-r-reis:
While writing tests I realized you can't actually use the method `SetIdempotence()` with LINQ because then you can't call `Execute()` (it's an `IStatement` object) so the only supported way is to use execution profiles which should work correctly.

joao-r-reis:
Actually nevermind, set idempotence is not available in the execution profile so we're still missing a way to set idempotence in LINQ but I don't think this PR completely fixes the issue.

bastianeicher:
We currently do it like this:

```csharp
    var query = new Table<MyTable>(session).Where(x => x.Id == id);
    query.SetIdempotence(true);
    var result = await query.ExecuteAsync();
```

joao-r-reis:
Ah I see... I'll create a separate ticket to make sure it is possible to use `.SetIdempotence()` in a chaining way.

joao-r-reis:
I'm almost done writing these tests and I already found something that should be changed here. `SetTimestamp` is actually implemented in inserts/updates/deletes and it generates the `USING TIMESTAMP` clause in the actual query string instead of setting the protocol option. Adding it here in this PR is unnecessary as the only thing it is doing is adding it for SELECT statements which doesn't make sense anyway... 

Due to this, can you remove the `SetTimestamp` change from this PR? sorry about this.

bastianeicher:
> Due to this, can you remove the `SetTimestamp` change from this PR? sorry about this.

Sure, no problem. Done. 👍

PR #600, opened 2023-11-07
CSHARP-1000 Fix support for astra custom domain

joao-r-reis:


src/Cassandra/DataStax/Cloud/CustomCACertificateValidator.cs
```diff
@@ -145,6 +148,55 @@ public bool Validate(X509Certificate cert, X509Chain chain, SslPolicyErrors erro
...
+
+            return false;
+        }
+
+        private IEnumerable<string> GetSubjectAlternativeNames(X509Certificate2 cert)
+        {
+            var result = new List<string>();
+
+            var subjectAlternativeName = cert.Extensions.Cast<X509Extension>()
+                .Where(n => n.Oid.Value == "2.5.29.17") //n.Oid.FriendlyName=="Subject Alternative Name")
```
absurdfarce, line 177: Perhaps use the SubjectAlternateNameOid constant. you defined above here instead?

src/Cassandra/DataStax/Cloud/CustomCACertificateValidator.cs
```diff
@@ -145,6 +148,55 @@ public bool Validate(X509Certificate cert, X509Chain chain, SslPolicyErrors erro
...
+                .Select(n => n.Format(true))
+                .FirstOrDefault();
+
+            if (subjectAlternativeName != null)
+            {
+                var alternativeNames = subjectAlternativeName.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
+
+                foreach (var alternativeName in alternativeNames)
+                {
+                    var groups = Regex.Match(alternativeName, @"^(.*)=(.*)").Groups; // @"^DNS Name=(.*)").Groups;
```
absurdfarce, line 188: Is this intended to... replace the DnsNameRegex referenced above?  Seems like it's trying to do the same thing with a different regex.

absurdfarce (approved):
Certainly seems to address the problem as described.  A few nits based on my understanding of the code but I don't think either of them are problematic enough to hold things up if you need to get a release out.

It might be nice to have a test to confirm that SANs were being used for validation but, again, I don't think it's essential.

src/Cassandra/DataStax/Cloud/CustomCACertificateValidator.cs
joao-r-reis: Ahh I had to try multiple solutions and forgot about those constants... fixed

src/Cassandra/DataStax/Cloud/CustomCACertificateValidator.cs
joao-r-reis: Ahh I had to try multiple solutions and forgot about those constants... fixed

absurdfarce:
👍 to the most recent round of changes.  Nice tests @joao-r-reis !

PR #601, opened 2023-11-15
Remove PackageLicenseFile in favor of PackageLicenseExpression

leotsarev:
Nuget org documentation recommends using license file only if license is custom.
https://learn.microsoft.com/en-us/nuget/reference/nuspec#license
For standard licenses using SPDX:
- Allows automatic checks
- Clearer for users
- Smaller .nupkg


joao-r-reis:
Hi, thanks for the PR. I'm going to investigate this and I will get back to this once I have more information. 

In the meantime, have you signed our [CLA](https://cla.datastax.com/)?

joao-r-reis:
Unfortunately I have been advised to reject this PR because there are plans to donate this driver to the [ASF](https://www.apache.org/) and ASF requires their license file to be used in all packages.

It's possible that ASF might be ok with setting that identifier but it's not clear at the moment so I'll close this.

leotsarev:
Ok, this is not a big deal.

PR #602, opened 2023-11-22
CSHARP-1003 Fix support for Astra custom domains in linux environments

joao-r-reis:
#600 added support for Astra custom domains but that implementation only works in Windows environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # ChangeLog - DataStax C# Driver
 
+## 3.19.5
+
+2023-11-23
+
+### Bug fixes
+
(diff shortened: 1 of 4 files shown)


joao-r-reis:
> Makes sense given the platform-specific variation of this functionality. It's very strange to me that this behaviour varies _at all_ by platform... but that's different than the concern of this PR.

[There's a new "official" library that parses asn.1 stuff](https://stackoverflow.com/a/68608697/10896275) that would make it possible to stop relying on formatting it as a string and then parsing it manually but that library is only available in a newer .NET version so we can't use it on our driver as it stands.

PR #605, opened 2024-02-09
Fix Improper Handling of Exceptional Conditions Newtonsoft

lamrecognitions:
Newtonsoft.Json vulnerable to Insecure Defaults due to improper handling of expressions with high nesting level that lead to StackOverFlow exception or high CPU and RAM usage. Exploiting this vulnerability results in Denial Of Service (DoS). The serialization and deserialization path have different properties regarding the issue.

Deserializing methods (like `JsonConvert.DeserializeObject`) will process the input that results in burning the CPU, allocating memory, and consuming a thread of execution. Quite high nesting level `(>10kk, or 9.5MB of {a:{a:{... input)` is needed to achieve the latency over 10 seconds, depending on the hardware.

To mitigate the issue one either need to update Newtonsoft.Json to 13.0.1 or set `MaxDepth` parameter in the `JsonSerializerSettings`. This can be done globally with the following statement. After that the parsing of the nested input will fail fast with `Newtonsoft.Json.JsonReaderException`:
```
JsonConvert.DefaultSettings = () => new JsonSerializerSettings { MaxDepth = 128 };
```
Repro code:
```
//Create a string representation of an highly nested object (JSON serialized)
int nRep = 25000;
string json = string.Concat(Enumerable.Repeat("{a:", nRep)) + "1" +
 string.Concat(Enumerable.Repeat("}", nRep));

//Parse this object (leads to high CPU/RAM consumption)
var parsedJson = JsonConvert.DeserializeObject(json);

// Methods below all throw stack overflow with nRep around 20k and higher
// string a = parsedJson.ToString();
// string b = JsonConvert.SerializeObject(parsedJson);
```
The original statement about the problem only affecting IIS applications is misleading. Any application is affected, however the IIS has a behavior that stops restarting the instance after some time resulting in a harder-to-fix DoS.**

## Impact
CWE-755
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`
CVE-2024-21907

### DIFF
--- a/src/Cassandra/Cassandra.csproj
+++ b/src/Cassandra/Cassandra.csproj
@@ -40,7 +40,7 @@
     <PackageReference Include="K4os.Compression.LZ4" Version="1.1.11" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.0.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.Management" Version="4.7.0" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" />
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.6.0" />


joao-r-reis:
JSON Serialization and Deserialization is only used by the driver internally when decoding data from Datastax Graph or when decoding the `config.json` file that is contained in a Secure Connect Bundle downloaded through Datastax Astra.

We've done some analysis on this CVE before and we don't believe the driver is affected by it. And even if the driver is affected by it (which needs to be proven, we don't think it is) we would prefer to resolve it by adding something like the following to the json related calls:

```
new JsonSerializerSettings
    {
        MaxDepth = 2
    }
```

We'd prefer avoiding the Newtonsoft.Json upgrade as long as possible so that existing users don't feel forced to upgrade it if they don't have to. This is assuming that users can upgrade Newtonsoft.Json right now without issues (which we believe is the case).

If users can't upgrade Newtonsoft.Json on their apps due to an API compatibility with the version of Newtonsoft.Json that this driver is using then that would be a different matter potentially.

PR #608, opened 2024-05-21
[CSHARP-969] - OpenTelemetry session level tracing

simaoribeiro:
As discussed in the [OpenTelemetry proposal](https://github.com/datastax/csharp-driver/blob/master/proposals/open-telemetry/tracing-rfc.md), this PR includes the session level tracing following the OpenTelemetry standard.

@joao-r-reis There are two things that are included in the proposal that are missing in this:
- The `server.address` and `server.port` attributes because the Host information is only set in the Node level tracker methods
- The operation name, as this is a more of a big deal, as the `RequestTrackingInfo` is instantiated in the `RequestHandler` but the method invocation is made before in the `Session` class. One solution is to instantiate the tracker at the session level so we can access the method, but I wanted to see your opinion before going for that route in this PR.


simaoribeiro:
Hi @joao-r-reis , a bit later than I was expecting but here is the PR so we can proceed with the feature :)

joao-r-reis:
> The operation name, as this is a more of a big deal, as the RequestTrackingInfo is instantiated in the RequestHandler but the method invocation is made before in the Session class. One solution is to instantiate the tracker at the session level so we can access the method, but I wanted to see your opinion before going for that route in this PR.

We didn't really cover how we were going to implement this part and now that I'm thinking a bit more about it we would have to add a new `Enum` or just pass a string in the `TrackingInfo` but it's a bit awkward because it really only would be useful for this specific implementation of the `RequestTracker`. 

At the same time I'm thinking whether this will actually be useful and what would we lose if we just made the OpenTelemetry extension hardcode the operation name to `Execute`... If we were to implement the new "field" on the tracker info object at best we would have 2 operation names: `Execute` and `ExecuteAsync` but then there's some cases where LINQ and Mapper call `ExecuteAsync` even if the top level method is a synchronous one so I'm actually leaning towards just hardcoding a string like `CassandraRequest` as the operation name in the extension (it should be a non existing driver method to avoid confusion I think)... And maybe in the future we can revisit this and potentially add a new field to the tracking info object (which wouldn't be a breaking change).

---

I can't do a proper review of the PR right now but I'll try finding some time soon to look at it, thanks for the contribution again!

zehenrique:
hey @joao-r-reis how's it going?

do you think you will be able to look at this soon? This particular subject is becoming an issue and we need to understand how we will move forward in the near future.

thanks

joao-r-reis:
> hey @joao-r-reis how's it going?
> 
> do you think you will be able to look at this soon? This particular subject is becoming an issue and we need to understand how we will move forward in the near future.
> 
> thanks

Hey, I'll do a proper review today but it was my impression that this was still a draft and there was still work left to be done.

simaoribeiro:
Hi @joao-r-reis , it was my bad. I didn't realized it was marked as draft. I believe this pull request implements what we discussed in technical document minus the operation name.

Regarding that, I agree that we can change it to something more generic like the hardcoded string you mentioned above.

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -389,7 +389,7 @@ private void HandleRequestError(IRequestError error, Host host)
                     break;
             }
 
-            _requestObserver.OnRequestError(host, retryInformation.Reason, retryInformation.Decision.DecisionType);
+            _requestObserver.OnNodeRequestError(host, retryInformation.Reason, retryInformation.Decision.DecisionType);
```
joao-r-reis, line 392: By the way, you can add the `OnNodeSuccess` call at the top of the `private void HandleRowSetResult(Response response)` method and with this we will have all the request tracker methods implemented in the core driver 👍

joao-r-reis (commented):
Looking really good! Added a few comments.

I think initially we discussed that the `OnNodeSuccess` and `OnNodeFailure` methods were more complex to implement so they were going to be out of scope for now but it turns out they are super easy to implement after all. The `OnNodeFailure` is pretty much already implemented in your changes and I added a comment that says where you can add the `OnNodeSuccess` call.

joao-r-reis:
The build is still failing due to some analyzers that are not enabled by default when you build locally. If you want to build with all the analyzers enabled locally you can follow [these instructions](https://github.com/datastax/csharp-driver/blob/master/CONTRIBUTING.md#automated-checks---code-analyzers).

src/Cassandra.IntegrationTests/OpenTelemetry/OpenTelemetryTests.cs
```diff
@@ -0,0 +1,150 @@
...
+using Cassandra.OpenTelemetry;
+using Cassandra.Tests;
+using NUnit.Framework;
+using OpenTelemetry;
+using OpenTelemetry.Trace;
+
+namespace Cassandra.IntegrationTests.OpenTelemetry
+{
+    [Category(TestCategory.Short)]
+    public class OpenTelemetryTests : SharedClusterTest
```
joao-r-reis, line 14: Can we have a test that verifies that the activity context is correctly propagated? I don't think there is one at the moment that verifies this unless I missed it. If I'm not mistaken one of the use cases for this is to properly correlate these traces with (for example) an API call so the context needs to be properly propagated correct? Ideally we would have a test for statements, one for the mapper and one for LINQ. And each test should use both sync and async variants of the execute methods.

We should also test that the context propagation works correctly with auto paging, although I'm not 100% clear what behavior we want in this case. I'm assuming the context we want is the one that is active the moment where the enumerator pages the next page so the current changes might already do that but we need a test for it.

simaoribeiro:
Hi @joao-r-reis , back at it again after some time off. Will start working on your review

src/Extensions/Cassandra.OpenTelemetry/CassandraInstrumentation.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.Reflection;
+
+namespace Cassandra.OpenTelemetry
+{
+    public static class CassandraInstrumentation
+    {
+        internal static readonly AssemblyName AssemblyName = typeof(CassandraInstrumentation).Assembly.GetName();
+        internal static readonly string Version = AssemblyName.Version.ToString();
+        public static readonly string ActivitySourceName = AssemblyName.Name;
```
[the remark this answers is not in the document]
simaoribeiro: Are we expecting it to change without being included in a major version? 🤔 

I followed the same logic as the OpenTelemetry contribution packages, e.g.: [SqlClient](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlActivitySourceHelper.cs#L19)

We also followed the same logic, last year, for our open-source [KafkaFlow](https://github.com/Farfetch/kafkaflow/blob/master/src/KafkaFlow.OpenTelemetry/ActivitySourceAccessor.cs) library and we didn't found any problems with the approach.

However, if you think its safer, I can change it to a hardcoded string without problems

src/Cassandra/OpenTelemetry/TracesRequestObserver.cs
```diff
@@ -0,0 +1,54 @@
...
+//   limitations under the License.
+//
+
+using System;
+using Cassandra.Observers.Abstractions;
+using Cassandra.Requests;
+
+namespace Cassandra.OpenTelemetry
+{
+    internal class TracesRequestObserver : IRequestObserver
```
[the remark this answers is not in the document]
simaoribeiro: As I'm on it, I will also include a folder for the Null and Composite observers, if it's ok for you

src/Cassandra.IntegrationTests/OpenTelemetry/OpenTelemetryTests.cs
simaoribeiro: I agree. The context propagation tests can be straightforward. I will include a method that will be called after the database requests and check if the traceId is the same and also that the database span id is the parent of the method span id.

src/Cassandra/OpenTelemetry/TracesRequestObserver.cs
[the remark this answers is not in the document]
joao-r-reis: yeah sure

src/Extensions/Cassandra.OpenTelemetry/CassandraInstrumentation.cs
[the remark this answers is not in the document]
joao-r-reis: Ok let's keep it as is then

src/Extensions/Cassandra.OpenTelemetry/Implementation/Trace.cs
```diff
@@ -0,0 +1,102 @@
...
+//
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using OpenTelemetry.Trace;
+
+namespace Cassandra.OpenTelemetry.Implementation
+{
+    internal class Trace : IRequestTracker
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Extensions/Cassandra.OpenTelemetry/Implementation/Trace.cs
```diff
@@ -0,0 +1,102 @@
...
+        private readonly CassandraInstrumentationOptions _instrumentationOptions;
+        private static readonly string otelActivityKey = "otel_activity";
+        private static readonly string operationName = "ExecuteAsync";
+
+        public Trace(CassandraInstrumentationOptions instrumentationOptions)
+        {
+            _instrumentationOptions = instrumentationOptions;
+        }
+
+        public Task OnStartAsync(RequestTrackingInfo request)
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Extensions/Cassandra.OpenTelemetry/BuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Cassandra.OpenTelemetry
+{
+    public static class BuilderExtensions
+    {
+        public static Builder AddOpenTelemetryInstrumentation(this Builder builder)
+        {
+            return AddOpenTelemetryInstrumentation(builder, null);
+        }
+
+        public static Builder AddOpenTelemetryInstrumentation(this Builder builder, Action<CassandraInstrumentationOptions> options)
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Cassandra/Requests/RequestExecution.cs
simaoribeiro: @joao-r-reis , I implemented this as you suggested and I also included a method named `OnNodeStart()` in the `RequestExecution.Start()`. 

The idea is to have nested activities for node level requests (that will appear inside the session level requests) that will give the user information regarding the host and execution retries.
As we already had the session start, error, and finish implementations, I think it didn't make much sense to have only the error and finish implementations for node requests as, for activities, there was nothing being started. With that in mind, the activity names are now called `Session Request` and `Node Request` instead of just `Request` as you suggested here.
https://github.com/datastax/csharp-driver/pull/608#issuecomment-2122837103

What do you think about this?

src/Cassandra/OpenTelemetry/TracesRequestObserver.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void OnRequestFailure(Exception ex, RequestTrackingInfo r)
+        {
+            _requestTracker.OnErrorAsync(r, ex);
+        }
+
+        public void OnRequestStart(RequestTrackingInfo r)
+        {
+            _requestTracker.OnStartAsync(r);
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Cassandra/OpenTelemetry/TracesRequestObserver.cs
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Cassandra/Builder.cs
```diff
@@ -1011,6 +1014,12 @@ public Builder WithMetrics(IDriverMetricsProvider driverMetricsProvider, DriverM
             return this;
         }
 
+        public Builder WithRequestTracker(IRequestTracker trace)
+        {
+            _driverTracer = trace;
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Cassandra/IRequestTracker.cs
```diff
@@ -0,0 +1,34 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+
+using System;
+using System.Threading.Tasks;
+
+namespace Cassandra
+{
+    public interface IRequestTracker
```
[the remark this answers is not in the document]
simaoribeiro: Done, please review

src/Cassandra.IntegrationTests/OpenTelemetry/OpenTelemetryTests.cs
simaoribeiro: Working on the tests currently 👍

src/Extensions/Cassandra.OpenTelemetry/Implementation/OpenTelemetryRequestTracker.cs
```diff
@@ -0,0 +1,226 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using OpenTelemetry.Trace;
+
+namespace Cassandra.OpenTelemetry.Implementation
```
joao-r-reis, line 21: I don't think the `Implementation` folder/namespace adds much value here, we can just put this class on the "root" directory, also makes it easier for users without an IDE that auto suggests namespace imports.

src/Cassandra/Requests/RequestExecution.cs
joao-r-reis: Makes perfect sense 👍 I like it.

src/Cassandra/Observers/RequestTracker/RequestTrackerObserver.cs
```diff
@@ -0,0 +1,78 @@
...
+        public async Task OnNodeRequestError(
+            Host host,
+            RequestErrorType errorType,
+            RetryDecision.RetryDecisionType decision,
+            RequestTrackingInfo r,
+            Exception ex)
+        {
+            var hostInfo = new HostTrackingInfo { Host = host };
+            
+            await _requestTracker.OnNodeErrorAsync(r, hostInfo, ex).ConfigureAwait(false);
```
joao-r-reis, line 42: There's a minor optimization that can be done here, I know this was true some years ago but I don't know if modern .NET versions have made this irrelevant:

In cases like this where there's only one `await` and it's the last "statement" of the method then we can just remove `async await` and just return the task itself.

```
public Task OnNodeRequestError(
            Host host,
            RequestErrorType errorType,
            RetryDecision.RetryDecisionType decision,
            RequestTrackingInfo r,
            Exception ex)
        {
            var hostInfo = new HostTrackingInfo { Host = host };

            return _requestTracker.OnNodeErrorAsync(r, hostInfo, ex);
```

joao-r-reis: This can also be done for the other methods of this class. Again it's a very minor thing and not that important but it does make it so there's one less `async/await` state machine that the compiler generates (unless modern .NET versions make this optimization automatically).

src/Cassandra/Observers/Abstractions/IRequestObserver.cs
```diff
@@ -26,10 +27,16 @@ internal interface IRequestObserver
...
+        Task OnNodeRequestError(Host host, RequestErrorType errorType, RetryDecision.RetryDecisionType decision, RequestTrackingInfo r, Exception ex);
 
-        void OnRequestFinish(Exception exception);
+        Task OnNodeSuccess(Host host, RequestTrackingInfo requestTrackingInfo);
+
+        Task OnRequestStart(RequestTrackingInfo requestTrackingInfo);
+
+        Task OnRequestFailure(Exception ex, RequestTrackingInfo requestTrackingInfo);
+
+        Task OnRequestSuccess(RequestTrackingInfo requestTrackingInfo);
```
joao-r-reis, line 40: Should probably add `Async` suffix to the names of these methods to make it consistent.

src/Cassandra/Builder.cs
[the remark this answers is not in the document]
joao-r-reis: This is still not addressed I think.

src/Cassandra/RequestTrackingInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class RequestTrackingInfo
+    {
+        public RequestTrackingInfo()
+        {
+            this.Items = new ConcurrentDictionary<string, object>();
+        }
+
+        public ConcurrentDictionary<string, object> Items { get; }
+
+        public IStatement Statement { get; set; }
```
joao-r-reis, line 30: Do we need a public `setter` here? Ideally the driver should set this when it creates this object and these properties should be immutable (well the `Items` dictionary can be modified but the properties themselves shouldn't).

src/Cassandra.IntegrationTests/OpenTelemetry/OpenTelemetryTests.cs
joao-r-reis: Let me know when I can take a look at the tests 👍

joao-r-reis (commented):
Looks really good, I think most discussions are resolved now, there's only a couple of things left to address. IIUC you're still working on the tests right? I can take another look once that's done.

src/Cassandra/Requests/RequestExecution.cs
```diff
@@ -389,7 +396,7 @@ private void HandleRequestError(IRequestError error, Host host)
                     break;
             }
 
-            _requestObserver.OnRequestError(host, retryInformation.Reason, retryInformation.Decision.DecisionType);
+            _requestObserver.OnNodeRequestError(host, retryInformation.Reason, retryInformation.Decision.DecisionType, _requestTrackingInfo, ex);
```
joao-r-reis, line 399: Hmm we might need to ensure this gets called before `_parent.SetCompleted` because otherwise the "OnRequestFailure" will get called before "OnNodeRequestError" on the request tracker...

And if the retry decision is `Retry` then the new "child trace" will be started before this one ends.

```
switch (retryInformation.Decision.DecisionType)
    {
        case RetryDecision.RetryDecisionType.Rethrow:
           _requestObserver.OnNodeRequestError(host, retryInformation.Reason, RetryDecision.RetryDecisionType.Rethrow, _requestTrackingInfo, ex);
            _parent.SetCompleted(ex);
            break;
        case RetryDecision.RetryDecisionType.Ignore:
            _requestObserver.OnNodeRequestError(host, retryInformation.Reason, RetryDecision.RetryDecisionType.Ignore, _requestTrackingInfo, ex);
            // The error was ignored by the RetryPolicy, return an empty rowset
            _parent.SetCompleted(null, FillRowSet(RowSet.Empty(), null));
            break;
        case RetryDecision.RetryDecisionType.Retry:
            _requestObserver.OnNodeRequestError(host, retryInformation.Reason, RetryDecision.RetryDecisionType.Retry, _requestTrackingInfo, ex);
            //Retry the Request using the new consistency level
            Retry(retryInformation.Decision.RetryConsistencyLevel, retryInformation.Decision.UseCurrentHost, host);
            break;
        default: 
           _requestObserver.OnNodeRequestError(host, retryInformation.Reason, retryInformation.Decision.DecisionType, _requestTrackingInfo, ex);
    }
```

The above isn't pretty but I don't think there's pretty and efficient solutions for this

src/Cassandra/Builder.cs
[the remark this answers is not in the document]
simaoribeiro: Done now.

src/Extensions/Cassandra.OpenTelemetry/Implementation/OpenTelemetryRequestTracker.cs
simaoribeiro: Yeah, I was following the https://github.com/open-telemetry/opentelemetry-dotnet-contrib structure as a guide but it doesn't make much sense in our implementation. I changed it in the last commit

src/Cassandra/Observers/RequestTracker/RequestTrackerObserver.cs
simaoribeiro: I honestly don't know if the latest .NET versions have made the optimization but given that the package is .NET Standard, which can be executed by a .NET Framework runtime, I think it's safer to make the optimization.
Changed it in the last commit

src/Cassandra/Observers/Abstractions/IRequestObserver.cs
simaoribeiro: Done, please review

src/Cassandra/RequestTrackingInfo.cs
simaoribeiro: I changed it so the statement is passed by constructor and the public property is getter only

src/Cassandra/Requests/RequestExecution.cs
simaoribeiro: Changed it in the last commit. Please review

src/Cassandra.IntegrationTests/OpenTelemetry/OpenTelemetryTests.cs
simaoribeiro: Currently I have the following tests:

- When the Keyspace is available check if DbOperation and DbName attributes are included;
- When the Keyspace is not available available check if DbName attribute is not included;
- With default builder options, the statement is not included as an attribute ;
- With the option `IncludeDatabaseStatement` the statement is included as an attribute;
- A test with `Execute` and `Execute` async methods in which I verify if the trace is created and also if the node span is child of session span;
- A test with `Mapper` and `Mapper` async methods in which I verify if the trace is created and also if the node span is child of session span;

Currently missing is the LINQ test and the context propagation that you mentioned above.
I also wanna make a test with a retry to see if multiple spans are generated that are child of the same session span.

I will try to finish in the next few days so, if you have time available, you can start checking the current tests :)

simaoribeiro: @joao-r-reis I finished the tests. Since my previous comment I included
- A test to validate the activity when executing using LINQ
- A test to validate if the same TraceId is used when invoking a different method with a call to Cassandra in the mix
- A test to validate if, when a host is down and a retry happens, if it is produced more than one node activity for the same session activity. During this test, I realized that if multiple retries happens on the same host, the node level activity would be reused instead of being created a new one, so I fixed it in this commit https://github.com/datastax/csharp-driver/pull/608/commits/13f3eec6001b1a5cc74f5cf155445d93037d0b6a
- A test for pagination

simaoribeiro:
There are 3 tests failing regarding to Metrics, I'm still trying to understand why. Anyways, I made a rebase against master to get the last changes and I the commit times ended up a bit messy, but the commits since your last review starts on this one:
https://github.com/datastax/csharp-driver/pull/608/commits/5555e417bc316ec538f2a94fc826ffdcaf81e8b7

joao-r-reis:
Thanks! I'm just wrapping up #611 and I'll take another look at this afterwards.

joao-r-reis:
I'll start taking a look at this now, I'll try to figure out why the tests are failing too

joao-r-reis:
So the issue is that the request observer methods are not being awaited. [I made the changes to support this and the changes turned out to be quite significant](https://github.com/datastax/csharp-driver/commit/b28863ed2898dbe2e5276e45c0f9707bd6eb9345). I'm running a quick benchmark to see if performance is impacted too much by this, if it does then we can just make the request observer (and request tracker) methods synchronous since they aren't needed by the open telemetry extension.

joao-r-reis:
Oh there's a separate issue, there is no `CompositeConnectionObserver`, so the connection observer that handles node metrics is not being used. I'm also fixing that on a separate branch and I'm benchmarking these changes, I'll open a PR that targets this branch afterwards.

joao-r-reis:
Can you add LINQ and Mapper test cases to the following test? `AddOpenTelemetry_WhenMethodIsInvokedAfterQuery_TraceIdIsTheSameThroughRequest`

Just to ensure the parent trace id is properly propagated even when using the mapper and linq.

Also add async to this test too to ensure that this propagation works with Statements,Mapper,LINQ AND both sync and async.

Everything else looks really good, I'm good to merge after this test is modified (or new test(s) is(are) added for these test cases that I mentioned).

joao-r-reis:
[Opened a PR against your branch ](https://github.com/simaoribeiro/csharp-driver/pull/3)to fix the await calls.

simaoribeiro:
Now makes sense why the tests were failing. I was having the problem of the metrics tests failing locally both on the master and opentelemetry branches so I wasn't sure why. Thanks for looking at it.

I'm adding the use cases you mentioned for `AddOpenTelemetry_WhenMethodIsInvokedAfterQuery_TraceIdIsTheSameThroughRequest`. I think I will have it finished by tomorrow afternoon so we can proceed with the merge.

Do you want me to write (or help you write) documentation regarding the feature, after this?

joao-r-reis:
Sorry @simaoribeiro I was about to write that I added the test case I mentioned to the PR I just opened against your branch.

> Do you want me to write (or help you write) documentation regarding the feature, after this?

Yeah, after my PR is merged to your branch I think that all that is left is the documentation. I'm planning on spending the next couple of days setting up a local env to test this new feature so if you can start writing the documentation it would be great.

simaoribeiro:
Sure. I will start working on the documentation tomorrow. Please tell me when I can merge your changes on my repo.

joao-r-reis:
There's 2 tests that are failing on appveyor but they didnt fail before, I'm confused. I'll take a look at them tomorrow.

joao-r-reis:
I think I found the cause of the tests flakyness, [here's the PR ](https://github.com/simaoribeiro/csharp-driver/pull/4)to fix it.

simaoribeiro:
Still failing 🤔 Also taking a look

joao-r-reis:
ok I think I fixed it for good now, just running tests on my branch to make sure

joao-r-reis:
[PR ](https://github.com/simaoribeiro/csharp-driver/pull/5)

simaoribeiro:
I think the value of exportedActivities may be empty sometimes as the tests can be fetching the Activities before they closed. Maybe we should implement a retry mechanism in `GetActivities(DateTime from))`

simaoribeiro:
https://github.com/datastax/csharp-driver/pull/608#issuecomment-2348691563
Yep, that's what I was implementing on my side too but yours looks more complete

joao-r-reis:
Yeah, I thought fixing the await on the ResultHandler would guarantee that we have the activities but the activity exporter probably runs on a separate context so there's still a small chance we won't have all the activities when we retrieve them in the test

joao-r-reis:
Ok they[ passed ](https://ci.appveyor.com/project/DataStax/csharp-driver/builds/50596010), you can merge my PR and let's see if this PR's test run is finally green 😅

simaoribeiro:
Merged. Hopefully this will be the last ride 😂

joao-r-reis:
Looks good now 👍 all that's left is documentation and then we can merge this

simaoribeiro:
@joao-r-reis Included documentation in this commit
https://github.com/datastax/csharp-driver/pull/608/commits/d9c697ba4a1a8a90dab708f61fd817aa8eae5073

I also included another commit to null check the `request.Statement` that was missing
https://github.com/datastax/csharp-driver/pull/608/commits/f5585cd6210129f9c5a77659b79ca33f1a4aeb01

joao-r-reis:
Looks good, I added a comment about null pointer exceptions, I think there's still a chance that they happen, it would be nice to have a unit test or something to validate this.

Can you also add a docs page on the new `Request Tracker` interface?

simaoribeiro:
Trying to finish until the end of the day

joao-r-reis:
Oh also apparently `StartActivity` can also return `null` so this can also throw a NPE:

```
 if (activity.IsAllDataRequested)
 {
     if (!string.IsNullOrEmpty(request.Statement.Keyspace))
     {
         activity.AddTag("db.name", request.Statement.Keyspace);
     }

     if (_instrumentationOptions.IncludeDatabaseStatement && request.Statement != null)
     {
         activity.AddTag("db.statement", request.Statement.ToString());
     }
 }
 ```

**Comment by joao-r-reis** [comment]
I'm currently trying this out with a local app and I just realized that the extension method should probably be changed from `AddOpenTelemetryInstrumentation` to `WithOpenTelemetryInstrumentation` because that's what almost all the other builder methods are named like.

**Comment by joao-r-reis** [comment]
I was actually able to reproduce a scenario where the `Activity` is `null` so we should just return right away if that happens and we need to add checks to other methods because the activity might not exist on the context dictionary.

**Comment by joao-r-reis** [comment]
```
Activity.Tags:
    db.system: cassandra
    db.operation: Session Request
    db.statement: Cassandra.BoundStatement
```
The statement CQL string is not being added, just the type of the statement. It's not a blocker for this PR though, it requires some changes to the statement classes so I'll work on this as a follow up PR to this, after this one is merged.

joao-r-reis:
Actually don't worry about the `AddOpenTelemetryInstrumentation` rename either, I'll handle that one too, let's just make sure we're checking if the `Activity` is `null` then I'll merge the PR and work on a follow up PR to take care of these 2 things I brought up.

simaoribeiro:
Regarding the extension method name, `WithOpenTelemetry` is more in line with the rest of Builder methods so makes perfectly sense to change it

simaoribeiro:
The CQL string is included in some use cases given that the test `AddOpenTelemetry_WithIncludeDatabaseStatementOption_DbStatementIsIncludedAsAttribute` is asserting the value of the activity tag against the query that is included in `session.Execute(statement);`

So, maybe we are lacking some other use cases

joao-r-reis:
> The CQL string is included in some use cases given that the test `AddOpenTelemetry_WithIncludeDatabaseStatementOption_DbStatementIsIncludedAsAttribute` is asserting the value of the activity tag against the query that is included in `session.Execute(statement);`
> 
> So, maybe we are lacking some other use cases

Yeah it works with SimpleStatements but not BatchStatements or BoundStatements, I already have a fix on my branch though so I'll merge that once this PR is merged

joao-r-reis (approved):
Awesome stuff! I'll merge this when CI finishes assuming it's green

simaoribeiro:
Nice. It's a big PR but a good foundation for the next steps. Thanks for all the guidance and help.

There are still some [other attributes](https://opentelemetry.io/docs/specs/semconv/database/cassandra/) that can be included or some behavior that may need to be fixed. I'm available to help if needed.

joao-r-reis:
Yeah, changing the attributes should be simple to do going forward (and don't constitute a breaking change) so even if they aren't 100% perfect on the first release it's not a big deal.

joao-r-reis:
I'll tag you on my follow up PR but it's mostly a justfyi in case you're interested knowing about the changes I make (should be very small though)

simaoribeiro:
Absolutely. Cheers!

PR #610, opened 2024-08-27
Jane Review for Vector Support

SiyaoIsHiding:
I will try to add some tests :)

### DIFF
--- a/src/Cassandra/CqlVector.cs
+++ b/src/Cassandra/CqlVector.cs
@@ -31,9 +31,9 @@ namespace Cassandra
 
         public CqlVector(int dimension)
         {
-            if (dimension < 0)
+            if (dimension <= 0) // C* only allows positive dimension
             {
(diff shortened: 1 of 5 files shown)


joao-r-reis:
Oh I should have clarified, it's not 100% ready for review yet, I only just tested the simplest happy path, there's a ton of test cases to cover yet and proper error handling. Thanks for helping out with these tests though!

src/Cassandra/CqlVector.cs
```diff
@@ -31,9 +31,9 @@ internal CqlVector()
 
         public CqlVector(int dimension)
         {
-            if (dimension < 0)
+            if (dimension <= 0) // C* only allows positive dimension
```
joao-r-reis, line 34: yeah I was planning on testing this but good to get some confirmation thanks!

joao-r-reis (approved):
nice!

PR #613, opened 2024-09-18
OpenTelemetry follow up PR

joao-r-reis:
- Add Distributed Tracing example
- Add `db.query.text` support for `BoundStatement` and `BatchStatement`
- Rename builder extension method
- Add support for speculative executions (mark pending executions as aborted when parent request is completed)

cc @simaoribeiro


joao-r-reis:
I just realized that the way we store the node level activity might not work if there are multiple requests going to the same node as part of the same request context. I think in normal circumstances this doesn't happen but it really depends on the load balancing policy. I'll probably add a `NodeExecutionId` to `HostTrackingInfo` and we can use that id for the key in the context dictionary.

src/Extensions/Cassandra.OpenTelemetry/OpenTelemetryRequestTracker.cs
```diff
@@ -60,14 +65,19 @@ public OpenTelemetryRequestTracker(CassandraInstrumentationOptions instrumentati
         /// <returns>Activity task.</returns>
         public virtual Task OnStartAsync(RequestTrackingInfo request)
         {
-            var activityName = !string.IsNullOrEmpty(request.Statement?.Keyspace) ? $"{sessionOperationName} {request.Statement.Keyspace}" : sessionOperationName;
+            var activityName = !string.IsNullOrEmpty(request.Statement?.Keyspace) ? $"{SessionOperationName} {request.Statement.Keyspace}" : SessionOperationName;
```
simaoribeiro, line 68: As we're changing the `db.operation.name` to include type, we should also change the activity name to follow the specification as we're currently using the pattern `{db.operation.name} {db.namespeace}` for it
https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name

src/Extensions/Cassandra.OpenTelemetry/OpenTelemetryRequestTracker.cs
```diff
@@ -200,25 +212,29 @@ public virtual Task OnNodeErrorAsync(RequestTrackingInfo request, HostTrackingIn
...
+            request.Items.TryGetValue(OtelActivityKey, out object sessionContext);
 
-            if (!(sessionContext is Activity parentActivity) || parentActivity.Context == null)
+            if (!(sessionContext is Activity parentActivity))
             {
                 return Task.CompletedTask;
             }
 
-            var activityName = !string.IsNullOrEmpty(request.Statement?.Keyspace) ? $"{nodeOperationName} {request.Statement.Keyspace}" : nodeOperationName;
+            var activityName = !string.IsNullOrEmpty(request.Statement?.Keyspace) ? $"{NodeOperationName} {request.Statement.Keyspace}" : NodeOperationName;
```
simaoribeiro, line 225: Same as above

simaoribeiro:
Yeah, I think the `NodeExecutionId` makes sense. I didn't know that use case when used the `HostId`

joao-r-reis:
> Yeah, I think the `NodeExecutionId` makes sense. I didn't know that use case when used the `HostId` as part of the key

yeah I just added it now, was a bit of work but it's worth it I think

src/Extensions/Cassandra.OpenTelemetry/OpenTelemetryRequestTracker.cs
joao-r-reis: fixed

simaoribeiro (approved):

simaoribeiro:
Looks really cool.

Btw, if you pass the number of speculative executions, you can add the attribute `[db.cassandra.speculative_execution_count](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/)`. You already did part of the job my parting it `unset` However, I don't think it's important for now and can be added later. \
The prepared statements would be a really useful feature to add but we can also include in the documentation that is not supported for now, so the users are aware.

examples/OpenTelemetry/DistributedTracing/Client/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+                    {
+                        using (var _ = ClientActivity.StartActivity(ActivityKind.Client))
+                        {
+                            await Console.Out.WriteLineAsync("TraceId: " + Activity.Current?.TraceId + Environment.NewLine + "Sending request.").ConfigureAwait(false);
+                            var forecastResponse = await httpClient.GetAsync(WeatherForecastEndpointUri, cts.Token).ConfigureAwait(false);
+
+                            if (forecastResponse.IsSuccessStatusCode)
+                            {
+                                var content = await forecastResponse.Content.ReadAsStringAsync(cts.Token).ConfigureAwait(false);
+                                //var forecast = JsonSerializer.DeserializeAsync<WeatherForecast>(content).ConfigureAwait(false);
```
simaoribeiro, line 42: I think this comment is supposed to be deleted, no?

joao-r-reis:
> Btw, if you pass the number of speculative executions, you can add the attribute [db.cassandra.speculative_execution_count](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/). You already did part of the job my marking it unset However, I don't think it's important for now and can be added later.

Unfortunately there's no way to tell whether an execution is speculative or not atm (could just be a retry) so I'd have to implement that on the driver. Something to consider in the future though.

examples/OpenTelemetry/DistributedTracing/Client/Program.cs
joao-r-reis: nice catch, fixed

doc/features/request-tracker/README.md
```diff
@@ -1,7 +1,7 @@
 # Request Tracker
 
 The driver provides the `IRequestTracker` interface that tracks the requests at Session and Node levels. It contains *start*, *finish*, and *error* events that can be subscribed by implementing the interface, and should be used by passing the implementation as an argument of the method `WithRequestTracker` that is available in the `Builder`.\
-An example of an `IRequestTracker` implementation is the extension package `Cassandra.OpenTelemetry` that can be checked in the [documentation](/doc/features/opentelemetry/README.md).
+An example of an `IRequestTracker` implementation is the extension package `Cassandra.OpenTelemetry` that can be checked in the [documentation](../opentelemetry/README.md).
 
 ## Available events
```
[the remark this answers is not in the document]
joao-r-reis: done

joao-r-reis:
I'm going to merge this when CI finishes.

PR #614, opened 2024-09-19
Add Prepare Request support to RequestTracker and OpenTelemetry extension

joao-r-reis:
cc @simaoribeiro 

It's a draft PR because I'll have to rebase this when #613 is merged, then I'll remove the draft status


joao-r-reis:
I decided to rename `RequestTrackingInfo` to `SessionRequestInfo` and `HostTrackingInfo` to `NodeRequestInfo`, I think these names are more appropriate, wdyt @simaoribeiro

simaoribeiro:
I think it makes sense @joao-r-reis . It makes more clear what those objects are

simaoribeiro:
I'm happy that you found the time to support Prepared Statements. This is looking really good

joao-r-reis:
Sneak peek:

image (4)
image (3)
image (6)
image (5)

simaoribeiro:
These images are looking sweet.

Btw, I just realized that we might have a problem regarding the Keyspace. The keyspace information only appears if its set in the `Statement` object which means that the `db.namespace` attribute may not appear in all use cases

joao-r-reis:
> Btw, I just realized that we might have a problem regarding the Keyspace. The keyspace information only appears if its set in the Statement object which means that the db.namespace attribute may not appear in all use cases

I already knew this but I thought this is ok since the attribute is only required if it's available and you even added a test that verifies that this attribute is not there when the keyspace is not set in the session.

I didn't test this but it is possible that for Prepared Statements the Keyspace will be there even if it is not set at the session level (but then again it's not that relevant since all we can do is set the attribute if the keyspace is there and not set it if it isn't)

simaoribeiro:
Oh, yeah, I forgot that I included that test. Yes, it's fine if the attribute is available when its set.

xsoheilalizadeh:
Hi @joao-r-reis, we have plan to use this feature on our codebase, do you have a ETA for releasing the 3.22.0?

joao-r-reis:
3.22.0 is going out today

PR #615, opened 2024-11-11
CSHARP-1019: Add support for C* 4.1, DSE 6.9.x, and HCD releases to CI

SiyaoIsHiding:
Got green here: https://jenkins-drivers.aws.dsinternal.org/blue/organizations/jenkins/drivers%2Fcsharp%2FJANE_TEST/detail/JANE_TEST/10/pipeline/


Jenkinsfile
```diff
@@ -129,13 +129,26 @@ CCM_CASSANDRA_VERSION=${DSE_FIXED_VERSION} # maintain for backwards compatibilit
 CCM_VERSION=${DSE_FIXED_VERSION}
 CCM_SERVER_TYPE=dse
 DSE_VERSION=${DSE_FIXED_VERSION}
-CCM_IS_DSE=true
+CCM_DISTRIBUTION=dse
```
joao-r-reis, line 132: There's another `Jenkinsfile` that needs this changes too probably. That jenkinsfile is responsible for daily/weekly builds.

Jenkinsfile
```diff
@@ -435,10 +448,13 @@ pipeline {
...
                   '4.0',    // latest 4.0.x Apache Cassandra�
-                  '5.0-beta1', // Development Apache Cassandra�
+                  '4.1',
+                  '5.0', // Development Apache Cassandra�
                   'dse-5.1.35', // latest 5.1.x DataStax Enterprise
                   'dse-6.7.17', // latest 6.7.x DataStax Enterprise
-                  'dse-6.8.30' // 6.8 current DataStax Enterprise
+                  'dse-6.8.30', // 6.8 current DataStax Enterprise
+                  'dse-6.9.3',
+                  'hcd-1.0.0'
```
joao-r-reis, line 457: We need to prune this a bit, the matrix is too long for a "per-commit" build. We can run all the versions we want on the daily/weekly builds.

Maybe we can use `3.11`,  `4.1`, `5.0`, `dse-5.1`, `dse-6.9.3` and `hcd-1.0.0` for `per-commit`

src/Cassandra.IntegrationTests/Core/SchemaMetadataTests.cs
```diff
@@ -34,11 +34,8 @@ public SchemaMetadataTests() :
             base(() =>
             {
                 string[] cassandraYaml = null;
-                if (TestClusterManager.CheckCassandraVersion(true, new Version(5, 0), Comparison.GreaterThanOrEqualsTo))
-                {
-                    cassandraYaml = new[] { "materialized_views_enabled: true" };
-                }
-                else if (TestClusterManager.CheckCassandraVersion(true, new Version(4, 0), Comparison.GreaterThanOrEqualsTo))
+                if (TestClusterManager.CheckCassandraVersion(true, new Version(4, 0), Comparison.GreaterThanOrEqualsTo) ||
+                    TestClusterManager.IsHcd)
```
joao-r-reis, line 38: Why did you remove `cassandraYaml = new[] { "materialized_views_enabled: true" };` for `5.0+`?

joao-r-reis: Same for user defined functions

src/Cassandra.IntegrationTests/TestClusterManagement/CcmCluster.cs
```diff
@@ -214,23 +214,17 @@ public void BootstrapNode(int nodeIdToStart, string dataCenterName, bool start =
         
         public void UpdateDseConfig(params string[] yamlChanges)
         {
-            if (yamlChanges == null) return;
-            var joinedChanges = string.Join(" ", yamlChanges.Select(s => $"\"{s}\""));
-            _ccm.ExecuteCcm($"updatedseconf {joinedChanges}");
+            _ccm.UpdateDseConfig(yamlChanges);
```
joao-r-reis, line 217: This is a nice way to reuse existing code but it will result in a different behavior and possibly longer execution times because it will now execute a ccm command per line instead of a single ccm command for all of them.

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
```diff
@@ -29,6 +29,7 @@ public static class TestClusterManager
         public const string DefaultKeyspaceName = "test_cluster_keyspace";
         private static ICcmProcessExecuter _executor;
 
+        private static readonly Version Version1Dot0 = new Version(1, 0);
```
joao-r-reis, line 32: We don't need this one do we?

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
```diff
@@ -70,15 +73,18 @@ public static Version CassandraVersion
                         // C* 3.0
                         return Version3Dot0;
                     }
-                    if (dseVersion < Version6Dot0)
+                    if (dseVersion <= Version6Dot9)
```
joao-r-reis, line 76: Hmm this is a tricky change. Technically DSE 6.x is in between 3.11 and 4.0 (it has some 4.0 features). This might result in some DSE 6.x tests not running because we declared the test as compatible with C* 4.0 and we relied on it working this way to run for DSE as well. Maybe try to check the total test count per server version before and after the changes to make sure we're not losing tests with these changes.

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
```diff
@@ -99,14 +105,50 @@ public static string DsePath
...
+        }
+        
+        /// <summary>
+        /// "hcd", "dse", or "cassandra" (default)
+        /// </summary>
+        public static BackendType CurrentBackendType
+        {
+            get
+            {
+                string distribution = Environment.GetEnvironmentVariable("CCM_DISTRIBUTION") ?? "cassandra";
```
joao-r-reis, line 122: Currently the test project is relying on checking for `DSE_VERSION` to see if the tests are running against DSE or not so we should also check if `DSE_VERSION` exists here before falling back to `cassandra`.

joao-r-reis: Ok I see you also changed the implementation of `IsDse`. I still think we should check the `DSE_VERSION` env variable here before falling back to `cassandra`

src/Cassandra/SchemaParser.cs
```diff
@@ -840,8 +840,7 @@ public override Task<AggregateMetadata> GetAggregateAsync(string keyspaceName, s
                             StateFunction = row.GetValue<string>("state_func"),
                             FinalFunction = row.GetValue<string>("final_func"),
                             InitialCondition = row.GetValue<string>("initcond"),
-                            Deterministic = row.GetColumn("deterministic") != null &&
-                                            row.GetValue<bool>("deterministic"),
+                            Deterministic = row.GetColumn("deterministic") != null && row.GetValue<bool?>("deterministic").GetValueOrDefault(false),
```
joao-r-reis, line 843: Isn't this a server bug? Or is it working as intended to get null values here?

src/Cassandra.IntegrationTests/Core/SchemaMetadataTests.cs
SiyaoIsHiding: Because we fix them here: https://github.com/datastax/csharp-driver/blob/7eaff90f96b786e725e33821fcef19460d112071/src/Cassandra.IntegrationTests/TestClusterManagement/CcmBridge.cs#L340-L373

src/Cassandra.IntegrationTests/TestClusterManagement/CcmCluster.cs
SiyaoIsHiding: I think it's still better to reuse and delegate to `CcmBridge.UpdateConfig`. `CcmBridge.UpdateConfig` also fixes the yaml configs for C* 4.1+, too.
I can change `CcmBridge.UpdateConfig` to join the config strings and run only one ccm command.

joao-r-reis: > I can change CcmBridge.UpdateConfig to join the config strings and run only one ccm command.

Yeah if this doesn't break stuff then it will be perfect 👍

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
SiyaoIsHiding: Before the fix, dse-6.7.17 net6: total tests 1145, skipped 108.
After the fix, dse-6.7.17 net6: total tests 1145, skipped 107.

Before fix, dse-6.8 net6, skipped 28
After fix, dse-6.8 net6, skipped 27

So after my fix, it actually skips fewer tests. I think it may be because my change in `CheckCassandraVersion`. Previously it maps C* version to DSE version when comparing, using `GetDseVersionFromCassandraVersion`. I changed it so it maps DSE version to C* version instead. I think it makes more sense, and align with other drivers better, too.

joao-r-reis: > Before fix, dse-6.8 net6, skipped 28
> After fix, dse-6.8 net6, skipped 27

Do you have a total number for `dse-6.8 net6` similar to `dse-6.7.17 net6`?

SiyaoIsHiding: dse-6.8.30

joao-r-reis: I mean total number of tests, you wrote the total number of tests dse-6.7 but not dse-6.8

SiyaoIsHiding: Oh the same, the number of total tests is 1145. This is the number of tests before skipping any, so it will be the same for any versions, before and after fix.

src/Cassandra/SchemaParser.cs
SiyaoIsHiding: Based on the current status of the ticket, it looks like a bug: https://datastax.jira.com/browse/DSP-24606
What do we want to do here?

joao-r-reis: let's remove all production code changes from this PR and keep the test changes and make the failing HCD tests skip when it's HCD with a message that refers `DSP-24606` (no full link since it's a private JIRA project).

The changes are trivial but I don't really want to do a release for the C# driver when this is something that will be fixed in HCD (and it's highly unlikely someone will run into this before an HCD fix is in place anyway).

src/Cassandra.IntegrationTests/TestClusterManagement/TestClusterManager.cs
SiyaoIsHiding: The one less skip count is about `Should_FailFast_When_PreparedStatementIdChangesOnReprepare`, because "This test relies on a bug that is fixed in this server version." I assume that test does not matter in this case.
Additionally, I just found out that `[TestCassandraVersion(x,x)]` never maps DSE version to C* OSS version. Say if you say `[TestCassandraVersion(5, 0)]` and the executing version is DSE-6.7, then 6.7 > 5.0, so the test will proceed to run.
Do we agree that
1. We should map DSE 6.x to C* 3.11, and
2. We change `[TestCassandraVersion(x,x)]` to map DSE version to C* version before it compares?

joao-r-reis:
Apparently there's some git conflicts? Did you create your branch from an old version of `master`?

src/Cassandra.IntegrationTests/TestBase/TestCassandraVersion.cs
```diff
@@ -24,7 +24,7 @@ namespace Cassandra.IntegrationTests.TestBase
     /// <summary>
     /// An attribute that filters the test to execute according to the current DSE version.
     /// </summary>
-    public class TestDseVersion : NUnitAttribute, IApplyToTest
+    public class TestCassandraVersion : NUnitAttribute, IApplyToTest
```
joao-r-reis, line 27: Isn't there a `TestCassandraVersion` attribute already?

SiyaoIsHiding:
Currently all the vector-related tests are marked with [TestBothServerVersions(5,0,6,9)] so they all fail with DSE-6.9, because those vector tests are with various subtypes, while DSE-6.9 only support vector<float>. So I changed the existing vector tests to [TestCassandraVersion(5,0)] and I m adding some vector<float> tests for DSE-6.9.

joao-r-reis:
>  while DSE-6.9 only support vector

I don't get this, you mean it only supports vector of floats?

joao-r-reis:
> Yes yes sorry I missed that. DSE-6.9 only supports vector of floats. I'm adding tests of vector of floats for DSE-6.9.

Could also just add an `if` to the test to make sure all test cases are skipped except when subtype is `float` when it's DSE-6.9

joao-r-reis:
Changes look good but it looks like we essentially doubled the matrix size for the "Per Commit" builds so we should try to trim it even further.

Let's do the following for Per Commit:

3.11
4.1
5.0
dse-5.1.35
dse-6.8.30
dse-6.9.3
hcd-1.0.0

Run all of these versions against `net6`.
Only `5.0`, `dse-6.9.3` and `hcd-1.0.0` against `mono`
Only `4.1`, `dse-5.1.35` and `dse-6.8.30` against `net8`

This will result in an increase of 2 builds (13) vs the number in master currently (11).

After doing this change and veirfying that CI is green you can merge this 👍

Jenkinsfile
```diff
@@ -463,7 +476,7 @@ pipeline {
             }
             axis {
               name 'SERVER_VERSION'
-              values 'dse-6.7.17', '3.11'
+              values '3.11', 5.0', 'dse-6.9.3', 'hcd-1.0.0'
```
joao-r-reis, line 479: missing `'` before `5.0`

SiyaoIsHiding:
13 builds now

joao-r-reis:
Matrix looks good now 👍

